PR #1, opened 2017-08-16
Add support for configurable import paths.

robshakir:
This change adds support for an external caller to specify the import path that is to be used for dependencies of the generated code, such that in build environments that do not use GitHub, an alternate path can be specified.

```
  * (M) ygen/codegen.go
  * (M) ygen/codegen_test.go
    - Add support for specifying the path to ygot, ytypes and goyang
      in the GeneratorConfig struct such that the calling binary can
      specify its own import paths.
  * (M) ygen/commongen.go
    - Remove the common header building function since its contents
      are no longer common with other languages.
  * (M) ygen/gogen.go
    - Update the Go template generating the header to accept the
      paths in the configuration that is supplied, using defaults
      where they are not set.
    - Remove consumption of the common header builder and rather
      retain this logic in the writeGoHeader function.
  * (M) ygen/testdata/schema/openconfig-options-explicit-schema.json
  * (M) ygen/testdata/schema/openconfig-options-explicit.formatted-txt
    - Add test case for overriding variables through config.
  * (M) generator/generator.go
    - Add support for passing the flags for import paths through to
      the ygen library.
```


ygen/codegen.go
```diff
@@ -87,6 +87,18 @@ type GoOpts struct {
 	// produced if the GenerateJSONSchema YANGCodeGenerator field is set to
 	// true.
 	SchemaVarName string
+	// GoyangImportPath specifies the path that should be used in the generated
+	// code for importing the goyang/pkg/yang package. If it is not specified
```
ostromart: Suggest omitting the last sentence, here and below, since this duplicates the information already in the flag default.

ygen/codegen_test.go
```diff
@@ -966,17 +966,14 @@ func TestBuildStructDefinitions(t *testing.T) {
...
-	wantStructsCodeFile  string   // The path of the generated Go code that the output of the test should be compared to.
-	wantErr              bool     // Whether to expect an error.
-	wantSchemaFile       string   // Path of the generated Schema JSON that the output of the test should be compared to.
+	name                string          // Name is the identifier for the test.
+	inFiles             []string        // inFiles is the set of inputFiles for the test.
+	inIncludePaths      []string        // inIncludePaths is the set of paths that should be searched for imports.
+	inExcludeModules    []string        // inExcludeModules is the set of modules that should be excluded from code generation.
+	inConfig            GeneratorConfig // inConfig specifies the configuration that should be used for the generator test case.
+	wantStructsCodeFile string          // The path of the generated Go code that the output of the test should be compared to.
+	wantErr             bool            // Whether to expect an error.
```
ostromart: nit: keep comment format consistent.

ygen/gogen.go
```diff
@@ -34,6 +34,20 @@ const (
...
+
+var (
+	// packageImportPaths stores the default import paths that should be used
+	// within the generated Go. It is keyed by a short name for a particular
+	// import.
+	packageImportPaths = map[string]string{
+		"ygot":   "github.com/openconfig/ygot/ygot",
+		"goyang": "github.com/openconfig/goyang/pkg/yang",
+		"ytypes": "github.com/openconfig/ygot/ytypes",
+	}
```
ostromart: These paths look to be duplicated in generator.go.

ygen/codegen.go
robshakir: The logic for having it in both places was that the `ygen` library can be called from places other than `generator`, so it seemed better to have it here, rather than in `generator`.

ygen/gogen.go
robshakir: They were, see the comment above for why this was the case. Essentially, the test cases within the library fail if we just have them in `generator.go`.

I refactored this such that we have them in `gogen.go` as constants that are referred to in `generator.go` such that the user of that binary knows what the import paths are expected to be by default. PTAL.

ygen/codegen_test.go
robshakir: Done.

ygen/codegen.go
robshakir: ACK --  that was the change that I made here: https://github.com/openconfig/ygot/pull/1/commits/b65a01e2cdbb3bc8cbc928752e8bb31c57ced425#diff-338bea067d2a01b69c90f1c032c6a24b

ostromart: I meant that I would suggest deleting "If not specified it is imported from github.com/openconfig/ygot/ygot." - because it's describing in some behavior that's not related to this structure (how this is initialized should be described in the function where it's actually set to that value) and also introduces a redundancy.

ygen/gogen.go
```diff
@@ -543,8 +555,48 @@ func makeTemplate(name, src string) *template.Template {
...
+	if cfg.Caller == "" {
+		cfg.Caller = callerName()
+	}
+
+	if cfg.PackageName == "" {
+		cfg.PackageName = defaultPackageName
+	}
+
+	if cfg.GoOptions.YgotImportPath == "" {
+		cfg.GoOptions.YgotImportPath = DefaultYgotImportPath
```
ostromart: Add something like "If cfg.GoOptions.YgotImportPath is empty, it is set to DefaultYgotImportPath." in the function description, for this and other defaults.

ygen/codegen.go
robshakir: Done.

ygen/gogen.go
robshakir: Done.

PR #4, opened 2017-08-18
Add initial protobuf generation code

robshakir:
This PR adds the initial functions for mapping YANG to protobuf in ygen. Primarily, it adds a new `yangTypeToProtoType` function to `genState`.

Some mechanical cleanup of testing that was mistakenly moved to `genstate_test.go` is done - such that this is now within `goelements_test.go`.

The pattern that this, and subsequent PRs, aim to implement is to add handlers for outputting proto files from a new `protogen.go` file. `protoelements.go` will handle the proto-specific functions for handling type mapping where it is required.

`protoelements` use `genState` as a receiver due to the fact that naming is required throughout these functions, and it is desirable to keep the Go-generating code separate from that which generates protos.


ygen/protoelements_test.go
```diff
@@ -0,0 +1,91 @@
...
+		want: mappedType{nativeType: "ywrapper.BoolValue"},
+	}, {
+		name: "string",
+		in:   resolveTypeArgs{yangType: &yang.YangType{Kind: yang.Ystring}},
+		want: mappedType{nativeType: "ywrapper.StringValue"},
+	}, {
+		name: "decimal64",
+		in:   resolveTypeArgs{yangType: &yang.YangType{Kind: yang.Ydecimal64}},
+		want: mappedType{nativeType: "ywrapper.Decimal64Value"},
+	}}
```
ostromart: Cover -ve cases?
Also, although this format is ok, in the future I suggest using a map or list of stucts at the top and just ranging through it to avoid lots of duplicated boilerplace.

robshakir: Refactored the tests to use a better list structure and covered the -ve cases.

PR #7, opened 2017-08-22
Scaffolding for Protobuf Generation

robshakir:
Add scaffolding for protobuf generation to `ygen` -- this PR adds a function which forms the basis of generating protobuf messages from input `yangStruct` structs. It includes a placeholder for functions to generate tags which is a TODO.


ygen/protogen.go
```diff
@@ -0,0 +1,115 @@
...
+			continue
+		case field.IsDir():
+			msgName, ok := state.uniqueStructNames[field.Path()]
+			if !ok {
+				errs = append(errs, fmt.Errorf("proto: could not resolve %s into a defined struct", field.Path()))
+				continue
+			}
+			fieldDef.Type = msgName
+		default:
+			// This is a YANG leaf, or leaf-list.
```
ostromart: Suggest changing this to explicitly check IsLeaf || IsLeafList and leave default as an error case.

ygen/protogen.go
```diff
@@ -0,0 +1,115 @@
...
+	definedFieldNames := map[string]bool{}
+
+	for name, field := range msg.fields {
+		fieldDef := &protoMsgField{
+			Name: makeNameUnique(safeProtoFieldName(name), definedFieldNames),
+		}
+
+		t, err := protoTagForEntry(field)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("proto: could not generate tag for field %s: %v", field.Name, err))
```
ostromart: continue here? see comment below for restructure suggestion.

ygen/protogen.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		fieldDef.Tag = t
+
+		switch {
+		case field.IsList():
+			errs = append(errs, fmt.Errorf("proto: list generation unimplemented for %s", field.Path()))
+			continue
```
ostromart: Suggest creating var err error at the top of the loop, then set it when you hit error inside here. Then check for error and append at the exit to the switch. This will remove the need to continue on error.

ygen/protogen_test.go
```diff
@@ -0,0 +1,208 @@
...
+		name: "simple message with only scalar fields",
+		inMsg: &yangStruct{
+			name: "MessageName",
+			entry: &yang.Entry{
+				Name: "message-name",
+				Dir:  map[string]*yang.Entry{},
+			},
+			fields: map[string]*yang.Entry{
+				"field-one": {
+					Name: "field-one",
```
ostromart: test unsafe field char replacement?

ygen/protogen_test.go
```diff
@@ -0,0 +1,208 @@
...
+							Type: &yang.YangType{Kind: yang.Ystring},
+						},
+					},
+					Key: "key",
+				},
+			},
+			path: []string{"", "a-messsage-with-a-list", "list"},
+		},
+		wantErr: true,
+	}}
```
ostromart: test for yang types that don't resolve to proto types.

ygen/protogen.go

ygen/protogen.go
robshakir: This is fatal, thanks.

ygen/protogen.go

ygen/protogen_test.go
robshakir: This actually does test the field character replacement -- since "-" is allowed in YANG and not in protobuf. I cleaned this up since there was actually a case that could never occur.

ygen/protogen_test.go
robshakir: Done.

robshakir:
Thanks for the review. PTAL at the updates.

PR #9, opened 2017-08-22
Rename yangStruct to yangDirectory, move tests

robshakir:
Mechanical changes to `ygen` to particularly:

 * Rename `yangStruct` to `yangDirectory` - since this can be used to represent a protobuf `message` as well as a Go `struct`.
 * Move test code from `codegen_test.go` to `genstate_test.go` where owning function had been moved during refactor.


robshakir:
Prefer `yangDirectory` since `yangNode` could be a scalar type rather than a type that contains other elements. I think this rename is consistent with the naming in goyang.

PR #11, opened 2017-08-24
Integrate first simple end-to-end proto3 generation.

robshakir:
This change adds the initial generation of proto3 messages in a form that can be used. It adds testing using simple YANG schema inputs through the toolchain.


ygen/codegen.go
```diff
@@ -367,6 +393,101 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
...
+			p.Packages[pkg] = Proto3Package{
+				Messages: []string{},
+			}
+			tp = p.Packages[pkg]
+		}
+		tp.Messages = append(tp.Messages, msg)
+		p.Packages[pkg] = tp
+	}
+
+	// Ensure that the base import path is a valid directory.
```
ostromart: Suggest turning this into a helper function (e.g. reformatDir) and implementing in terms of filepath e.g. 
return filepath.Join(strings.Split(path, "/")...)
You could add the trailing slash or (better) continue to use filepath.Join wherever the dir is used. 
Also, the comment suggests an error check so it may be better to say something like "Reformat the directory path into canonical form".

ygen/codegen_test.go
```diff
@@ -641,3 +624,96 @@ func TestFindRootEntries(t *testing.T) {
...
+		inFiles        []string
+		inIncludePaths []string
+		inConfig       GeneratorConfig
+		// wantOutputFiles is a map keyed on protobuf package name with a path
+		// to the file that is expected for each package.
+		wantOutputFiles map[string]string
+		wantErr         bool
+	}{{
+		name:     "simple protobuf test with compression",
+		inFiles:  []string{filepath.Join(TestRoot, "testdata", "proto", "proto-test-a.yang")},
```
ostromart: To avoid repeating this Join, you could define a subdir e.g. baseDir := TestRoot + "/testdata/proto" and use filepath.Join in the test loop to add just the filename. btw. filepath will correctly handle Join("a//b/", "/c") -> "a/b/c"

ygen/protogen.go
```diff
@@ -88,21 +101,69 @@ message {{ .Name }} {
 	}
 )
 
+// writeProtoHeader outputs the header for a proto3 generated file. The package
+// name, base for the generated packages, files for which the protobuf was
+// generated, and the paths used for include are supplied as arguments. In
+// addition, whether path compression is being performed, and the calling
+// binary name are provided as arguments. Returns a string containing the
```
ostromart: For such a long arg list it may be better split it into list format e.g. 
  pkg blah
  basePkg blah blah, set to defaultBasePackageName if empty
nit: go readabilty reviewers would insist on "It returns...". Up to you.

ygen/protogen.go
```diff
@@ -88,21 +101,69 @@ message {{ .Name }} {
...
+
+	if caller == "" {
+		caller = callerName()
+	}
+
+	if basePkg == "" {
+		basePkg = defaultBasePackageName
+	}
+
+	switch pkg {
```
ostromart: Seems like if/else would be more compact/suitable here. Up to you.

ygen/protogen.go
```diff
@@ -88,21 +101,69 @@ message {{ .Name }} {
...
 // it uses the context of other messages to be generated (msgs), and the generator
 // state stored in state to determine names of other messages. compressPaths indicates
 // whether path compression should be enabled for the code generation. Returns a string
 // containing the name of the package that the message is within, a string containing
-// the generated code for the protobuf message, and any errors encountered during
+// the generated code for the protobuf message, a slice of strings containing the child
+// packages that are required by this message and any errors encountered during
 // proto generation.
-func writeProtoMsg(msg *yangDirectory, msgs map[string]*yangDirectory, state *genState, compressPaths bool) (string, string, []error) {
+func writeProtoMsg(msg *yangDirectory, msgs map[string]*yangDirectory, state *genState, compressPaths bool) (string, string, []string, []error) {
```
ostromart: Is this just proto3? If so it may be better  to call it writeProto3Msg. Just to be clear, I'm suggesting keeping type/function names version neutral if they support/can support all versions but adding version where a particular version is expected or emitted.
Another possibility is to have a enum parameter ProtoVersion2, ProtoVersion3 and passing that around.

ygen/protogen.go
```diff
@@ -88,21 +101,69 @@ message {{ .Name }} {
...
 	msgDef, errs := genProtoMsg(msg, msgs, state, compressPaths)
```
ostromart: genProto3Msg?

ygen/protogen.go
```diff
@@ -88,21 +101,69 @@ message {{ .Name }} {
...
-		return "", "", errs
+		return "", "", nil, errs
 	}
 
 	if msg.entry.Parent == nil {
-		return "", "", []error{fmt.Errorf("YANG schema element %s does not have a parent, protobuf messages are not generated for modules", msg.entry.Path())}
+		return "", "", nil, []error{fmt.Errorf("YANG schema element %s does not have a parent, protobuf messages are not generated for modules", msg.entry.Path())}
 	}
 
 	// pkg is the name of the protobuf package, if the entry's parent has already
```
ostromart: incomplete comment?

ygen/protogen.go
```diff
@@ -134,8 +195,20 @@ func genProtoMsg(msg *yangDirectory, msgs map[string]*yangDirectory, state *genS
 	}
 
 	definedFieldNames := map[string]bool{}
+	imports := []string{}
+
+	// Traverse the fields in alphabetical order to ensure determinsitic output.
```
ostromart: deterministic

ygen/codegen.go
```diff
@@ -99,6 +101,16 @@ type GoOpts struct {
 	YtypesImportPath string
 }
 
+// ProtoOpts sotres Protobuf specific options for the code generation library.
```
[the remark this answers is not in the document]

ygen/codegen.go
```diff
@@ -178,6 +190,20 @@ type GeneratedGoCode struct {
 	RawJSONSchema []byte
 }
 
+// GeneratedProto3 stores a set of generated protobuf3 packages.
+type GeneratedProto3 struct {
```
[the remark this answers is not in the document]
robshakir: SGTM!

ygen/codegen.go
```diff
@@ -367,6 +393,101 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
 	}, nil
 }
 
+// GenerateProto3 generates Protobuf 3 code for the input set of YANG files.
+// The YANG schemas for which protobufs are to be created is supplied as the
+// yangFiles argument, with included modules being searched for in includePaths.
+// Returns a GeneratedProto3 struct containing the messages that are to be
+// output, along with any associated values (e.g., enumerations).
+func (cg *YANGCodeGenerator) GenerateProto3(yangFiles, includePaths []string) (*GeneratedProto3, *YANGCodeGeneratorError) {
+	// TODO(robjs): Handle enumerated types in proto messages.
```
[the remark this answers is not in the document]
robshakir: Agreed -- I guess we just need to decide how best we should log errors. Perhaps just create an issue here? The plan is to resolve this in a subsequent CL here.

ygen/codegen.go
```diff
@@ -367,6 +393,101 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
...
+	msgs, _, st, errs := langAgnosticDefinitions(yangFiles, includePaths, cg.Config)
+	if len(errs) > 0 {
```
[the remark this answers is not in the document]

ygen/codegen.go
```diff
@@ -367,6 +393,101 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
...
+
+	cg.state.schematree = st
+
+	protoMsgs, errs := cg.state.buildDirectoryDefinitions(msgs, cg.Config.CompressOCPaths, cg.Config.GenerateFakeRoot, protobuf)
+
+	if len(errs) > 0 {
+		return nil, &YANGCodeGeneratorError{Errors: errs}
+	}
+
+	p := &GeneratedProto3{
```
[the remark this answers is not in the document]
robshakir: Done - renamed to `genProto`.

ygen/codegen.go
```diff
@@ -367,6 +393,101 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
...
+		Packages: map[string]Proto3Package{},
+	}
+	ye := NewYANGCodeGeneratorError()
+
+	// Write out the protobuf messages in a determinstic order.
```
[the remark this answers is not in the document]
robshakir: Corrected. Left comment as is because I think alphabetical and deterministic are both important here.

ygen/codegen.go
```diff
@@ -367,6 +393,101 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
...
+			if !found {
+				pkgImports[pkg] = append(pkgImports[pkg], i)
+			}
+		}
+
+		if len(errs) > 0 {
+			ye.Errors = append(ye.Errors, errs...)
+		}
+
+		tp, ok := p.Packages[pkg]
```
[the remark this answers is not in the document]

ygen/codegen.go
robshakir: Done - added template helper functions that do this.

ygen/codegen_test.go
robshakir: Since we need a filepath.Join everywhere, it doesn't seem like it removes enough to do this. Prefer this approach.

ygen/protogen.go

ygen/protogen.go

ygen/protogen.go
robshakir: It is just proto3. I'll rename these functions to proto3.

ygen/protogen.go

ygen/protogen.go
robshakir: Fixed.

ygen/protogen.go
robshakir: Fixed elsewhere.

robshakir:
Thanks for the review, PTAL at the changes.

ygen/codegen.go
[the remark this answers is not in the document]
ostromart: I'd probably create an issue and put TODO(some ptr to issue) in general for anything major. I think it's worth having in this case so that someone who comes along doesn't expect the code to be fully functional.

ygen/protogen.go
```diff
@@ -101,14 +101,21 @@ message {{ .Name }} {
...
+//  - imports - the other protobuf packages that should be imported by the
+//    protobuf package.
+//  - yangFiles - the set of files that were input to the library.
+//  - includePaths - the set of include paths that were input to the library.
+//  - compressPaths - a bool indicating whether path compression should
+//    be used.
+//  - caller - the name of the calling binary - included in the package
+//    header.
+// It returns a string containing the generated header code.
+func writeProto3Header(pkg, basePkg, baseImportPath string, imports, yangFiles, includePaths []string, compressPaths bool, caller string) (string, error) {
```
ostromart: (optional) sorry, should have mentioned in the first round: the list of params here is very long and has a bunch of fields of the same type which are easy for the caller to mix up - consider defining a struct.

ygen/codegen.go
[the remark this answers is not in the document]
robshakir: ACK -- I created #20 here, which is addressed in a subsequent PR in this stack of them.

ygen/codegen.go
```diff
@@ -367,6 +393,101 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
...
+	for n, pkg := range p.Packages {
+		h, err := writeProtoHeader(n, cg.Config.ProtoOptions.BasePackageName, baseImportPath, pkgImports[n], yangFiles, includePaths, cg.Config.CompressOCPaths, cg.Config.Caller)
+		if err != nil {
+			ye.Errors = append(ye.Errors, errs...)
+		}
+		pkg.Header = h
+		p.Packages[n] = pkg
+	}
+
+	if len(ye.Errors) > 0 {
```
ostromart: ye.Errors != nil

ygen/protogen.go
ostromart: I mean you have a typo there: s/determinsitic/deterministic/

ygen/protogen.go

robshakir:
Fixed the typo too.

PR #12, opened 2017-08-28
Add files to allow dep ensure to be run.

robshakir:
```
 * (A) .gitignore
   - Don't keep the vendor directory within the ygot dir.
 * (A) Gopkg.lock
 * (A) Gopkg.toml
   - Add files to allow go dep to be used against ygot.
```


Gopkg.toml
```diff
@@ -0,0 +1,46 @@
...
+  branch = "master"
+  name = "github.com/kylelemons/godebug"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/openconfig/gnmi"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/openconfig/goyang"
```
ostromart: We may want consider controlling the versions for goyang and gnmi at least. That way if their behavior changes it will be easy to go back in time to the right compatible versions (should we ever need to do that).

robshakir: Yeah, I think this would be worth considering. Right now, I don't expect either changes to be backwards incompatible, but we should determine whether it's still worth fixing the versions. 

The tests use `go get` rather than `dep ensure` so would be decoupled from this. I wonder whether it's possible for us to get travis to run a set of tests that must pass (using `dep ensure`) and then some informational smoke tests with unfixed versions and pulling the OpenConfig schema.

PR #16, opened 2017-08-31
Support Maps and non-Directories that are direct children of the root.

robshakir:
This PR adds support for entries that are directly at the root, particularly:
* Maps: Previously, a map that was directly at the root would output the key, and then the parent map name after this, this PR fixes this by making the map entries have relative paths, and including the map name in the path at the root.
* Non-directories (`leaf` and `leaf-list`). In these cases, since these are not "mappable entities" then there was no code output for a non-directory entry that was at the root. This PR adds support for these to be output in the generated code.

It adds an uncompressed schema test in `demo/uncompressed` which is also used for an integration test for these changes.


ygen/gogen.go
```diff
@@ -1094,12 +1094,16 @@ func findMapPaths(parent *yangStruct, field *yang.Entry, compressOCPaths bool) (
...
-		// the mapping path specified correctly. Otherwise, we take all the elements other
-		// than the first.
+		// that has been mapped, if it is a container, then an empty string can be
+		// specified as the container has its own name encoded. In the case that it is
+		// a list or a non-directory entry, then we need to include the name of the entry
+		// in the path.
 		switch len(fieldSlicePath) {
 		case 2:
 			childPath = []string{}
+			if field.IsList() || field.IsLeaf() || field.IsLeafList() {
```
ostromart: !IsContainer?

ygen/gogen.go
```diff
@@ -1123,7 +1127,14 @@ func findMapPaths(parent *yangStruct, field *yang.Entry, compressOCPaths bool) (
 			// is prefixed with a /.
 			// Absolute paths are used for top-level entities such rendering functions
 			// can be agnostic to whether the fakeroot was created.
-			childPath = append(childPath, []string{"", parentPath[2]}...)
+			// A special case exists where there are lists at the root, since in this case
+			// we cannot simply have the parent path prepended to the current element's
+			// name, and therefore we do not use absolute paths.
+			if parent.entry != nil && parent.entry.IsList() {
```
ostromart: Isn't this something that's more specific to json output? If so, would it be better to put special handling for lists there?

robshakir: I think it's more general -- although this comment in written around this case. For `Validate` - when we have compressed schemas, then we have `path:"foo/bar"` (where `bar` is the list) in the fakeroot - this allows you to find the `bar` schema. Prior to this change we have `path:""` in the fakeroot - such that `Validate` can't find an empty schema for the list there. Adding this here means that `Validate` can use `path` to find the list in the compressed and uncompressed cases.

ygen/gogen.go
robshakir: Done.

PR #31, opened 2017-09-11
Fix NewNode ptr type handling, add real schema tests.

ostromart:
- Fix handling for ptr types in NewNode (would previously return correct type, but not initialized).
 - Add coverage to handle non-ptr types.
 - Add tests with exampleoc schema.
 - Remove schema return value in public GetNode API, as this can be obtained from SchemaTree.

### DIFF
--- a/ytypes/schema_tests/validate_test.go
+++ b/ytypes/schema_tests/validate_test.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// TODO(mostrowski): create tests against an uncompressed schema.
 package validate
 
 import (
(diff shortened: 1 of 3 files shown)


ytypes/schema_tests/validate_test.go
```diff
@@ -452,6 +742,40 @@ func diffJSON(a, b []byte) (string, error) {
...
+		return "{ " + out + " }"
+	}
+	out := fmt.Sprintf("%v (type %v)", value, kind)
+	if len(out) > maxValueStrLen {
+		out = out[:maxValueStrLen] + "..."
+	}
+	return out
+}
+
 // TODO(mostrowski): move below funtions into a helper package, or from common
```
robshakir: s/funtions/functions/

ostromart: Done.

PR #35, opened 2017-09-15
Fix non-struct union type handling, add tests.

ostromart:
- Handle the special case where all union elements spanning only one type results in the field being that type rather than an interface.
- Clean up data tree path handling.
- Add validation test cases from sanitized configs.


ytypes/common.go
```diff
@@ -385,21 +385,24 @@ func getJSONTreeValForField(parentSchema, schema *yang.Entry, f reflect.StructFi
...
 
-	return t, true
+	t, ok := tree.(map[string]interface{})
+	if !ok {
+		return nil, false
+	}
+
+	for k, v := range t {
+		if path[0] == stripModulePrefix(k) {
+			ret, ok := getJSONTreeValForPath(v, path[1:])
```
robshakir: Consider collapsing the if within this block to be `if ret, ok := getJSONTreeValForPath(...); ok { }`.

ytypes/common.go
```diff
@@ -768,7 +771,7 @@ func enumStringToIntValue(parent interface{}, fieldName, value string) (int64, e
 	}
 	ft := field.Type()
 	// leaf-list case
-	if ft.Kind() == reflect.Slice{
+	if ft.Kind() == reflect.Slice {
```
robshakir: With this change, it might be worth ensuring that `gofmt -w -s` doesn't make any changes to the existing code.

ytypes/common.go
ostromart: Done.

ytypes/common.go
ostromart: I usually just run -w, this was a mis-format from a previous commit.

PR #36, opened 2017-09-15
Fix pattern matching to consider full string match only.

ostromart:
https://tools.ietf.org/html/rfc6020#section-9.4.6 states that 
"pattern, which takes as an argument a regular expression string, is used to restrict the built-in type "string", or types derived from "string", to values that match the pattern."
Here, "match" is used to mean that the entire string must match. However, in regex terminology, a match means any matching substring e.g. "abc" matches the regex "[a|b]+" with the substring "ab".


robshakir:
`fixYangRegexp` should do this already -- i.e., if we have an input regexp that doesn't have `^` at the beginning and `$` at the ends of it, then we'll add them there. Is there a reason that this approach doesn't work? (There are other fixes therein too)

ostromart:
The issue was with the lack of () around the entire expression, which led to subexpressions incorrectly associating with the leading/trailing ^ / $.

PR #39, opened 2017-09-16
Handle leaf-list of union, add generator binary, add field numbering.

robshakir:
This PR adds:
* Field numbering using `fnv` consistent hashing.
* Handling of `leaf-list` of union which would previously result in an invalid protobuf being generated.
 * A calling binary for the `GenerateProto3` function.


proto_generator/protogenerator.go
```diff
@@ -0,0 +1,122 @@
+// Copyright 2017 Google Inc.
```
ostromart: Is this just a move with no changes? I don't recall if there's much history but it may be worth doing this with git mv.

ygen/codegen.go
```diff
@@ -491,7 +493,10 @@ func (cg *YANGCodeGenerator) GenerateProto3(yangFiles, includePaths []string) (*
 		// entries in the msgNames set to fall within the same package.
 		tp, ok := genProto.Packages[genMsg.packageName]
 		if !ok {
+			pp := strings.Split(genMsg.packageName, ".")
+			np := append(pp[:len(pp)-1], fmt.Sprintf("%s.proto", pp[len(pp)-1]))
```
ostromart: My usual helper comment: consider moving into a small descriptive function.

ygen/protogen.go
```diff
@@ -639,9 +649,23 @@ func safeProtoIdentifierName(name string) string {
 // between 1 and 2^29-1. The values 19,000-19,999 are excluded as these are explicitly
 // reserved for protobuf-internal use by https://developers.google.com/protocol-buffers/docs/proto3.
 func protoTagForEntry(e *yang.Entry) (uint32, error) {
-	// TODO(robjs): Replace this function with the final implementation
-	// once concluded.
-	return 1, nil
+	return fieldTag(e.Path())
```
ostromart: Above comment seems to be duplicated, these two should be consolidated.

ygen/protogen.go
```diff
@@ -639,9 +649,23 @@ func safeProtoIdentifierName(name string) string {
...
+}
+
+// fieldTag takes an input string and calculates a FNV hash for the value. If the
+// hash is in the range 19,000-19,999 or 1-1,000, the input string has _ appended to
+// it and the hash is calculated.
+func fieldTag(s string) (uint32, error) {
```
ostromart: People might wonder why FNV was chosen. I guess because it's simple?

ygen/protogen.go
```diff
@@ -639,9 +649,23 @@ func safeProtoIdentifierName(name string) string {
...
+	h := fnv.New32()
+	if _, err := h.Write([]byte(s)); err != nil {
+		return 0, fmt.Errorf("could not write field path to hash: %v", err)
+	}
+
+	v := h.Sum32() & 0x1fffffff
```
ostromart: add // 2^29-1?
To be squeaky clean I recommend defining it as a package const maxProtoTagValue = 1 << 29 -1 and putting the blurb about range there.

proto_generator/protogenerator.go
robshakir: It's a new file -- we have `generator` and it seemed cleaner to add a new binary for doing the proto generation since there are some different behaviours here (e.g., writing out to a dir vs. writing out files).

ygen/protogen.go
robshakir: I'm planning to publish the document showing the behaviour of different hash functions that I considered here. It's sufficiently simple and it provides good enough hashing to ensure that we do not get field number clashes.

ygen/codegen.go
robshakir: Done, sufficiently helped :-)

ygen/protogen.go
robshakir: De-duplicated the comment. Sorry, this was a refactor for ensuring that the hashing code wasn't duplicated.

ygen/protogen.go
robshakir: It seems like this splits the definition of the allowed ranges across multiple places (shouldn't we then make 19,000-19,999 a const somewhere too? My view here has been only use const where it might be referenced in >1 place, or it's a clear design choice in the package.

robshakir:
@ostromart -- thanks for the review! PTAL at the updates.

ygen/protogen.go
```diff
@@ -635,13 +645,25 @@ func safeProtoIdentifierName(name string) string {
 	return replacer.Replace(name)
 }
 
-// fieldTag returns a protobuf tag value for the entry e. The tag value supplied is
-// between 1 and 2^29-1. The values 19,000-19,999 are excluded as these are explicitly
-// reserved for protobuf-internal use by https://developers.google.com/protocol-buffers/docs/proto3.
+// fieldTag returns a protobuf tag value for the entry e.
```
ostromart: Update this comment too (for protoTagForEntry).

proto_generator/protogenerator.go
```diff
@@ -0,0 +1,122 @@
...
+
+	log "github.com/golang/glog"
+	"github.com/openconfig/goyang/pkg/yang"
+	"github.com/openconfig/ygot/ygen"
+)
+
+var (
+	yangPaths        = flag.String("path", "", "Comma separated list of paths to be recursively searched for included modules or submodules within the defined YANG modules.")
+	compressPaths    = flag.Bool("compress_paths", false, "If set to true, the schema's paths are compressed, according to OpenConfig YANG module conventions.")
+	excludeModules   = flag.String("exclude_modules", "", "Comma separated set of module names that should be excluded from code generation this can be used to ensure overlapping namespaces can be ignored.")
```
ostromart: super-nit: suggest new sentence "...generation. This can be used..."

proto_generator/protogenerator.go
```diff
@@ -0,0 +1,122 @@
...
+	packageName      = flag.String("package_name", "openconfig", "The name of the Proto package that generated messages should belong to as their parent.")
+	enumPackageName  = flag.String("enum_package_name", "enums", "The name of the package within the generated package that should contain global enum definitions.")
+	outputDir        = flag.String("output_dir", "", "The path to which files should be output, hierarchical folders are created for the generated messages.")
+	ignoreCircDeps   = flag.Bool("ignore_circdeps", false, "If set to true, circular dependencies between submodules are ignored.")
+	generateFakeRoot = flag.Bool("generate_fakeroot", false, "If set to true, a fake element at the root of the data tree is generated. By default the fake root entity is named Device, its name can be controlled with the fakeroot_name flag.")
+	fakeRootName     = flag.String("fakeroot_name", "", "The name of the fake root entity.")
```
ostromart: Add "Device" default? Then you can trim the generateFakeRoot comment.

proto_generator/protogenerator.go
robshakir: :+1:

proto_generator/protogenerator.go
robshakir: Done.

ygen/protogen.go
robshakir: Done, thanks, missed this before.

PR #42, opened 2017-09-19
Split common.go into smaller files.

ostromart:
common.go has grown too large. As a first step towards creating separate common packages with public APIs, split this file into logical units:

util_json - JSON handling
util_errs - support for []error type
util_debug - functions to aid in debugging
util_schema - schema related functions
util_types - type conversion between YANG, Go and JSON worlds
util_reflect - reflect helper code (renamed from reflect_utils)


ostromart:
Yes, purely moves. Unfortunately while moves are easy with git, copies seem to be anything but, so some history may be hard to trace.

PR #46, opened 2017-09-20
Add coverage via coveralls.io

robshakir:
```
  * (M) .travis.yml
    - Add trigger for coverage.
  * (A) scripts/coverage.sh
    - Script to run coverage for all packages.
```


robshakir:
@ostromart PTAL, this simply adds a coverage check using coveralls.io to the CI tests that we're running.

PR #48, opened 2017-09-21
Add missing validation handling for enum types in union.

ostromart:
Enum types in unions were not handled. Usually enums are checked by compiler but in the case of union, there may be other types possible, hence enum must be added to the set of possible types to check against (although the actual value need not be checked because the compiler will do this).


robshakir:
Change looks OK, please consider adding tests within the ytypes package for this scenario as well as the overall schema validation.

ostromart:
Done.

PR #49, opened 2017-09-21
Add gNMI PathElem support for render

robshakir:
This change adds support for the new gNMI `PathElem` type as a prefix and path format when rendering gNMI notifications. This change is made in a generally backwards compatible manner, but did require some changes to the public API.


ygen/gogen.go
```diff
@@ -218,6 +218,25 @@ type generatedGoListMethod struct {
...
+	Keys []*yangFieldMap
+}
+
+// yangFieldMap maps a YANG identifier to its Go identifier.
+type yangFieldMap struct {
+	// YANGName is the field's name in the YANG schema.
+	YANGName string
+	// GoName is the field's name in the Go struct.
+	GoName string
+	// IsPtr indicates that the key field is a pointer.
```
ostromart: This seems a bit strange. Is it not possible to infer this from someplace?

ygen/gogen.go
```diff
@@ -930,13 +981,81 @@ func writeGoStruct(targetStruct *yangDirectory, goStructElements map[string]*yan
...
+//	}
+//
+//	return map[string]interface{}{
+//	  "bar": *t.Bar,
+//	  "baz": *t.Baz,
+//	}
+//  }
+func generateGetListKey(buf *bytes.Buffer, s *yangDirectory, nameMap map[string]*yangFieldMap) error {
+	if s.listAttr == nil {
+		return nil
```
ostromart: Is this an error?

ygot/render.go
```diff
@@ -40,16 +40,33 @@ type path struct {
...
+	// to false, the path format used in the pre-0.4.0 format using
+	// a slice of strings as the path elements.
+	UsePathElem bool
+}
+
 // TogNMINotifications takes an input GoStruct and renders it to slice of
 // Notification messages, marked with the specified timestamp. If a parentPath
 // is used, it is used as a prefix path for the notifications returned.
-func TogNMINotifications(s GoStruct, ts int64, prefix []interface{}) ([]*gnmipb.Notification, error) {
+func TogNMINotifications(s GoStruct, ts int64, prefix []interface{}, cfg *GNMINotificationsConfig) ([]*gnmipb.Notification, error) {
```
ostromart: I'm a bit worried about simulating polymorphism with an interface and a bool here :-)
A cleaner way to achieve this would be to define a base type that can hold either type of path and some methods to emit or transform the path into something the helper functions are expecting (rather than passing the bool all the way down the chain).

ygot/render.go
```diff
@@ -177,14 +205,84 @@ func findUpdatedLeaves(leaves map[*path]interface{}, s GoStruct, parentPath []in
...
+	newElem.Key = map[string]string{}
+	for kn, k := range km {
+		v, err := keyValueAsString(k)
+		if err != nil {
+			return nil, err
+		}
+		newElem.Key[kn] = v
+	}
+
+	nm := append([]interface{}(nil), mapPath...)
```
ostromart: more idiomatic to use copy here.

ygot/types.go
```diff
@@ -37,6 +37,16 @@ type ValidatedGoStruct interface {
 	Validate() error
 }
 
+// KeyHelperGoStruct is an interface which can be implemented by Go structs
+// that are generated to represent a YANG container or list member that has
+// the corresponding function to retrieve the list keys as a map.
+type KeyHelperGoStruct interface {
+	GoStruct // Must be a valid GoStruct
```
ostromart: Just to avoid linter problems down the road, suggest putting this as // GoStruct blah above.

ygen/gogen.go
robshakir: It isn't, since this struct is only used as input to a template, so the template has no real logic to be able to know whether the strings that it received is a pointer or not. The ideal here is that there is zero logic other than very simple `if`/`else`/`range` in the template as they are terrible to maintain.

ygen/gogen.go
robshakir: It's not -- I'll add a comment. This was taking a pattern that you've suggested before, whereby we perform the checks of whether the function is relevant inside the function. In this case, if the `yangDirectory` was an unkeyed list, or just a container, then we just return with nothing to do.

ygot/render.go
robshakir: That's fair -- let me refactor. I was initially attempting to keep the public API unchanged (since there is code that uses this running elsewhere), but since there was a config public API change, then I think making this some form of typed prefix and within the config would probably be possible. This then allows cleaner handling of the paths as you say.

ygot/render.go
robshakir: Sure - this seemed cleaner since for `copy` we need two lines :-)

ygot/types.go
robshakir: Done.

ygot/render.go
```diff
@@ -96,17 +166,40 @@ func findUpdatedLeaves(leaves map[*path]interface{}, s GoStruct, parentPath []in
 		case reflect.Map:
 			// We need to map each child along with its key value.
 			for _, k := range fval.MapKeys() {
-				// mapPaths can only be one element long for a YANG list.
-				keyval := k.Interface()
-				if _, isEnum := keyval.(GoEnum); isEnum {
-					name, _, err := enumFieldToString(k, false)
+				var childPath *pathWrapper
+				switch parent.pathFormat {
+				case ElementPath:
```
ostromart: farm this switch out to a sub-function?

ygot/render.go
```diff
@@ -177,26 +270,92 @@ func findUpdatedLeaves(leaves map[*path]interface{}, s GoStruct, parentPath []in
...
+func appendgNMIPathElemKey(v reflect.Value, p *pathWrapper) (*pathWrapper, error) {
+	if p == nil {
+		return nil, fmt.Errorf("nil path supplied")
+	}
+
+	if p.pathFormat != PathElemPath {
+		return nil, fmt.Errorf("invalid path type to append keys: %v", p.pathFormat)
+	}
+
+	if len(p.structuredPath) == 0 {
```
ostromart: This sort of check could be consolidated to IsValid() for the gNMIPath type.

ygot/render.go
```diff
@@ -222,31 +381,42 @@ func sliceToScalarArray(v []interface{}) (*gnmipb.ScalarArray, error) {
...
 	}
 
-	pfx := interfacePathAsgNMIPath(prefix)
-	if len(pfx.Element) > 0 {
-		n.Prefix = pfx
-	}
-
-	for p, v := range leaves {
-		path, err := stripPrefix(p.p, prefix)
+	if pfx != nil {
```
ostromart: IsValid()

ygot/render.go
```diff
@@ -222,31 +381,42 @@ func sliceToScalarArray(v []interface{}) (*gnmipb.ScalarArray, error) {
...
+		if pfx.pathFormat != ElementPath && pfx.pathFormat != PathElemPath {
+			return nil, fmt.Errorf("invalid path format specified: %v", pfx.pathFormat)
+		}
 
-		for _, p := range path {
-			if p == nil {
-				log.Infof("leavesToNotifications got nil in path: %v", path)
+		switch {
```
ostromart: You could probably omit this switch and always perform both actions i.e. 
n.Prefix = &gnmipb.Path{
￼ Element: pfx.elementPath,
￼ Elem: pfx.structuredPath,
}

ygot/render.go
```diff
@@ -222,31 +381,42 @@ func sliceToScalarArray(v []interface{}) (*gnmipb.ScalarArray, error) {
...
+		path, err := stripPrefix(p.p, pfx)
 		if err != nil {
 			return nil, err
 		}
 
 		u := &gnmipb.Update{
-			Path: interfacePathAsgNMIPath(path),
+			Path: &gnmipb.Path{},
+		}
+		switch pfx.pathFormat {
```
ostromart: Same here, should be able to omit.

ygot/render.go
```diff
@@ -282,6 +452,39 @@ func leavesToNotifications(leaves map[*path]interface{}, ts int64, prefix []inte
...
 
+// stripPrefix removes the prefix pfx from the supplied path, and returns the more
+// specific path elements of the path. If the prefix is invalid for the path an
+// error is return.
+func stripPrefix(path *pathWrapper, pfx *pathWrapper) (*pathWrapper, error) {
+	if path.pathFormat != pfx.pathFormat {
+		return nil, fmt.Errorf("mismatched path formats in prefix and path, %v != %s", path.pathFormat, pfx.pathFormat)
+	}
+
+	switch path.pathFormat {
```
ostromart: Optional: a nice pattern to follow here is to have stripPrefix() top level, then
if path.IsStringSlicePath() {
return stripPrefixStringSlicePath()
}
return stripPrefixPathElementPath()

ygot/struct_validation_map.go
```diff
@@ -39,27 +41,47 @@ const (
 // structTagsToLibPaths takes an input struct field as a reflect.Type, and determines
 // the set of validation library paths that it maps to. Returns the paths as a slice of
 // empty interface slices, or an error.
-func structTagToLibPaths(f reflect.StructField, parentPath []interface{}) ([][]interface{}, error) {
+func structTagToLibPaths(f reflect.StructField, parentPath *pathWrapper) ([]*pathWrapper, error) {
+	if parentPath.pathFormat != ElementPath && parentPath.pathFormat != PathElemPath {
```
ostromart: IsValid()

ygot/render.go
```diff
@@ -37,19 +36,86 @@ const (
 // path stores the elements of a path for a particular leaf,
 // such that it can be used as a key for maps.
 type path struct {
-	p []interface{}
+	p *pathWrapper
+}
+
+// GNMIPathFormat specifies the type of paths that are to be
+// used in gNMI.
+type GNMIPathFormat int64
```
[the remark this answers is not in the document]
robshakir: I made this change, but I'm not 100% happy here since it seems a little fragile to me that `[]string{} != nil` infers quite so much through the code. Isn't an explicit signal here to the code somewhat more robust? It also helps handle the case whereby a prefix that is supplied is explicitly meant to be `nil`.  thoughts?

ygot/render.go
```diff
@@ -37,19 +36,86 @@ const (
...
+
+const (
+	// Unspecified is the uninitialised value of the GNMIPathFormat enumerated
+	// value.
+	InvalidPathFormat GNMIPathFormat = iota
+	// ElementPath indicates that the gNMI <= 0.3.1 path format should be
+	// utilised, this format consists of a slice of string 'elements' within
+	// each message.
+	ElementPath
```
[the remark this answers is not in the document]
robshakir: Removed entirely.

ygot/render.go
```diff
@@ -37,19 +36,86 @@ const (
...
+	// used. This format consists of a slice of gNMI PathElem messages with
+	// a name and set of keys within them.
+	PathElemPath
+)
+
+// pathWrapper provides a wrapper for gNMI path types, particularly
+// containing the Element-based paths which are used in gNMI pre-0.3.1 and
+// PathElem-based paths which are used in gNMI 0.4.0 and above.
+type pathWrapper struct {
+	// pathFormat stores the format of paths that are expected to be stored.
```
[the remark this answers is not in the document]

ygot/render.go
```diff
@@ -37,19 +36,86 @@ const (
...
+	pathFormat GNMIPathFormat
```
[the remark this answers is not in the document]

ygot/render.go
```diff
@@ -37,19 +36,86 @@ const (
...
+	// elementPath stores a path expressed as a series of scalar elements. On output it is
+	// rendered to a []string which is placed in the gNMI element field.
+	elementPath []string
```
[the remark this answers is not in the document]

ygot/render.go
```diff
@@ -37,19 +36,86 @@ const (
...
+type GNMINotificationsConfig struct {
+	// PathFormat specifies the format of paths to be utilised when
+	// generating notifications.
+	PathFormat GNMIPathFormat
+	// ElementPrefix stores the prefix that should be used within the
+	// Prefix field of the gNMI Notification message expressed as a slice
+	// of strings as per the path definition in gNMI 0.3.1 and below.
+	// If UsePathElem is set to true, this prefix is ignored.
+	ElementPrefix []string
+	// PathElemPrefix stores the prefix that should be used withinthe
```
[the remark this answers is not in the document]

ygot/render.go

ygot/render.go
robshakir: We can't simply do this, since we'll end up with `n.Prefix.Element == []string{}` rather than `n.Prefix.Element == nil` in the case that we have asked for string slice paths, but we can handle this case specifically.

ygot/render.go

ygot/render.go
robshakir: This ended up being relatively low complexity, but I agree this pattern works.

ygot/struct_validation_map.go

ygot/render.go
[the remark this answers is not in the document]
ostromart: I wouldn't think so, provided that you treat this type as a class rather than simple struct (i.e. use methods rather than direct field access). This should be robust assuming that both path types cannot be set at once (which you now check in isValid). If you need to check if a path is nil, I recommend adding a method called isNil rather than checking directly. Then the caller doesn't even have to worry about what type it is (unless they want to and then they can do p.isStringSlicePath() && p.isNil() for example).

ygot/render.go
```diff
@@ -174,30 +238,130 @@ func findUpdatedLeaves(leaves map[*path]interface{}, s GoStruct, parentPath []in
...
+
+	if len(p.pathElemPath) == 0 {
+		return nil, fmt.Errorf("invalid path element path length, can't append keys to 0 length path: %v", p.pathElemPath)
+	}
+
+	np := &gnmiPath{}
+	np.pathElemPath = make([]*gnmipb.PathElem, len(p.pathElemPath))
+	copy(np.pathElemPath, p.pathElemPath)
+	newElem := *np.pathElemPath[len(np.pathElemPath)-1]
+
```
ostromart: Suggestion something like the following to abstract the internals:
np := p.Copy()
newElem := *p.At(p.Len()-1) // or *p.Last() maybe?

ygot/render.go
```diff
@@ -223,33 +387,41 @@ func sliceToScalarArray(v []interface{}) (*gnmipb.ScalarArray, error) {
...
-	for p, v := range leaves {
-		path, err := stripPrefix(p.p, prefix)
-
-		for _, p := range path {
-			if p == nil {
-				log.Infof("leavesToNotifications got nil in path: %v", path)
-			}
+	// If the length of the path was 0, then we leave the values in the
+	// output protobuf message as nil.
+	if len(pfx.stringSlicePath) != 0 || len(pfx.pathElemPath) != 0 {
```
ostromart: pfx.Len() > 0

ygot/render.go
```diff
@@ -223,33 +387,41 @@ func sliceToScalarArray(v []interface{}) (*gnmipb.ScalarArray, error) {
...
+		n.Prefix = &gnmipb.Path{
+			Element: pfx.stringSlicePath,
+			Elem:    pfx.pathElemPath,
```
ostromart: n.Prefix, err := pfx.ToProto() (can include above checks in there)

ygot/render.go
```diff
@@ -223,33 +387,41 @@ func sliceToScalarArray(v []interface{}) (*gnmipb.ScalarArray, error) {
...
 		if err != nil {
 			return nil, err
 		}
 
 		u := &gnmipb.Update{
-			Path: interfacePathAsgNMIPath(path),
+			Path: &gnmipb.Path{},
+		}
+
+		if !pfx.isValid() {
```
ostromart: Not sure why this is here - delete it?

ygot/render.go
```diff
@@ -283,6 +455,40 @@ func leavesToNotifications(leaves map[*path]interface{}, ts int64, prefix []inte
 	return []*gnmipb.Notification{n}, nil
 }
 
+// stripPrefix removes the prefix pfx from the supplied path, and returns the more
+// specific path elements of the path. If the prefix is invalid for the path an
+// error is return.
```
ostromart: It returns an error if the paths are invalid, their types are different, or the prefix does not match the path.

ygot/render.go
```diff
@@ -283,6 +455,40 @@ func leavesToNotifications(leaves map[*path]interface{}, ts int64, prefix []inte
...
+func stripPrefix(path *gnmiPath, pfx *gnmiPath) (*gnmiPath, error) {
+	if path.isStringSlicePath() != pfx.isStringSlicePath() {
```
ostromart: Optional: !path.isSameType(pfx)
It may seem like overkill, but the casual reader does not know that the line above is equivalent for gnmiPath type without digging.

ygot/render.go
```diff
@@ -283,6 +455,40 @@ func leavesToNotifications(leaves map[*path]interface{}, ts int64, prefix []inte
 	return []*gnmipb.Notification{n}, nil
 }
 
+// stripPrefix removes the prefix pfx from the supplied path, and returns the more
+// specific path elements of the path. If the prefix is invalid for the path an
+// error is return.
+func stripPrefix(path *gnmiPath, pfx *gnmiPath) (*gnmiPath, error) {
```
ostromart: suggest making this a method of gnmiPrefix since it deals with internals.

ygot/render.go
```diff
@@ -283,6 +455,40 @@ func leavesToNotifications(leaves map[*path]interface{}, ts int64, prefix []inte
...
+	if path.isStringSlicePath() != pfx.isStringSlicePath() {
+		return nil, fmt.Errorf("mismatched path formats in prefix and path, isElementPath: %v != %v", path.isStringSlicePath(), pfx.isStringSlicePath())
+	}
+
+	if !path.isValid() || !pfx.isValid() {
+		return nil, fmt.Errorf("invalid paths supplied for stripPrefix: %v, %v", path, pfx)
+	}
+
+	if pfx.isStringSlicePath() {
```
ostromart: This could be rewritten generically in terms of At() method e.g. 
for i := 0; i < pfx.Len(); i++ {
  if pfx.At(i) != path.At(i) { // no match }
}
return path.SubPath(i, path.Len())

ygot/render.go
```diff
@@ -519,7 +727,7 @@ func constructJSON(s GoStruct, parentMod string, args jsonOutputConfig) (map[str
 
 		for _, p := range mapPaths {
 			v, ok := value.(map[string]interface{})
-			switch len(p) {
+			switch len(p.stringSlicePath) {
```
ostromart: p.Len(), here and below.

ygot/render.go
```diff
@@ -536,26 +744,17 @@ func constructJSON(s GoStruct, parentMod string, args jsonOutputConfig) (map[str
 					continue
 				}
 			case 1:
-				pelem, ok := p[0].(string)
-				if !ok {
-					errs.Add(fmt.Errorf("could not convert path %v into a string", p))
-					continue
-				}
+				pelem := p.stringSlicePath[0]
```
ostromart: p.At(0), here and below.

ygot/struct_validation_map.go
```diff
@@ -39,27 +41,44 @@ const (
...
-	var mapPaths [][]interface{}
+	var mapPaths []*gnmiPath
 	tagPaths := strings.Split(pathAnnotation, "|")
 	for _, p := range tagPaths {
-		// Make a copy of the existing aprent path so we can append to it without
+		// Make a copy of the existing parent path so we can append to it without
 		// modifying it for future paths.
-		elementPath := make([]interface{}, len(parentPath))
-		copy(elementPath, parentPath)
+		ePath := &gnmiPath{}
```
ostromart: ePath := parentPath.Copy()

ygot/struct_validation_map.go
```diff
@@ -39,27 +41,44 @@ const (
...
+		}
+
 		for _, pp := range strings.Split(p, "/") {
-			if pp != "" {
-				elementPath = append(elementPath, interface{}(pp))
+			// Handle empty path tags.
+			if pp == "" {
+				continue
+			}
+			if ePath.isStringSlicePath() {
```
ostromart: ePath.AppendName(pp)

ygot/struct_validation_map_test.go
```diff
@@ -52,6 +55,142 @@ func errToString(err error) string {
 	return err.Error()
 }
 
+func TestStructTagToLibPaths(t *testing.T) {
```
ostromart: Suggest adding tests specifically for gnmiPath type. This splits up the logic specific to that type vs. tag to lib paths code. Several of the tests below could be moved into that test function.

ygot/render.go
```diff
@@ -497,7 +809,7 @@ func constructJSON(s GoStruct, parentMod string, args jsonOutputConfig) (map[str
 			appendModName = true
 		}
 
-		mapPaths, err := structTagToLibPaths(fType, nil)
+		mapPaths, err := structTagToLibPaths(fType, &gnmiPath{stringSlicePath: []string{}})
 		if err != nil {
```
ostromart: use newStringSlicegNMIPath

ygot/render.go
ostromart: Per discussion, let's make this StringAt() (string, error) and PathElemAt() etc.
You can revisit further abstraction if it makes sense in the future.

ygot/render.go
```diff
@@ -72,6 +93,137 @@ func (g *gnmiPath) isPathElemPath() bool {
...
+	}
+	return &gnmipb.Path{Elem: g.pathElemPath}, nil
+}
+
+// IsSameType returns true if the path supplied is the same type as the
+// receiver.
+func (g *gnmiPath) isSameType(p *gnmiPath) bool {
+	return g.isStringSlicePath() == p.isStringSlicePath()
+}
+
```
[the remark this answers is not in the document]

ygot/render.go
```diff
@@ -72,6 +93,137 @@ func (g *gnmiPath) isPathElemPath() bool {
...
+// stripPrefix removes the prefix pfx from the supplied path, and returns the more
+// specific path elements of the path. It returns an error if the paths are invalid,
+// their types are different, or the prefix does not match the path.
+func (g *gnmiPath) stripPrefix(pfx *gnmiPath) (*gnmiPath, error) {
```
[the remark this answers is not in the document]
robshakir: Yes, I think so.

ygot/render.go
```diff
@@ -51,12 +52,32 @@ type gnmiPath struct {
...
 // isValid determines whether a gnmiPath is valid by determing whether the
 // elementPath and structuredPath are both set or both unset.
 func (g *gnmiPath) isValid() bool {
 	if g.stringSlicePath != nil && g.pathElemPath != nil {
 		return false
 	}
+
+	if g.stringSlicePath == nil && g.pathElemPath == nil {
+		return false
+	}
```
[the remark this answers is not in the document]
robshakir: This needs to be != for the two, if stringSlicePath == nil && pathElemPath == nil then we have a bad path, corrected.

ygot/render.go
```diff
@@ -392,35 +542,23 @@ func leavesToNotifications(leaves map[*path]interface{}, ts int64, pfx *gnmiPath
...
-		}
+	p, err := pfx.ToProto()
+	if err != nil {
+		return nil, err
 	}
+	n.Prefix = p
 
 	for p, v := range leaves {
-		path, err := stripPrefix(p.p, pfx)
+		path, err := p.p.stripPrefix(pfx)
```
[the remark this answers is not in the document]

ygot/struct_validation_map.go

ygot/render.go

ygot/struct_validation_map.go

ygot/struct_validation_map_test.go
robshakir: Done.

PR #50, opened 2017-09-21
Move common utility function to util package.

ostromart:
Create a new package called util containing common code and update references inside ytypes to reflect this. Change is purely mechanical renaming.
This is a partial step towards integrating common code from ytypes and ygen/ygot etc.


robshakir:
Let's add a TODO on test coverage when we've made these changes, just so we ensure that our public-facing APIs are well covered.

ostromart:
Created https://github.com/openconfig/ygot/issues/53

PR #54, opened 2017-09-22
Remove dep on ygot.

ostromart:
Move to pointer util functions inside the test package that uses them to remove ygot dep.


robshakir:
LGTM!

PR #56, opened 2017-09-23
Clean up util/errs and add more unit tests for util package.

ostromart:


util/common_test.go
```diff
@@ -0,0 +1,50 @@
...
+// limitations under the License.
+
+package util
+
+import (
+	"testing"
+)
+
+func TestIsValueNil(t *testing.T) {
+	if !IsValueNil(nil) {
```
robshakir: Any reason not to make this test function table-driven per other tests?

ostromart: The reason is that it's not possible in this case to have an input field to share across the tests. It would have to be cast to interface{} and we would not get the same behavior.

robshakir: Ah, I'd missed that the arguments here weren't `interface{}`. Thanks.

Issue #40, opened 2017-09-18
Protobuf Generation should support generating a Fakeroot


robshakir:
Currently this isn't possible as the proto generator will error when generating the created entity since it has not got a parent.

PR #57, opened 2017-09-25
Fix single-type unions, generating fake root, and incorrect imports.

robshakir:
This is a tidy-up CL that fixes:

* When a single type union was generated, then it could result in a `oneof` that was empty.
* When such a single type union was created, if it were the key of a list, then the wrapper type vs. the scalar type was used.
* Adds support for generating the fake root per #40 
* Ensures that where there is a `leaf-list` of `union` that includes an `identityref` or `enumeration` then the enums package is imported by that message.


ygen/protoelements.go
```diff
@@ -30,9 +45,9 @@ import (
 // can be distinguished from one set to the nil value.
 //
 // TODO(robjs): Add a link to the translation specification when published.
-func (s *genState) yangTypeToProtoType(args resolveTypeArgs, basePackageName, enumPackageName string) (*mappedType, error) {
+func (s *genState) yangTypeToProtoType(args resolveTypeArgs, pargs resolveProtoTypeArgs) (*mappedType, error) {
```
ostromart: Since you are often passing args and pargs, it suggests that resolveProtoTypeArgs could include resolveTypeArgs as a base, though I can't tell for sure if such a relationship makes sense.

robshakir: The reason I chose this approach was that `resolveTypeArgs` is common across Go and Proto, and this style makes it a bit more consistent as to what is proto specific and what isn't. I agree that it could be embedded, but this perhaps breaks this consistency. Any thoughts on whether it should be changed?

ostromart: Hard to say without some live discussion, so feel free to leave a TODO to move ahead.
It depends on the relationship between the two I think. If it doesn't make sense to think of resolveProtoTypeArgs as containing resolveTypeArgs, keep them separate. If you always have a concept of some "resolve args" with a common base and parts that are proto/Go specific, consider embedding because it will keep signatures more compact as you pass things around. I made the comment because it seems that in a bunch of places they are being passed as a package deal.
You can still pass the base resolveTypeArgs in situations that are not proto/Go specific.

robshakir: It keeps the signatures cleaner, but it does have an impact to use.
If we have:

```go
type resolveTypeArgs struct {
   yangType *yang.YangType
   ...
}

type resolveProtoTypeArgs struct {
  resolveTypeArgs
}
```

Then AFAICS, we can never do `someFunction(resolveProtoTypeArgs{yangType: ... })` since we can't initialise an embedded type in a struct literal like that. This means that throughout we lose the succinctness when we need to make a subsequent `yangTypeToProtoType` call.

I added a TODO, let's discuss this at some point going forward.

PR #62, opened 2017-09-27
Refactor and clean up unmarshal list and container to use common code.

ostromart:
This change cleans up and pulls out duplicated code that was shared between unmarshal container and list.


ytypes/container.go
```diff
@@ -100,46 +100,55 @@ func validateContainer(schema *yang.Entry, value ygot.GoStruct) (errors []error)
...
 		return fmt.Errorf("unmarshalContainer got parent type %T, expect struct ptr", parent)
 	}
-	pv := pvp.Elem()
 
+	return unmarshalStruct(schema, parent, jt)
+}
+
+// unmarshalStruct unmarshals a JSON tree into a struct.
+//   schema is the schema of the schema node corresponding to the struct being
+//     unmamshaled into.
```
robshakir: "unmarshalled"

ytypes/container.go
```diff
@@ -100,46 +100,55 @@ func validateContainer(schema *yang.Entry, value ygot.GoStruct) (errors []error)
...
 	if !util.IsValueStructPtr(pvp) {
 		return fmt.Errorf("unmarshalContainer got parent type %T, expect struct ptr", parent)
 	}
-	pv := pvp.Elem()
 
+	return unmarshalStruct(schema, parent, jt)
+}
+
+// unmarshalStruct unmarshals a JSON tree into a struct.
+//   schema is the schema of the schema node corresponding to the struct being
```
robshakir: "schema is the YANG schema of the node corresponding to the struct being unmarshalled into"?

ytypes/list.go
```diff
@@ -424,6 +348,61 @@ func unmarshalList(schema *yang.Entry, parent interface{}, value interface{}) er
 	return nil
 }
 
+// makeKeyForInsert returns a key for inserting a struct newVal into the parent,
+// which must be a map. 
+func makeKeyForInsert(schema *yang.Entry, parentMap interface{}, newVal reflect.Value) (reflect.Value, error) {
```
robshakir: There seems to be enough logic here that we might want to test this function separately. Consider adding coverage (I'm OK for this to be a follow up if you feel it's best?)

ostromart: https://github.com/openconfig/ygot/issues/65
I think the coverage esp in list could use a revisit since the tests are out of date and several did not get fully fleshed out.

ytypes/container.go
ostromart: Done.

ytypes/container.go
ostromart: Done.

PR #64, opened 2017-09-28
Add ability to merge two ValidatedGoStructs

robshakir:
Per discussions - this CL adds the ability to merge two `ValidatedGoStruct` objects of the same type.


ygot/struct_validation_map.go
```diff
@@ -314,3 +314,161 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+			return err
+		}
+		nm.SetMapIndex(k, d)
+	}
+	dstField.Set(nm)
+	return nil
+}
+
+// copySliceField copies srcField into dstField. Both srcField and dstField are
+// reflect.Value structs which represent slices.
```
ostromart: Would you mind expanding the comment to cover what types of inputs are expected/supported and how the function behaves for the expected inputs (e.g. wrt overwriting vs appending)?
Without some explanation I'd assume that this function would recursively merge any duplicated elements in the slice but that doesn't seem to be the behavior here.

ygot/struct_validation_map_test.go
```diff
@@ -872,3 +872,315 @@ func TestMergeStructJSON(t *testing.T) {
...
+			}},
+		},
+		inDst: &copyTest{},
+		wantDst: &copyTest{
+			StructSlice: []*copyTest{{
+				StringField: String("russian-river-pliny-the-elder"),
+			}, {
+				StringField: String("lagunitas-brown-shugga"),
+			}},
+		},
```
ostromart: Case where struct has list and the non-list members are equal but list elements have partial overlap?

ygot/struct_validation_map_test.go
```diff
@@ -872,3 +872,315 @@ func TestMergeStructJSON(t *testing.T) {
...
+	}, {
+		name: "string slice",
+		inSrc: &copyTest{
+			StringSlice: []string{"sierra-nevada-pale-ale", "stone-ipa"},
+		},
+		inDst: &copyTest{},
+		wantDst: &copyTest{
+			StringSlice: []string{"sierra-nevada-pale-ale", "stone-ipa"},
+		},
```
ostromart: Case where the two lists intersect?

ygot/struct_validation_map.go
```diff
@@ -314,3 +314,161 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
 
 	return o, nil
 }
+
+// MergeStructs takes two input ValidatedGoStructs and merges their contents,
+// returning a new ValidatedGoStruct. If the input structs a and b are of
+// different types, an error is returned.
```
ostromart: It may be worth spelling out how the more complex situations involving slices pan out e.g. if structs ptr fields are equal but descendant list members differ.

ygot/struct_validation_map.go
```diff
@@ -314,3 +314,161 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+		return nil, fmt.Errorf("error merging b to new struct: %v", err)
+	}
+
+	return n.Interface().(ValidatedGoStruct), nil
+
+}
+
+// copyStruct copies the fields of srcVal into the dstVal struct in-place.
+func copyStruct(dstVal, srcVal reflect.Value) error {
+	if srcVal.Type() != dstVal.Type() {
```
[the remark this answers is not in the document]
robshakir: ACK, these are always structs, so used `IsValueStruct`. Thanks.

ygot/struct_validation_map.go
robshakir: I've added a TODO per the above on this. Previously it was an oversight not to have made these return errors. PTAL.

ygot/struct_validation_map_test.go
robshakir: Added test that return errors to verify.

ygot/struct_validation_map.go
```diff
@@ -319,3 +321,189 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+
+// copyPtrField copies srcField to dstField. srcField and dstField must be
+// reflect.Value structs which represent pointers.
+func copyPtrField(dstField, srcField reflect.Value) error {
+	if srcField.IsNil() || !srcField.IsValid() {
+		return nil
+	}
+
+	switch srcField.Elem().Kind() {
+	case reflect.Struct:
```
ostromart: if util.IsValueStructPtr(srcField) { // avoid potential panic on .Elem()
 blah
 return nil
} 
// default case

ygot/struct_validation_map.go
```diff
@@ -319,3 +321,189 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+}
+
+// copyInterfaceField copies srcField into dstField. Both srcField and dstField
+// are reflect.Value structs which contain an interface value.
+func copyInterfaceField(dstField, srcField reflect.Value) error {
+	if srcField.IsNil() || !srcField.IsValid() {
+		return nil
+	}
+
+	if srcField.Elem().Kind() != reflect.Ptr {
```
ostromart: maybe add util.IsValueInterfacePtr?
Otherwise at least I suggest checking for .Kind() == reflect.Interface() before .Elem().

ygot/struct_validation_map.go
```diff
@@ -319,3 +321,189 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+		return fmt.Errorf("invalid interface type received: %T", srcField.Interface())
+	}
+
+	switch srcField.Elem().Elem().Kind() {
```
ostromart: Rather than using srcField.Elem().Elem() throughout, suggest the following:
srcField = srcField.Elem().Elem() (after checks above)
then 
if !util.IsValueStruct(srcField) {
 return error
} 
// default.

ygot/struct_validation_map.go
```diff
@@ -319,3 +321,189 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+// an error is returned.
+// TODO(robjs): Implement merging of slice fields when they are populated in the
+// dstField, see https://github.com/openconfig/ygot/issues/74.
+func copySliceField(dstField, srcField reflect.Value) error {
+	if dstField.Len() != 0 {
+		return fmt.Errorf("unimplemented: cannot map slice where destination was set: %v == %v", srcField.Type().Name(), dstField.Interface())
+	}
+
+	switch {
+	case srcField.Type().Elem().Kind() != reflect.Ptr:
```
ostromart: Same suggestions re: Elem() as above.

ygot/struct_validation_map_test.go
```diff
@@ -872,3 +872,385 @@ func TestMergeStructJSON(t *testing.T) {
...
+		inSrc: &copyTest{
+			StringSlice: []string{"stone-and-wood-pacific", "pirate-life-brewing-iipa"},
+		},
+		inDst: &copyTest{
+			StringSlice: []string{"feral-brewing-co-hop-hog", "balter-brewing-xpa"},
+		},
+		wantDst: &copyTest{
+			StringSlice: []string{"stone-and-wood-pacific", "pirate-life-brewing-iipa", "feral-brewing-co-hop-hog", "balter-brewing-xpa"},
+		},
+		wantErr: true, // unimplemented.
```
ostromart: expand comment to "input combination not yet supported: dest slice must be nil at this time."
Also suggest removing wantDst unless you're planning to implement this feature soonish.
Here and below.

ygot/struct_validation_map.go
robshakir: I'd erred on the side of keeping these internal functions doing less validation. Since this is internal and we'd already checked for `reflect.Ptr` before calling this function (it's really there to divide the logic) then this seems to be a little more checking than might be required.

Have updated the function - but can't just do this is we want to avoid the panic, since we need to call `.Elem()` in both cases to get the thing pointed to, so I've added two cases.

ygot/struct_validation_map.go
robshakir: See above for why this wasn't done -- still a bit unsure about whether this is useful code. Have implemented the suggestion (we can just use `IsValueInterface` and then `IsValueStructPtr` I think).

ygot/struct_validation_map.go

ygot/struct_validation_map.go
robshakir: I don't think we can do that here, because we need to check that this thing actually isn't a slice of struct pointers, adopted a method from `util/reflect.go` that makes this clearer.

ygot/struct_validation_map_test.go
robshakir: Done.

ygot/struct_validation_map.go
```diff
@@ -319,3 +321,190 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+		}
+		nm.SetMapIndex(k, d)
+	}
+	dstField.Set(nm)
+	return nil
+}
+
+// copySliceField copies srcField into dstField. Both srcField and dstField are
+// reflect.Value structs which represent slices. If the slice in dstField is populated
+// an error is returned.
```
ostromart: Both srcField and dstField are reflect.Value structs which represent slices ->
Both srcField and dstField must be reflect.Slice kind and contain struct ptr elements?

ygot/struct_validation_map.go
ostromart: Ok, sgtm.

ygot/struct_validation_map.go
robshakir: Updated with similar wording.

PR #66, opened 2017-09-29
Add function to access ΛEnumTypes from generated code.

robshakir:
This CL adds the ability to call ΛEnumTypeMap() on a generated struct to retrieve the set of types that are associated with enums within the generated code file.


ygen/gogen.go
```diff
@@ -563,6 +563,15 @@ var ΛEnumTypes = map[string][]reflect.Type{
 }
 `
 
+	// goEnumTypeMapAccessTemplate provides a teplate to output an accessor
```
ostromart: template

robshakir: Done :+1:

robshakir:
No worries - hopefully this works for you.

PR #67, opened 2017-09-30
Add e2e gNMI Telemetry Example

robshakir:
This CL adds an example for gNMI telemetry using the `ygot` library, and fixes an issue found in that demo.


demo/gnmi_telemetry/gnmi.go
```diff
@@ -0,0 +1,117 @@
...
+	"flag"
+	"fmt"
+	"time"
+
+	log "github.com/golang/glog"
+	"github.com/golang/protobuf/proto"
+	oc "github.com/openconfig/ygot/exampleoc"
+	"github.com/openconfig/ygot/ygot"
+
+	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
```
ostromart: group with oc?

ygot/render.go
```diff
@@ -1231,15 +1219,33 @@ func unionInterfaceValue(v reflect.Value, appendModuleName bool) (interface{}, e
 	case v.Elem().Elem().NumField() != 1:
 		return nil, fmt.Errorf("received a union type which did not have one field, had: %v", v.Elem().Elem().NumField())
 	}
```
ostromart: Consider creating a helper IsInterfaceToStructPtr in utils. 
Also, first two cases imply you can have a ptr to ptr to struct - is this expected?

ygot/render.go
```diff
@@ -1231,15 +1219,33 @@ func unionInterfaceValue(v reflect.Value, appendModuleName bool) (interface{}, e
...
+
+// unionPtrValue returns the value of a union when it is stored as a pointer. The
+// type of the union field is as per the description in unionInterfaceValue. Union
+// pointer values are used when a list is keyed by a union.
+func unionPtrValue(v reflect.Value, appendModuleName bool) (interface{}, error) {
+	switch {
+	case v.Kind() != reflect.Ptr:
+		return nil, fmt.Errorf("received a union pointer type that wasn't a pointer, got: %v", v.Kind())
+	case v.Elem().Kind() != reflect.Struct:
+		return nil, fmt.Errorf("received a union pointer that didn't contain a struct, got: %v", v.Elem().Kind())
```
ostromart: Can use util.IsTypeStructPtr or IsValueStructPtr here.

ygot/render.go
robshakir: No - good catch. Not sure where the first test for ptr came from. :-)

ygot/render.go

ygot/render.go
robshakir: Created `IsValueInterfaceToStructPtr` in the helper library and added a bunch of unit tests for the functionality therein.

ygot/render.go
```diff
@@ -1249,25 +1237,41 @@ func constructJSONSlice(field reflect.Value, parentMod string, args jsonOutputCo
...
+	if !util.IsValueInterfaceToStructPtr(v) {
 		return nil, fmt.Errorf("received a union type which was invalid: %v", v.Kind())
-	case v.Elem().Kind() != reflect.Ptr:
-		return nil, fmt.Errorf("received a union type which was not a pointer: %v", v.Kind())
-	case v.Elem().Elem().Kind() != reflect.Struct:
-		return nil, fmt.Errorf("received a union type that did not contain a struct: %v", v.Kind())
-	case v.Elem().Elem().NumField() != 1:
+	}
+
+	if !util.StructValueHasNFields(v.Elem().Elem(), 1) {
```
ostromart: nit: to avoid doing .Elem().Elem() multiple times, do v = v.Elem().Elem() // now a struct

util/reflect.go
```diff
@@ -95,6 +101,12 @@ func PtrToValue(value reflect.Value) reflect.Value {
 	return value
 }
 
+// StructValueHasNFields returns true if the reflect.Value representing a struct
+// v has n fields.
+func StructValueHasNFields(v reflect.Value, n int) bool {
```
ostromart: The functions in util.reflect are meant to be panic free - suggest renaming this to IsValueStructWithNFields if you want to keep the bool return value and add check for struct.

robshakir: Done.

ygot/render.go

PR #68, opened 2017-09-30
Add handling for unmarshalling unions with enums.

ostromart:
Unions with enums depend on new type resolution in the generated code. 
Enum types are strings in JSON source tree. To determine whether they can be represented as enums, the list of all possible enum types and their values is needed from the local schema context. 
If any of these values match, the field is cast as the corresponding type that implements the union interface. Otherwise, unmarshal is attempted into the other, non-enum types.

### DIFF
--- a/ytypes/leaf.go
+++ b/ytypes/leaf.go
@@ -25,7 +25,6 @@ import (
 	log "github.com/golang/glog"
 	"github.com/openconfig/goyang/pkg/yang"
 	"github.com/openconfig/ygot/util"
-	"github.com/openconfig/ygot/ygen"
 	"github.com/openconfig/ygot/ygot"
 )
 
@@ -705,7 +704,7 @@ func schemaToEnumTypes(schema *yang.Entry, t reflect.Type) ([]reflect.Type, erro
 		return nil, fmt.Errorf("%s ΛEnumTypes function returned wrong type %T, want map[string][]reflect.Type", t, ei)
 	}
 
-	util.DbgPrint("path is %s for schema %s", ygen.EntrySchemaPath(schema), schema.Name)
+	util.DbgPrint("path is %s for schema %s", absoluteSchemaDataPath(schema), schema.Name)
 
 	return enumTypesMap[absoluteSchemaDataPath(schema)], nil
 }
--- a/ytypes/leaf.go
+++ b/ytypes/leaf.go
@@ -620,8 +620,8 @@ func setFieldWithTypedValue(parentT reflect.Type, destV reflect.Value, destElemT
 	return nil
 }
 
-// getUnionKindsAndTypes returns the YANG kinds of basic types and and 
-// reflect.Types of enum types that are possible for values for the union with 
+// getUnionKindsAndTypes returns the YANG kinds of basic types and and
+// reflect.Types of enum types that are possible for values for the union with
 // the given schema and type.
 func getUnionKindsAndTypes(schema *yang.Entry, parentT reflect.Type) ([]yang.TypeKind, []reflect.Type, error) {
 	uks, err := getUnionKindsNotEnums(schema)
@@ -654,7 +654,7 @@ func getUnionKindsNotEnums(schema *yang.Entry) ([]yang.TypeKind, error) {
 	return uks, nil
 }
 
-// getUnionTypesNotEnums returns all the non-enum YANG types under the given schema 
+// getUnionTypesNotEnums returns all the non-enum YANG types under the given schema
 // node, dereferencing any refs.
 func getUnionTypesNotEnums(schema *yang.Entry, yt *yang.YangType) ([]*yang.YangType, error) {
 	var uts []*yang.YangType
@@ -687,7 +687,8 @@ func getUnionTypesNotEnums(schema *yang.Entry, yt *yang.YangType) ([]*yang.YangT
 }
 
 // schemaToEnumTypes returns the actual enum types (rather than the interface
-// type) for a given schema, which must be for an enum type.
+// type) for a given schema, which must be for an enum type. t is the type of
+// the containing parent struct.
 func schemaToEnumTypes(schema *yang.Entry, t reflect.Type) ([]reflect.Type, error) {
 	enumTypesMethod := reflect.New(t).Elem().MethodByName("ΛEnumTypeMap")
 	if !enumTypesMethod.IsValid() {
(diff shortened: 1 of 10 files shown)


ytypes/leaf.go
```diff
@@ -508,117 +509,203 @@ the resulting Bgp_Neighbor_RouteReflector would have field
...
-	vf := v.Elem().FieldByName(fieldName)
-	if !vf.IsValid() {
+	// Get the value and type of the field to set, which may have slice or
+	// interface types for leaf-list and union cases.
+	destUnionFieldV := parentV.Elem().FieldByName(fieldName)
+	if !destUnionFieldV.IsValid() {
 		return fmt.Errorf("%s is not a valid field name in %T", fieldName, parent)
 	}
-	ft, err := getFieldElemType(parent, fieldName)
+	dft, _ := parentT.Elem().FieldByName(fieldName)
```
robshakir: What is the `_` here? AFAICS `FieldByName` only returns one `reflect.Value`?

ytypes/leaf.go
```diff
@@ -508,117 +509,203 @@ the resulting Bgp_Neighbor_RouteReflector would have field
...
 }
 
-func getFieldElemType(parent interface{}, fieldName string) (reflect.Type, error) {
-	t := reflect.TypeOf(parent)
-	ft, ok := t.Elem().FieldByName(fieldName)
-	if !ok {
-		return reflect.TypeOf(nil), fmt.Errorf("%s is not a valid field name in %T", fieldName, parent)
+// schemaToEnumTypes returns the actual enum types (rather than the interface
+// type) for a given schema, which must be for an enum type.
+func schemaToEnumTypes(schema *yang.Entry, t reflect.Type) ([]reflect.Type, error) {
```
robshakir: Assume here that `t` must be the struct that contains the enum field? Consider describing this in the comment above.

ytypes/leaf.go
```diff
@@ -508,117 +509,203 @@ the resulting Bgp_Neighbor_RouteReflector would have field
...
-		return ft.Type.Elem().Elem(), nil
+	ei := ec[0].Interface()
+	enumTypesMap, ok := ei.(map[string][]reflect.Type)
+	if !ok {
+		return nil, fmt.Errorf("%s ΛEnumTypes function returned wrong type %T, want map[string][]reflect.Type", t, ei)
 	}
+	util.DbgPrint("path is %s for schema %s", ygen.EntrySchemaPath(schema), schema.Name)
 
-	return reflect.TypeOf(nil), fmt.Errorf("%T is not a valid parent type", parent)
+	return enumTypesMap[ygen.EntrySchemaPath(schema)], nil
```
robshakir: What happens here if `enumTypeMap` doesn't contain an entry with the relevant path? Should this be a non-error case for the function -- I think not. Consider checking whether the key exists and returning an error if it doesn't.

ytypes/leaf_test.go
```diff
@@ -865,6 +895,7 @@ func TestUnmarshalLeaf(t *testing.T) {
 			want: LeafContainerStruct{Uint64Leaf: ygot.Uint64(42)},
 		},
 		{
+			// TODO DEBUG 8
```
robshakir: Remove comment?

ytypes/leaf_test.go
```diff
@@ -880,6 +911,17 @@ func TestUnmarshalLeaf(t *testing.T) {
 			want: LeafContainerStruct{UnionLeaf: &UnionLeafType_Uint32{Uint32: 42}},
 		},
 		{
+			// TODO DEBUG 11
```
robshakir: Remove comment?

ytypes/leaf.go
ostromart: This is the method for the Type (which doesn't come up in godoc, seems like a rendering bug):
// FieldByName returns the struct field with the given name
        // and a boolean indicating if the field was found.
        FieldByName(name string) (StructField, bool)
We don't need to check ok because we known it exists from the line above.

ytypes/leaf.go
ostromart: Done.

ytypes/leaf.go
ostromart: I had originally intended to handle this separately by returning nil but apparently didn't in the end. Good catch.

ytypes/leaf_test.go
ostromart: Done.

ytypes/leaf_test.go
ostromart: Done.

ytypes/leaf.go
ostromart: It's not an error. The function that calls this aggregates all possible union values, which can include non-enum types. If this returns nil, it just means that the union doesn't contain enum values.

ostromart:
Note that I removed the ygen dependency and moved the schema to path translation to a local function. 
Couple of questions around that:
 - why is choice/case not part of the the path?
 - traverseElementSchemaPath doesn't seem to consider fakeroot i.e. I don't see the fakeroot in the EnumTypes map but there's no check for it in that function. I thought there's a fakeroot element in the schema now? At least I am testing for it in isFakeRoot.
 - maybe consider streamlining the function? The reversing seems unnecessary and the fact that you are converting to absolute path may be more meaningfully expressed by prepending "/" to the resulting string.

robshakir:
1) `choice` and `case` statements generally aren't included in the schema path since there must be uniqueness in the top-level names of the `case` statements of a choice. So this function skips them.
2) In the parsed `yang.Entry` that `goyang/pkg/yang` creates, there is no fake root. This is added by `ygen`. In the `yang.Entry` tree that is serialised for use in `ytypes`, the schema tree *is* included such that that element exists for parsing of the fake root. The tree in `goyang/pkg/yang` has multiple roots (one per module). To get between modules, you need to use the module set that the parsed module was associated with.
3) Not entirely sure how to avoid the reversing. This function starts at the leaf and traverses back to the root -- since we don't know how many parents there are, it seems like pushing these names into a slice and then reversing it is a reasonable strategy? 
3) For the absolute path, you're referencing `goyangPathToSchemaPath` right? `traverseElementSchemaPath` walks up the tree and does provide the module name in the path, so isn't quite an absolute path - `slicePathToString` is not required to be absolute. Only `goyangPathToSchemaPath` is absolute, and is prefixed by `/` by adding `""` to the start of the string slice.

ytypes/leaf.go
robshakir: Ah, thanks -- strange, I looked a GoDoc, weird that it has a bug here.

ytypes/leaf.go
robshakir: Ah, understood.

PR #70, opened 2017-10-02
Enable go vet, clean up vet errors.

ostromart:


.travis.yml
```diff
@@ -8,5 +8,9 @@ before_install:
 install:
         - go get -u github.com/golang/protobuf/protoc-gen-go
         - make install
+script:
+        - go tool vet -composites=false ./ygot
+        - go tool vet ./ygen
+        - go tool vet ./ytypes
```
robshakir: Maybe also add `diff -u <(echo -n) <(gofmt -d -s <dir>)` here? This would let us catch any issues with `gofmt` not having been run before submitting.

ostromart: What's the advantage of doing that? I verified that when vet fails, you see the resulting error in the build output.

robshakir: I was just considering that we'd had a few whitespace issues where `gofmt` had figured these. I'm not sure that `vet` catches these?

PR #73, opened 2017-10-03
Fix issue with marshalling entities at the root which are in different modules to their children.

robshakir:
See this diff:

```diff
                 {
                -  "Cisco-IOS-XR-crypto-sam-cfg:Cisco-IOS-XR-crypto-ssh-cfg:crypto": {
                -    "ssh": {
                +  "Cisco-IOS-XR-crypto-sam-cfg:crypto": {
                +    "Cisco-IOS-XR-crypto-ssh-cfg:ssh": {
```

In the case of the 'crypto' leaf, since it is a different module to its SSH child, it was erroneously rendered. This is based on an assumption in the `ygot` library around where `ygen` path compresses, which was not robust in the case of the fake root and uncompressed schemas. This CL adds a fix, and tests for this case.

### DIFF
--- a/ygot/render.go
+++ b/ygot/render.go
@@ -879,6 +879,7 @@ func constructJSON(s GoStruct, parentMod string, args jsonOutputConfig) (map[str
 				}
 				jsonout[pelem] = value
 			default:
+				var nilParent bool
 				parent := jsonout
 				for i := 0; i < p.Len()-1; i++ {
 					k, err := p.StringElemAt(i)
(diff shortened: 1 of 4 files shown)


robshakir:
@ejbrever PTAL also.

ejbrever:
LGTM, thanks!

ygot/render.go
```diff
@@ -50,6 +50,9 @@ type gnmiPath struct {
 	stringSlicePath []string
 	// pathElemPath stores a path expressed as a series of PathElem messages.
 	pathElemPath []*gnmipb.PathElem
+	// isAbsolute determines whether the stored path is absolute (when set), or relative
+	// when unset.
+	isAbsolute bool
```
ostromart: Now that this is a full class you could get the same functionality by having a nil first element and isAbsolute() method.

ygot/struct_validation_map.go
```diff
@@ -63,6 +63,11 @@ func structTagToLibPaths(f reflect.StructField, parentPath *gnmiPath) ([]*gnmiPa
 			}
 			ePath.AppendName(pp)
 		}
+
+		if len(p) > 0 && p[0] == '/' {
+			ePath.isAbsolute = true
+		}
+
```
ostromart: if you were to use isAbsolute() method, you could omit this and skip only if idx != 0 in the for loop above.

ygot/render.go
robshakir: I didn't take this approach because specifically the encoding `[]string{""}` does not mean `/` in the gNMI path encoding. I decided an explicit signal was better such that the caller could indicate that it was absolute vs. not.

ygot/struct_validation_map.go
robshakir: Agreed -- see above.

robshakir:
Will submit using this approach -- we can revisit if you think that this shouldn't be the case.

PR #75, opened 2017-10-03
Add support for the YANG empty type.

robshakir:
Previously, `ygen` simply represented `empty` leaves (not used in OpenConfig) as a boolean value, this meant that it was not possible to distinguish it from boolean when rendering. This CL adds support for a derived type of `bool` to be used such that it can be distinguished from the bool case, and rendered to `[null]` in JSON.


robshakir:
@ejbrever -- another one for you to take a look at please.

ejbrever:
Looks great, thanks!

Just to check, in my code, I just need to set a leaf equal to "ygot.EmptyTypeName" ?

robshakir:
No worries. You’ll set it to cisco.YANGEmpty(true) — the empty type is
added to the generated code that we output.

r.

ejbrever:
Sounds good, thanks!

ygot/render.go
```diff
@@ -1147,6 +1150,17 @@ func constructJSONValue(field reflect.Value, parentMod string, args jsonOutputCo
 		if args.jType == RFC7951 {
 			value = writeIETFScalarJSON(value)
 		}
+	case reflect.Bool:
+		// A non-pointer field of type boolean is an empty leaf within the YANG schema.
+		// For RFC7951 this is represented as a null JSON array (i.e., [null]), if the
+		// case that this leaf is present in internal JSON, then it is rendered as 'true'
+		// only if the leaf is set, otherwise it is nil.
```
ostromart: For RFC7951 this is represented as a null JSON array (i.e., [null]). 
For internal JSON if the leaf is present and set it is rendered as 'true', or as nil otherwise.
(Not sure if that's the correct description, but something like that...)

PR #77, opened 2017-10-03
Add support for validating empty schema nodes.

robshakir:
Required to allow empty leaves to be validated in the GoStructs.


ytypes/leaf.go
```diff
@@ -62,6 +62,12 @@ func validateLeaf(inSchema *yang.Entry, value interface{}) (errors []error) {
 			return util.AppendErr(errors, fmt.Errorf("bad leaf type: expect Int64 for enum type for schema %s, have type %v",
 				schema.Name, ykind))
 		}
+	case reflect.Bool:
+		if ykind != yang.Yempty {
```
ostromart: && ykind != yang.Ybool ?
If that was indeed missing could you please add test case?

ytypes/leaf.go
```diff
@@ -73,7 +79,7 @@ func validateLeaf(inSchema *yang.Entry, value interface{}) (errors []error) {
 		return nil
 		// TODO(mostrowski): restore when representation is decided.
 		//return util.AppendErr(errors, validateBitset(schema, rv))
-	case yang.Ybool:
+	case yang.Ybool, yang.Yempty:
```
ostromart: Since handling is changing, it would be better to create a new function validateEmpty.

ytypes/bool_type.go
```diff
@@ -30,6 +32,14 @@ func validateBool(schema *yang.Entry, value interface{}) error {
 		return err
 	}
 
+	// If the schema is type empty, check that it had the correct type name.
```
ostromart: better to split this to a new function validateEmpty rather than overloading validateBool.

ytypes/leaf.go
robshakir: `yang.Ybool` is expected to be a ptr to a bool, whereas the empty is not since it can't be unset - it's really just there or not.

ytypes/bool_type.go
robshakir: Done.

ytypes/leaf.go
robshakir: Done

PR #80, opened 2017-10-03
Add coverage for utils package.

ostromart:


util/reflect.go
```diff
@@ -23,26 +23,42 @@ import (
...
+	}
 	return t.Kind() == reflect.Map
 }
 
 // IsTypeInterface reports whether v is an interface.
 func IsTypeInterface(t reflect.Type) bool {
+	if t == reflect.TypeOf(nil) {
+		return false
+	}
+	fmt.Printf("type is %v\n", t)
```
robshakir: Remove stale code?

util/reflect.go
```diff
@@ -23,26 +23,42 @@ import (
 
 // IsTypeStructPtr reports whether v is a struct ptr type.
 func IsTypeStructPtr(t reflect.Type) bool {
+	if t == reflect.TypeOf(nil) {
```
robshakir: This is a bit messy - but I don't see a different way to do this. It's a shame `reflect.Type` doesn't have `IsNil` or `IsValid` or somesuch.

util/reflect_test.go
```diff
@@ -86,6 +86,163 @@ func toStringPtr(s string) *string { return &s }
...
+		t.Error("got IsValueNil(slice) true, want false")
+	}
+	if IsValueNil((interface{})(42)) {
+		t.Error("got IsValueNil(interface) true, want false")
+	}
+}
+
+// TODO(mostrowski): check interface types.
+
+func TestIsValueFuncs(t *testing.T) {
```
robshakir: I added a bunch of different tests for the `IsValue...` and `IsType...` in https://github.com/openconfig/ygot/blob/master/util/reflect_test.go#L557 -- can you PTAL since it seems like we have some duplication of testing here?

util/reflect_test.go
```diff
@@ -86,6 +86,163 @@ func toStringPtr(s string) *string { return &s }
...
+		{
+			desc:     "IsValueScalar",
+			function: IsValueScalar,
+			okValues: []interface{}{testInt, &testInt},
+		},
+	}
+
+	for _, tt := range tests {
+		for _, v := range allValues {
+			/*fmt.Printf("check %s with %s ", tt.desc, reflect.TypeOf(v))
```
robshakir: Remove old code?

util/reflect.go
ostromart: Another annoyance in a long list around reflect :-)

util/reflect.go
ostromart: Done.

util/reflect_test.go
ostromart: I've merged the two test sets - I think this format exercises a few more type/function combos. Let me know if something got lost in the process.

util/reflect_test.go
ostromart: Done.

util/reflect_test.go
```diff
@@ -145,6 +143,11 @@ func TestIsValueFuncs(t *testing.T) {
 			okValues: []interface{}{testStruct},
 		},
 		{
+			desc:     "IsValueInterface",
+			function: IsValueInterface,
+			okValues: []interface{}{},
+		},
+		{
```
robshakir: Consider adding something that covers `IsValueInterfaceToStructPtr` here, which is currently uncovered but should be relatively easy to add?

util/reflect_test.go
robshakir: I'm happy with this format -- just figured we didn't necessarily want two sets of tests that covered exactly the same thing.

util/reflect.go
```diff
@@ -124,6 +123,21 @@ func IsValueScalar(v reflect.Value) bool {
...
 
+// IsInterfaceToStructPtr reports whether v is an interface that contains a pointer
+// to a struct.
+func IsValueInterfaceToStructPtr(v reflect.Value) bool {
+	return IsValueInterface(v) && IsValueStructPtr(v.Elem())
+}
+
+// PtrToValue returns the dereferenced reflect.Value of value if it is a ptr, or
+// value if it is not.
+func PtrToValue(value reflect.Value) reflect.Value {
```
robshakir: This is currently not covered in test coverage per https://coveralls.io/builds/13569303/source?filename=util%2Freflect.go

util/reflect_test.go
ostromart: Done.

util/reflect.go
ostromart: I was meaning to take this out, it's not general enough for this package.

PR #83, opened 2017-10-05
Beef up coverage.

ostromart:


ytypes/int_type_test.go
```diff
@@ -307,6 +341,75 @@ func TestValidateInt(t *testing.T) {
...
+			ranges: yang.YangRange{
+				yang.YRange{Min: yang.FromInt(0), Max: yang.FromInt(3)},
+				yang.YRange{Min: yang.FromInt(5), Max: yang.FromInt(10)},
+			},
+			inValues:  []int64{0, 1, 2, 3, 5, 7, 10},
+			outValues: []int64{4, 11},
+		},
+	}
+
+	//yangIntTypes := []yang.TypeKind{yang.Yuint8, yang.Yuint16, yang.Yuint32, yang.Yuint64}
```
robshakir: Uncomment this line and remove the one below? Looks like this one gives wider sets of tests.

ytypes/leaf_test.go
```diff
@@ -896,7 +945,7 @@ func TestUnmarshalLeaf(t *testing.T) {
 			json: `{"uint16-leaf" : 42}`,
 			want: LeafContainerStruct{Uint16Leaf: ygot.Uint16(42)},
 		},
-		{
+		{ // 5
```
robshakir: Remove comment?

ytypes/leaf_test.go
```diff
@@ -921,6 +970,21 @@ func TestUnmarshalLeaf(t *testing.T) {
 			json: `{"enum-leaf" : "E_VALUE_FORTY_TWO"}`,
 			want: LeafContainerStruct{EnumLeaf: 42},
 		},
+		{ // 10
```
robshakir: Same here?

ytypes/util_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"testing"
+
+	"github.com/openconfig/goyang/pkg/yang"
+	"github.com/openconfig/ygot/ygot"
+)
+
+func TestDataSchemaTreesString(t *testing.T) {
+	containerWithLeafListSchema := &yang.Entry{
```
robshakir: Maybe I'm mis-parsing here, but this doesn't have any leaf-lists in it? So should this be `containerWithListSchema`?

ytypes/int_type_test.go
```diff
@@ -72,6 +72,40 @@ func TestValidateIntSchema(t *testing.T) {
...
+		{
+			desc: "bad range Min",
+			schema: &yang.Entry{
+				Name: "bad-range-schema",
+				Type: &yang.YangType{
+					Kind: yang.Yint8,
+					Range: yang.YangRange{
+						yang.YRange{
+							Min: yang.Number{
+								Kind: yang.MaxNumber,
```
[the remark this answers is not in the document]
ostromart: A couple of sides to this. First, the concept being represented: MaxNumber represents infinity, and to say a number must lie between +/- infinity and (same +/-) infinity is currently not permissible. Is this a meaningful statement, conceptually? I don't know but I don't think so. 
Second, what is the effect of actually allowing this in implementation. In practice, if a range were [MaxNumber, MaxNumber] there would not be any number that could fall in this range, because the values in the data tree are just regular numbers, there's no concept of yang.Number there (Max/MinNumber are types, not values). So just for that practical reason I don't think it's a good idea to allow this.

ytypes/int_type_test.go
```diff
@@ -217,7 +251,7 @@ func TestValidateInt(t *testing.T) {
 			inValues:  []int64{-10, -7, -5, -3, -2, 0, 5, 10},
 			outValues: []int64{-11, -4, 11},
 		},
-		{
+		{ // 5
```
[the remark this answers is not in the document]
ostromart: I got tired of counting and decided to leave these markers for posterity in case someone needs to run just one test.

ytypes/int_type_test.go
ostromart: Done.

ytypes/util_test.go
ostromart: Yes, it should. Done.

ytypes/int_type_test.go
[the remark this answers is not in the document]
robshakir: Ah, OK -- I didn't (lazily, sorry) go and check that MaxNumber wasn't a defined actual number. If this is the case, then yes, I agree.

ytypes/int_type_test.go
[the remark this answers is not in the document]
robshakir: Is there some way to do that without changing the test code? I'm OK with leaving them if they have a reason :-).

ytypes/int_type_test.go
```diff
@@ -384,8 +384,8 @@ func TestValidateUint(t *testing.T) {
 		},
 	}
 
-	//yangIntTypes := []yang.TypeKind{yang.Yuint8, yang.Yuint16, yang.Yuint32, yang.Yuint64}
-	yangIntTypes := []yang.TypeKind{yang.Yuint8}
+	yangIntTypes := []yang.TypeKind{yang.Yuint8, yang.Yuint16, yang.Yuint32, yang.Yuint64}
+	//yangIntTypes := []yang.TypeKind{yang.Yuint8}
```
robshakir: Did you want to remove this line?

ytypes/int_type_test.go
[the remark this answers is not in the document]
ostromart: Not that I know of, the issue is that if you want to run just some one test, you can do say 
for _, tt := range[4:5] but there's no self-updating way to mark which number each test is. Anyway, I removed it, I suppose counting practice is a good way to ward off dementia :-)

PR #94, opened 2017-10-14
Adjust protobuf generation to use relative paths for messages.

robshakir:


ygen/protogen.go
```diff
@@ -1059,3 +1061,19 @@ func protoSchemaPathAnnotation(msg *yangDirectory, field *yang.Entry, compressPa
 	b.WriteRune('"')
 	return &protoOption{Name: protoSchemaAnnotationOption, Value: b.String()}, nil
 }
+
+// stripPackagePrefix removes the prefix of pfx from the path supplied. If pfx
+// is not a prefix of path the entire path is returned.
+func stripPackagePrefix(pfx, path string) string {
```
ostromart: It would make the function more generally useful to return whether the prefix was stripped in a bool. Any callers that don't care can ignore that value. Otherwise, consider renaming the function to something like "stripIfHasPrefix".

robshakir: Good point, done! :-)

PR #96, opened 2017-10-17
Fix bug whereby BuildEmptyTree + Merge results in empty struct.

robshakir:
Fix bug found by @ejbrever in `MergeStructs`

```
   - Fix an issue whereby an empty dst struct ptr was always used
     (created with reflect.New) in the case that structs were being
     merged. This means that if BuildEmptyTree had been used on the
     structs being merged, then even empty branches would be
     overwritten.
   - Return an error if a non-struct ptr is non-nil in two different
     structs.
   - Add testing for returned errors.
   - Add testing to validate the BuildEmptyTree issue above.
```

### DIFF
--- a/ygot/struct_validation_map.go
+++ b/ygot/struct_validation_map.go
@@ -400,7 +400,10 @@ func copyStruct(dstVal, srcVal reflect.Value) error {
 }
 
 // copyPtrField copies srcField to dstField. srcField and dstField must be
-// reflect.Value structs which represent pointers.
+// reflect.Value structs which represent pointers. If the source and destination
+// are struct pointers, then their contents are merged. If the source and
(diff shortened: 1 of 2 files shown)


ygot/struct_validation_map.go
```diff
@@ -414,14 +414,28 @@ func copyPtrField(dstField, srcField reflect.Value) error {
 	// Check for struct ptr, or ptr to avoid panic.
 	if util.IsValueStructPtr(srcField) {
 		var d reflect.Value
-		d = reflect.New(srcField.Type().Elem())
+
+		// If the destination value is non-nil, then we maintain its contents
```
ostromart: Suggest adding something about maintaining existing values & overwrite behavior in the function comment.

robshakir:
Thanks!

PR #98, opened 2017-10-18
Beef up coverage.

ostromart:


util/debug_test.go
```diff
@@ -0,0 +1,179 @@
...
+	maxValueStrLen = maxValueStrLenOrig
+}
+
+func TestDbgPrint(t *testing.T) {
+	saveDebugVals()
+	defer restoreDebugVals()
+
+	debugLibrary = true
+	maxCharsPerLine = 10
+	DbgPrint("test debug <this should be truncated>")
```
robshakir: These tests write to stdout still AFAICS? Did you decide they were worth having?

ostromart: The alternatives would be either dependency injection or changing the behavior through a package global var - but it seems to me these would do more harm than good. At least these tests allow us to visually inspect the output, so we can think of them as example code rather than a real unit test.

PR #102, opened 2017-10-20
Fix various nits

ostromart:
- use NewErrs instead of AppendErr(errors, ...
- change signatures to return Errors so that caller can immediately print return value
- make various error strings consistent:
  - wanted error -> want error
  - use "want error?" only for bools
  - use errs where util.Errors are returned 
- other minor cleanups

### DIFF
--- a/util/reflect.go
+++ b/util/reflect.go
@@ -320,7 +320,8 @@ func InsertIntoMapStructField(parentStruct interface{}, fieldName string, key, f
 }
 
 // isFieldTypeCompatible reports whether f.Set(v) can be called successfully on
-// a struct field f corresponding to ft.
+// a struct field f corresponding to ft. It is assumed that f is exported and 
+// addressable.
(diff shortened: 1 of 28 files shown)


util/reflect.go
```diff
@@ -318,19 +319,23 @@ func InsertIntoMapStructField(parentStruct interface{}, fieldName string, key, f
 	return nil
 }
 
+// isFieldTypeCompatible reports whether f.Set(v) can be called successfully on
+// a struct field f corresponding to ft.
```
robshakir: There is `func (reflect.Value) CanSet() bool` -- which checks `CanAddr` and whether the struct field was unexported. I think in our case, there is no such case whereby we can get an unexported field, but perhaps worth checking this too?

ostromart: In this case the field itself is not currently available (only the StructField and value to set the field to) so I'd have to extend the API to pass in the field. It doesn't seem worth it to me given our top level assumption of everything being exported. However, I updated the comments to make these assumptions clear.

Issue #85, opened 2017-10-06
Path compression in schema causes lookup complexity/bugs.


robshakir:
@ostromart 

Let's scheme on this next week. Currently unmarshalling `/components/component/state/type` fails, since because it contains two identityref types within a union. The lookup that we get to is `/component/state/type` because calling `Path()` on the schema tree `yang.Entry` gives the path relative to the fakeroot. This means that we have the same issue where root-level containers that are compressed out of the schema are not reflected in the paths. 

There are two fixes for this - one is demonstrated in #84. This approach says never to use `Path()` on the returned `yang.Entry` in the schematree, and rather use the `schemapath` annotation that's added there. This is a pretty kludgy fix.

The alternate approach is that we create a faithful representation of the schematree as goyang would give us. This means some changes:

 * The schematree could include the modules as root level `yang.Entry` elements.
 * Some entities that were previously compressed out at the root would be reinjected.

Looking at the code, this seems like it'd only really affect the root level entities/handling of the fakeroot. I think it's OK to handle in `ygen` -- but it seems like this has caused us enough trips that we should probably re-examine at this point.

robshakir:
In ygen, we could just serialise something that looks like the module set, but this has complexities because we'd always need to discard these root entries (paths are `/module/rootcontainer/entity`) - alternatively we can find all the entries at the root of all modules, and form some new 'root' that is part of the schematree. It should be transparent to most functions that this happens.

robshakir:
We fixed this in #103.

PR #103, opened 2017-10-24
Restructure path storage in ygot/ygen/ytypes.

robshakir:
Per #85 - the way that path compression was achieved in ygot previously meant that there was complexity for consumers. Particularly around the fake root entity. This CL restructures the path storage in ygot. Apologies for the size, it was difficult to fix and keep the tests passing in a smaller CL.

The design adopted here is that the schema tree *always* generates a hierarchy starting at a 'fake' root `yang.Entry` struct. If this struct corresponds to a generated struct, then its name is appended to that entry - ensuring that we keep the ability to look up the fake root in the schema tree. The entries that are returned are annotated with the `structname` and `schemapath` annotations as per previous implementation.

The changes here are implemented only within `ygen` and `ygot` -- since `ygot` provided the `GzipToSchema` public API which makes this transparent to `ytypes`. The changes in `ytypes` are to remove handling functionality that dealt with the complexity.

The demo BGP program has changes, since this change also means that when serialising e.g., `exampleoc.Bgp` then rather than getting a document which starts as `{"bgp": ... }` we rather return one that is all children of BGP. This means that downstream handlers (particularly the reference google/gnxi implementation) will need to change to reflect this new behaviour. Propose that as we merge this, we give these folks a heads up.


robshakir:
Will take a look at the coverage decrease.

demo/bgp/main_test.go
```diff
@@ -85,6 +86,8 @@ func TestBGPDemo(t *testing.T) {
 			t.Fatalf("TestBGPDemo %s: ioutil.ReadFile(%s/%s): could not open file: %v", tt.name, TestRoot, tt.wantFile, ioerr)
 		}
 
+		fmt.Printf("%v\n", tt.got)
+
```
ostromart: Leftover debug?

ygen/codegen.go
```diff
@@ -409,13 +408,14 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
...
 		var err error
-		if rawSchema, err = serialiseStructDefinitions(goStructs, cg.Config.GenerateFakeRoot, cg.Config.FakeRootName, cg.Config.CompressOCPaths); err != nil {
-			codegenErr.Errors = append(codegenErr.Errors, err)
+		rawSchema, err = buildJSONTree(mdef.modules, cg.state.uniqueDirectoryNames, mdef.directoryEntries["/"])
+		if err != nil {
+			codegenErr.Errors = append(codegenErr.Errors, fmt.Errorf("error marshalling JSON schema: %v", err))
 		}
 
-		if rawSchema != nil {
+		if len(rawSchema) > 0 {
```
ostromart: It's better to fix the method that returns this to return nil if the value has len 0, particularly since the value is used to make decisions.

ygen/codegen.go
```diff
@@ -667,10 +670,10 @@ type mappedYANGDefinitions struct {
 //	- cfg: the current generator's configuration.
 // It returns a mappedYANGDefinitions struct populated with the directory and enum
 // entries in the input schemas, along with the calculated schema tree.
-func mappedDefinitions(yangFiles, includePaths []string, cfg GeneratorConfig) (mappedYANGDefinitions, []error) {
+func mappedDefinitions(yangFiles, includePaths []string, cfg GeneratorConfig) (*mappedYANGDefinitions, []error) {
```
ostromart: If it's not too painful of a change, consider passing cfg by ptr too. Not for performance but rather to stick with convention. Sorry, should have mentioned this earlier.

ygen/codegen.go
```diff
@@ -692,27 +695,41 @@ func mappedDefinitions(yangFiles, includePaths []string, cfg GeneratorConfig) (m
...
+
+	// For all non-excluded modules, we store these to be
+	// used as the schema tree.
+	emm := map[string]bool{}
+	for _, e := range cfg.ExcludeModules {
+		emm[e] = true
+	}
+	ms := []*yang.Entry{}
+	for _, m := range modules {
+		if _, ex := emm[m.Name]; !ex {
```
ostromart: super nit: emm->excluded, ex->ok would be a bit clearer/more conventional.

ygen/schemaparse.go
```diff
@@ -17,9 +17,79 @@ package ygen
...
+		annotateChildren(m, dn)
+		for _, ch := range children(m) {
+			if _, ex := rootEntry.Dir[ch.Name]; ex {
+				return nil, fmt.Errorf("overlapping root children for key %s", ch.Name)
+			}
+			rootEntry.Dir[ch.Name] = ch
+		}
+	}
+
+	if fakeroot != nil {
```
ostromart: Is fakeroot == nil going to be allowed? If so, the function comment should be updated.

ygen/schemaparse.go
```diff
@@ -17,9 +17,79 @@ package ygen
...
+		rootEntry.Name = fakeroot.Name
+		rootEntry.Annotation["schemapath"] = "/"
+		rootEntry.Annotation["structname"] = dn[fakeroot.Path()]
+		rootEntry.Kind = yang.DirectoryEntry
+	}
+	// Always annotate the root as a fake root, so that it is not treated
+	// as a path element in ytypes.
+	rootEntry.Annotation["isFakeRoot"] = true
```
ostromart: I wonder if we should just always have a fakeroot and assume that in ytypes.

ygen/schemaparse.go
```diff
@@ -17,9 +17,79 @@ package ygen
...
+			ch.Annotation["schemapath"] = ch.Path()
+			// Recurse to annotate the children of this entry.
+			annotateChildren(ch, dn)
+		}
+		// In the compressed schema, we remove the human-readable Description
+		// of each child element, since this is not required by any machine
+		// reconstructing the data.
+		ch.Description = ""
+	}
+	e.Description = ""
```
ostromart: e itself doesn't seem to get annotated?

ygot/schema_test.go
```diff
@@ -15,63 +15,88 @@
...
+			if err.Error() != tt.wantErr {
+				t.Errorf("%s: GzipToSchema(%v): got unexpected error, got: %v, want: %v\n", tt.name, tt.in, err, tt.wantErr)
 			}
 			continue
 		}
 
 		if !reflect.DeepEqual(got, tt.want) {
-			t.Errorf("%s: GzipToSchema(%v): did not get expected output, got: %#v, want: %#v", tt.name, tt.in, got, tt.want)
+			gotj, _ := json.MarshalIndent(got, "", strings.Repeat(" ", 4))
+			wantj, _ := json.MarshalIndent(tt.want, "", strings.Repeat(" ", 4))
```
ostromart: may be worthwhile to check for err here - if want is malformed if will lead to a hard to debug error.

ytypes/leaf_test.go
```diff
@@ -1318,3 +1313,130 @@ func TestUnmarshalLeafRef(t *testing.T) {
...
+		wantName: "two",
+	}, {
+		name:     "valid without prefix",
+		inName:   "two",
+		wantName: "two",
+	}, {
+		name:    "invalid input",
+		inName:  "foo:bar:foo",
+		wantErr: "path element did not form a valid name (name, prefix:name): foo:bar:foo",
+	}}
```
ostromart: suggest adding empty string check.

demo/bgp/main_test.go
robshakir: Yes, thanks, removed.

ygen/codegen.go

ygen/codegen.go
robshakir: Good catch, done.

ygen/schemaparse.go
robshakir: It just made for a very big CL to change it in this CL too, so I was planning to make that change in a subsequent one. Updated the function comment.

ygen/schemaparse.go
robshakir: Agree. I suggest that we make this change in the next CL.

ygen/schemaparse.go
robshakir: Fixed this - with the current logic it's not actually required since we always call this with a module entry, which isn't included in the tree, so necessarily all containers get annotated, but just in case we change, and to make this consistent I fixed it.

ygot/schema_test.go
robshakir: True, done.

ytypes/leaf_test.go
robshakir: This isn't an error thus far - but check added.

robshakir:
@ostromart - thanks for the review, PTAL.

Issue #100, opened 2017-10-18
Add support for merging overlapping list entries within a struct.


robshakir:

robshakir:
This was completed in #104.

PR #104, opened 2017-10-25
Add ability to merge two maps that have the same key in both.

robshakir:
This CL adds the ability to merge two maps that have the same key specified in them to the `MergeStructs` functionality.


robshakir:
Re: #100

ygot/struct_validation_map.go
```diff
@@ -506,20 +506,21 @@ func copyMapField(dstField, srcField reflect.Value) error {
...
 	for _, m := range mapsToMap {
 		for _, k := range m.keys {
 			// Check that this key has not already been mapped. We do not support
 			// the case where there are overlapping keys.
-			// TODO(robjs): Implement mapping of entities that exist in both maps.
+			v := m.field.MapIndex(k)
+			var d reflect.Value
 			if _, ok := existingKeys[k.Interface()]; ok {
-				return fmt.Errorf("cannot map element %v, overlaps in both maps", k.Interface())
+				d = existingKeys[k.Interface()]
```
ostromart: if d, ok = existingKeys[k.Interface()]; !ok {
	d = reflect.New(v.Elem().Type())
￼ 	existingKeys[k.Interface()] = v
}

ygot/struct_validation_map_test.go
```diff
@@ -1115,18 +1201,31 @@ func TestCopyStruct(t *testing.T) {
...
 			},
 		},
 		inDst: &copyTest{
 			StructMap: map[copyMapKey]*copyTest{
-				{"fourpure"}: {StringField: String("oatmeal-stout")},
+				{"fourpure"}: {
+					Uint32Field:  Uint32(42),
+					Uint16Field:  Uint16(16),
+					Float64Field: Float64(42.42),
+				},
```
ostromart: Check case where inDst also has StringField. Actually it's unclear what happens in such a case, I'd assume error.

ygot/struct_validation_map.go
```diff
@@ -506,20 +506,21 @@ func copyMapField(dstField, srcField reflect.Value) error {
...
 	}
-	existingKeys := map[interface{}]bool{}
+	existingKeys := map[interface{}]reflect.Value{}
 
 	for _, m := range mapsToMap {
 		for _, k := range m.keys {
 			// Check that this key has not already been mapped. We do not support
 			// the case where there are overlapping keys.
-			// TODO(robjs): Implement mapping of entities that exist in both maps.
+			v := m.field.MapIndex(k)
```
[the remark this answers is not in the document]

ygot/struct_validation_map.go
robshakir: Done, thanks.

ygot/struct_validation_map_test.go
robshakir: Done - PTAL. This is an error, yes, since the same logic is implemented for any struct (overlapping fields that aren't maps are treated as errors).

robshakir:
@ostromart Thanks for the review, PTAL!

ygot/struct_validation_map.go
```diff
@@ -468,10 +468,9 @@ func copyInterfaceField(dstField, srcField reflect.Value) error {
 
 // copyMapField copies srcField into dstField. Both srcField and dstField are
 // reflect.Value structs which contain a map value. If both srcField and dstField
-// are populated, and have non-overlapping keys, they are merged. If a key exists
-// in both the srcField and dstField map, an error is returned.
-// TODO(robjs): Implement merging maps where there are key values taht overlap.
-//  See https://github.com/openconfig/ygot/issues/74.
+// are populated, and have non-overlapping keys, they are merged. If the same
+// key is populated in srcField and dstField, their contents are merged unless
+// there are overlapping fields populated within the map values.
```
ostromart: suggestion: ...dstField, their contents are merged if they do not overlap, or an error is returned if they overlap.? something like that.

robshakir: Done, thanks.

PR #108, opened 2017-11-04
Add nested message generation to protobuf output.

robshakir:
This CL changes the default behaviour of `proto_generator` to output nested messages, rather than per-schematree level packages. This has the advantage of keeping autocompletion working on the generated proto code, and requiring significantly fewer build targets. The `nested_messages` flag to `proto_generator` can be set to `false` to return to the previous behaviour, which is advantageous if one wants packages that can be imported to a particular level of the schema tree.

Apologies for the large change, a lot is testdata.


ygen/protogen.go
```diff
@@ -292,40 +296,162 @@ type protoMsgConfig struct {
...
 	case msg.entry.Parent == nil:
-		return nil, []error{fmt.Errorf("YANG schema element %s does not have a parent, protobuf messages are not generated for modules", msg.entry.Path())}
-	default:
-		// pkg is the name of the protobuf package, if the entry's parent has already
-		// been seen in the schema, the same package name as for siblings of this
-		// entry will be returned.
-		pkg = state.protobufPackage(msg.entry, cfg.compressPaths)
+		return "", fmt.Errorf("YANG schema element %s does not have a parent, protobuf messages are not generated for modules", msg.entry.Path())
+	}
+	// pkg is the name of the protobuf package, if the entry's parent has already
```
ostromart: update comment

ygen/protogen.go
```diff
@@ -292,40 +296,162 @@ type protoMsgConfig struct {
...
+func genProto3MsgCode(pkg string, msgDefs []protoMsg) (*generatedProto3Message, []error) {
 	var b bytes.Buffer
 	imports := map[string]interface{}{}
-	for _, msgDef := range msgDefs {
+	for i, msgDef := range msgDefs {
+		// Sort the child messages into a determinstic order. We cannot use the
+		// package name as a key as it may be the same for multiple packages, therefore
+		// use the code.
+		cmsgs := map[string]*generatedProto3Message{}
+		cstrs := []string{}
```
ostromart: var ...

ygen/protogen.go
```diff
@@ -292,40 +296,162 @@ type protoMsgConfig struct {
...
+		for _, m := range msgDef.ChildMsgs {
+			cmsgs[m.MessageCode] = m
+			cstrs = append(cstrs, m.MessageCode)
+		}
+		sort.Strings(cstrs)
+		nm := []*generatedProto3Message{}
```
ostromart: var...

ygen/protogen.go
```diff
@@ -292,40 +296,162 @@ type protoMsgConfig struct {
...
+			nm = append(nm, cmsgs[c])
+		}
+		msgDef.ChildMsgs = nm
+
 		if err := protoTemplates["msg"].Execute(&b, msgDef); err != nil {
 			return nil, []error{err}
 		}
 		addNewKeys(imports, msgDef.Imports)
+		if i != len(msgDefs)-1 {
+			b.WriteRune('\n')
```
ostromart: Do this write outside the for loop without checking for value of i?

ygen/protogen.go
```diff
@@ -337,17 +463,18 @@ func writeProto3Msg(msg *yangDirectory, msgs map[string]*yangDirectory, state *g
...
 
 	msgDef := protoMsg{
 		// msg.name is already specified to be CamelCase in the form we expect it
 		// to be for the protobuf message name.
-		Name:     msg.name,
-		YANGPath: slicePathToString(msg.path),
-		Enums:    make(map[string]*protoMsgEnum),
+		Name:      msg.name,
+		YANGPath:  slicePathToString(msg.path),
+		Enums:     make(map[string]*protoMsgEnum),
```
ostromart: Keep map init form consistent i.e. either make or {} (btw. the former is preferred per readability comments I've had but I can't find a reference).

ygen/protogen.go
```diff
@@ -337,17 +463,18 @@ func writeProto3Msg(msg *yangDirectory, msgs map[string]*yangDirectory, state *g
 // package for the protobuf message(s) that are being generated, such that relative
 // paths can be used in the messages.
 // TODO(robjs): Split the logic of this function into multiple subfunctions.
-func genProto3Msg(msg *yangDirectory, msgs map[string]*yangDirectory, state *genState, cfg protoMsgConfig, parentPkg string) ([]protoMsg, []error) {
+func genProto3Msg(msg *yangDirectory, msgs map[string]*yangDirectory, state *genState, cfg protoMsgConfig, parentPkg string, childMsgs []*generatedProto3Message) ([]protoMsg, []error) {
```
ostromart: Consider []error -> Errors, here and elsewhere.

ygen/protogen.go
```diff
@@ -395,6 +522,7 @@ func genProto3Msg(msg *yangDirectory, msgs map[string]*yangDirectory, state *gen
 				annotateSchemaPaths: cfg.annotateSchemaPaths,
 				annotateEnumNames:   cfg.annotateEnumNames,
 				parentPackage:       parentPkg,
+				nestedMessages:      cfg.nestedMessages,
```
ostromart: Enough of cfg seems to be going in to consider just embedding the whole type.

ygen/protogen.go
```diff
@@ -403,7 +531,19 @@ func genProto3Msg(msg *yangDirectory, msgs map[string]*yangDirectory, state *gen
 			}
 
 			if keyMsg != nil {
-				msgDefs = append(msgDefs, *keyMsg)
+				if cfg.nestedMessages {
+					// If nested messages are being output, we must ensure that the
+					// generated key message is output within the parent message - hence
+					// it is generated directly here and appended to the child messages.
+					kc, cerrs := genProto3MsgCode(parentPkg, []protoMsg{*keyMsg})
```
ostromart: The fact that *keyMsg is being used suggests that perhaps genProto3MsgCode should take a slice of ptr instead of value?

ygen/protogen.go
```diff
@@ -426,11 +566,17 @@ func genProto3Msg(msg *yangDirectory, msgs map[string]*yangDirectory, state *gen
...
+						}
 					}
 
 					p, _ := stripPackagePrefix(parentPkg, childpkg)
-					pfx = fmt.Sprintf("%s.", p)
+					if !cfg.nestedMessages || msg.isFakeRoot {
+						pfx = fmt.Sprintf("%s.", p)
+					} else {
+						pfx = ""
+					}
```
ostromart: remove else and put pfx = "" above if.

ygen/protogen.go
```diff
@@ -337,17 +463,18 @@ func writeProto3Msg(msg *yangDirectory, msgs map[string]*yangDirectory, state *g
 // package for the protobuf message(s) that are being generated, such that relative
 // paths can be used in the messages.
 // TODO(robjs): Split the logic of this function into multiple subfunctions.
```
ostromart: The situation with this function is becoming dire - it's really hard to keep track of what's going on due to the length. Consider creating an issue if you don't want to address it in this PR.

ygen/protogen.go
```diff
@@ -668,12 +824,18 @@ func protoListDefinition(args protoDefinitionArgs) (*protoMsgListField, *protoMs
...
-			imports:  []string{importPath(args.baseImportPath, args.basePackageName, childPkg)},
+		// repeated field of the list message.
+		if !args.nestedMessages {
+			p := fmt.Sprintf("%s.%s.%s", args.basePackageName, childPkg, listMsgName)
+			p, _ = stripPackagePrefix(fmt.Sprintf("%s.%s", args.basePackageName, args.parentPackage), p)
+			listDef = &protoMsgListField{
+				listType: p,
+			}
+			listDef.imports = []string{importPath(args.baseImportPath, args.basePackageName, childPkg)}
+		} else {
```
ostromart: Consider removing else and putting this as default init.

ygen/protogen.go
```diff
@@ -941,12 +1104,19 @@ func genListKeyProto(listPackage string, listName string, args protoDefinitionAr
...
-	ltype := fmt.Sprintf("%s.%s", p, listName)
-	if listPackage == "" {
-		// Handle the case that the context of the list is already the base package.
+	var ltype string
+	if args.nestedMessages {
+		// Since the protobuf resolution rules mean that the parent scope is
+		// searched, then we do not need to qualify the name of the list message,
+		// even though it is in the parent's namespace.
 		ltype = listName
+	} else {
```
ostromart: Consider refactoring to make this the if case, with ltype := listName above the if.

ygen/yang_helpers.go
```diff
@@ -316,3 +316,47 @@ func listKeyFieldsMap(e *yang.Entry) map[string]bool {
...
+// whether c would be a direct child after schema compression is performed.
+func isDirectEntryChild(p, c *yang.Entry, compressPaths bool) bool {
+	ppp := strings.Split(p.Path(), "/")
+	cpp := strings.Split(c.Path(), "/")
+	dc := isPathChild(ppp, cpp)
+
+	// If we are not compressing paths, the child is not a list entry,
+	// or the child path is not a grandchild
+	// of the current element, then this cannot be a valid direct child.
+	if !compressPaths || !c.IsList() || len(cpp) > len(ppp)+2 || len(cpp) < len(ppp) {
```
ostromart: The +2 here needs an explanation or a helper.

ygen/yang_helpers.go
```diff
@@ -316,3 +316,47 @@ func listKeyFieldsMap(e *yang.Entry) map[string]bool {
...
+		// Can't be a valid child because the parent of the entity doesn't exist
+		// within this container.
+		return false
+	}
+
+	if !hasOnlyChild(ppe) {
+		return false
+	}
+
+	return children(ppe)[0].Path() == c.Path()
```
ostromart: comment why we are guaranteed to have > 0 children here.

ygen/yang_helpers.go
```diff
@@ -316,3 +316,47 @@ func listKeyFieldsMap(e *yang.Entry) map[string]bool {
...
+}
+
+// isPathChild takes an input slice of strings representing a path and determines
+// whether b is a child of a within the YANG schema.
+func isPathChild(a, b []string) bool {
+	if len(b) <= len(a) || len(b) > len(a)+1 {
```
ostromart: It not clear why || len(b) > len(a)+1 ?
If the case of partial match is rare, consider dropping this check altogether since then the loop below would be short.

ygen/protogen.go
```diff
@@ -292,40 +296,162 @@ type protoMsgConfig struct {
...
+// supplied, which is expected to be a top-level message that code generation is
+// being performed for. It takes:
+// 	- msg: the top-level directory definition
+// 	- msgs: the set of message definitions (keyed by path) that are to be output
+//	- state: the current code generation state.
+//  - cfg: the configuration for the current code generation.
+// It returns a generated protobuf3 message.
+func writeProto3MsgNested(msg *yangDirectory, msgs map[string]*yangDirectory, state *genState, cfg protoMsgConfig) (*generatedProto3Message, []error) {
+	var gerrs []error
+	childMsgs := []*generatedProto3Message{}
```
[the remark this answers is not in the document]
robshakir: Done, thanks.

ygen/protogen.go
```diff
@@ -292,40 +296,162 @@ type protoMsgConfig struct {
...
+	gmsg, errs := genProto3MsgCode(pkg, msgDefs)
+	if errs != nil {
+		return nil, append(gerrs, errs...)
+	}
+
+	// Inherit the set of imports that are required for this child. We
+	// skip any that are relative imports as these are only needed for
+	// the case that we have different files per hierarchy level and
+	// are not nesting messages.
+	imports := []string{}
```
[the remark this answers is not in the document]
robshakir: Done, thanks.

ygen/protogen.go
```diff
@@ -292,40 +296,162 @@ type protoMsgConfig struct {
...
+	if msg.isFakeRoot {
+		imports = gmsg.RequiredImports
+	} else {
+		allImports := map[string]bool{}
```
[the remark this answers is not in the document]
robshakir: We can't do `var allImports map[string]bool{}` since this will mean that @364 here, we'll get a panic since we assign to a nil map. Fixed elsewhere though.

ygen/protogen.go
```diff
@@ -292,40 +296,162 @@ type protoMsgConfig struct {
...
+			allImports[i] = true
+		}
+
+		epk := filepath.Join(cfg.baseImportPath, cfg.basePackageName, cfg.enumPackageName, fmt.Sprintf("%s.proto", cfg.enumPackageName))
+		for i := range allImports {
+			if !strings.HasPrefix(i, cfg.baseImportPath) {
+				fmt.Printf("appending %s in %s\n", i, msg.name)
+				imports = append(imports, i)
+			}
+			if _, ok := allImports[epk]; ok {
```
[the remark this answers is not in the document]
robshakir: Thanks -- I'd reverted to a pattern here that I didn't think about, thanks.

ygen/protogen.go

ygen/protogen.go

ygen/protogen.go

ygen/protogen.go
robshakir: I don't think we can do that, can we? Here we're wanting to append if the value is not the last value - so we need to do some check.

ygen/protogen.go

ygen/protogen.go
robshakir: Not sure where this make came from -- consistently have used the latter throughout the library.

ygen/protogen.go
robshakir: Done - and I think this change makes a lot of things neater (reuse of the `protoDefinitionArgs` throughout rather than a lot of repopulation).

ygen/protogen.go
robshakir: Cleaned this all up - since we can do a negative check here and then just leave `pfx` unchanged rather than needing handling in this case.

ygen/protogen.go

ygen/protogen.go
robshakir: Done - yes. Not sure what I was thinking here.

ygen/protogen.go
robshakir: Yeah -  this wasn't cool. I've gone and fixed this, refactored into specific functions. They have some inplace modification - but I think it is clear what's happening. PTAL and LMK if you don't think so. This refactor also changed a bunch of structs that were being passed to ptrs to those structs, and made much better reuse of the config struct, rather than recreating it all over the place.

ygen/yang_helpers.go
robshakir: This was attempted to be covered in `child path is not a granchild`, tried to make this clearer in the comment.

ygen/yang_helpers.go
robshakir: Done - `hasOnlyChild` guarantees this for us.

ygen/yang_helpers.go
robshakir: The case of partial match is the common case, hence adding this in (there are many longer paths than shorter paths, and there are many subsets of paths) - hence keeping in here. If len(b) is greater than a+1, even if the paths match then it can only be a a grandchild. Added a comment.

docs/protobuf_getting_started.md
```diff
@@ -13,11 +13,14 @@ This document provides some background as to how to generate a Protobuf definiti
...
 
-### Filesystem Hierarchy
+The `ygot` package contains a `proto_generator` binary ([source](https://github.com/openconfig/ygot/tree/master/proto_generator)). This binary is used to generate the protocol buffer definitions. Rather than outputting an individual protobuf file with all message definitions within it, this binary has two modes of output:
 
-The `ygot` package contains a `proto_generator` binary ([source](https://github.com/openconfig/ygot/tree/master/proto_generator)). This binary is used to generate the protocol buffer definitions. Rather than outputting an individual protobuf file with all message definitions within it, the generate outputs a hierarchy of files following the schema of the YANG module that is supplied.
+  * **Nested Messages**: In this output mode, one protobuf package per YANG module is output.
+  * **Hierarchical Packages**: In this case, each level of the YANG schema tree hierarchy has an individual protobuf package output for it.
 
-For example,  with the following YANG module:
+The default output mode is to produce nested messages. The `-nested_messages` flag of the `proto_generator` binary can be set to false to produce hierarchical packages.
```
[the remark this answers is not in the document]
ostromart: In general I think it's better because most people think just including a boolean flag makes it true i.e. "--recurse" would be true by default if it's included and false if omitted - see https://gobyexample.com/command-line-flags for a similar example.
But it's probably case by case? i.e. if it means you have to call a flag something silly like "dontRecurse" it's probably better to leave it to true by default. My 2c :-)

ygen/protogen.go
```diff
@@ -37,17 +37,17 @@ const (
...
+	// DefaultEnumPackageName defines the default package name that is
 	// used for the package that defines enumerated types that are
 	// used throughout the schema.
 	DefaultEnumPackageName = "enums"
-	// defaultYwrapperPath defines the default import path for the ywrapper.proto file,
+	// DefaultYwrapperPath defines the default import path for the ywrapper.proto file,
 	// excluding the filename.
 	DefaultYwrapperPath = "github.com/openconfig/ygot/proto/ywrapper"
-	// defaultYextPath defines the default import path for the yext.proto file, excluding
+	// DefaultYextPath defines the default import path for the yext.proto file, excluding
```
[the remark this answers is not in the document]
ostromart: In that case I recommend at least putting them at the top of const block with a blank line to separate them from private - or better put them in a separate const block.

ygen/protogen.go
```diff
@@ -267,9 +270,9 @@ func writeProto3Header(in proto3Header) (string, error) {
 
 // generatedProto3Message contains the code for a proto3 message.
 type generatedProto3Message struct {
-	packageName     string   // packageName is the name of the package that the proto3 message is within.
-	messageCode     string   // messageCode contains the proto3 definition of the message.
-	requiredImports []string // requiredImports contains the imports that are required by the generated message.
+	PackageName     string   // packageName is the name of the package that the proto3 message is within.
+	MessageCode     string   // messageCode contains the proto3 definition of the message.
+	RequiredImports []string // requiredImports contains the imports that are required by the generated message.
```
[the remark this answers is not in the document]
ostromart: Ok - suggest putting a comment to that effect in the first instance this occurs. Also, update comment first char case.

ygen/protogen.go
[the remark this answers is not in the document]
ostromart: I do see this still in other places (e.g. skipFields). Might be worth doing a find over the file to see if anything else is lingering.

ygen/protogen.go
ostromart: yes, you'd still need to check for empty. i guess it's just a personal preference for being slightly easier to read, but your version is fine.

ygen/protogen.go
```diff
@@ -506,27 +584,141 @@ func genProto3Msg(msg *yangDirectory, msgs map[string]*yangDirectory, state *gen
...
+	imports := map[string]interface{}{}
+
+	var pfx string
+	if !(args.cfg.compressPaths && args.directory.isFakeRoot) {
+		childpkg := args.state.protobufPackage(childmsg.entry, args.cfg.compressPaths)
+		// Add the import to the slice of imports if it is not already
+		// there. This allows the message file to import the required
+		// child packages.
+		childpath := importPath(args.cfg.baseImportPath, args.cfg.basePackageName, childpkg)
+		if _, ok := imports[childpath]; !ok {
```
ostromart: if imports[childpath] != nil

ygen/protogen.go
```diff
@@ -506,27 +584,141 @@ func genProto3Msg(msg *yangDirectory, msgs map[string]*yangDirectory, state *gen
...
+			if !args.cfg.nestedMessages || args.directory.isFakeRoot {
+				imports[childpath] = true
```
ostromart: mixing interface{} with bool? maybe make imports map[string]bool above.

ygen/protogen.go
ostromart: Changes look good, this is greatly improved :+1:

ygen/protogen.go
ostromart: Yep, good stuff, this is much more readable.

docs/protobuf_getting_started.md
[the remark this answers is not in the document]
robshakir: Done -- the flag changed here, but I left the switch in the library being the other way around since this is a relatively huge change through that code. I think it's OK to have these two defaults separate.

ygen/protogen.go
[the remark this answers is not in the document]

ygen/protogen.go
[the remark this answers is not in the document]

ygen/protogen.go
robshakir: Left this as is for now.

ygen/protogen.go
robshakir: Done -- `if imports[childpath] == nil` since this is `!ok`.

ygen/protogen.go
robshakir: `stringKeys` below takes `map[string]interface{}` - just to use this common method, it's declared with`interface{}` rather than `bool`.

robshakir:
Thanks again for the review Martin, PTAL - all comments should be addressed :-)

PR #109, opened 2017-11-07
Add utilities for parsing gNMI paths.

robshakir:
This CL open sources utilities for parsing gNMI paths into strings, and vice versa.


ygot/pathstrings.go
```diff
@@ -0,0 +1,259 @@
...
+
+// StringToStringSlicePath takes a string representing a path, and converts it into a
+// gnmi.Path. For example, if the Path "/a/b[c=d]/e" is input, it is converted
+// to a gnmi.Path{Element: []string{"a", "b[c=d]", "e"}} which is returned.
+func StringToStringSlicePath(path string) (*gnmipb.Path, error) {
+	parsedPath := &gnmipb.Path{}
+
+	var inKey, inEscape bool
+	var buf bytes.Buffer
+	for i, ch := range path {
```
ostromart: Suggestion to simplify this function:
pe := strings.Split(path, "/") to get elements
for _, e := range pe {
k, v := ExtractKeyVal(e)
...
}

To get the key-value, do something like:
https://goplay.googleplex.com/p/aiqBFCWjRo
This includes some more error checking, though I'm not sure what the expectations are for the incoming string.

ygot/pathstrings.go
```diff
@@ -0,0 +1,259 @@
...
+		parsedPath.Element = append(parsedPath.Element, buf.String())
+	}
+
+	return parsedPath, nil
+}
+
+// StringToStructuredPath takes a string representing a path, and converts it to
+// a gnmi.Path, using the PathElem element message that is defined in gNMI 0.4.0.
+func StringToStructuredPath(path string) (*gnmipb.Path, error) {
+	parsedPath := &gnmipb.Path{}
```
ostromart: Can refactor this in terms of ExtractKeyValue above.

ygot/pathstrings.go
```diff
@@ -0,0 +1,259 @@
...
+	// the 'elem' field of the gNMI Path message.
+	StructuredPath PathType = iota
+	// StringSlicePath represents a Path using the 'Element' repeated string field
+	// of the gNMI path message.
+	StringSlicePath
+)
+
+// StringToPath takes an input string representing a path in gNMI, and converts
+// it to a gNMI Path message, populated with the specified path encodings.
+func StringToPath(path string, pathTypes ...PathType) (*gnmipb.Path, util.Errors) {
```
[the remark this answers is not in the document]
awly: We specifically want a func that populates both formats at the same time.
That's how we'll handle transition from Element to Elem without breaking any APIs.

This func can be changed to take only one PathType, but then add another value PathBoth (can't come up with better name) that would populate both.

awly: Can we return just error here?

ygot/pathstrings.go
ostromart: btw. line 40 len(kv) != 2 should be an error case, not return nil.

ygot/pathstrings.go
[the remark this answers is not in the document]
ostromart: Does it have to take a PathType? i.e. could we have three functions:
StringToGNMIStringPath
StringToGNMIStructuredPath
StringToGNMIPath 
It's better to have three smaller functions than one big, complex one.

ygot/pathstrings.go
robshakir: We can't do `strings.Split` in this code, because "/" can exist in an escaped path element, so we have to do per-character parsing of the path. I don't see a way to be able to use split since it wouldn't be aware of the escaping?

ygot/pathstrings.go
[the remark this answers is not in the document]
robshakir: I'm not sure I get this -- there are already two functions that do this separately. For example, we have `StringToStringSlicePath` which is essentially `StringToGNMIStringPath` -- the same thing exists for the structured path. Only in the StringToPath public API are both populated, and this is just a wrapper around the two paths.

The reason I added the `...PathType` here, was to avoid `PathBoth`, which seemed klunky compared to just a slice of the path types one wants. They're all public APIs of this package right now.

ygot/pathstrings.go
```diff
@@ -0,0 +1,259 @@
...
+)
+
+// PathToString takes a gNMI Path and provides its string representation. For example,
+// the path Path{Element: []string{"one", "two", "three"} is converted to the string
+// "/one/two/three" and returned. Both the pre-0.4.0 "element"-based paths, and the
+//  >0.4.0 paths based on "elem" are supported. In the case that post-0.4.0 paths are
+// specified, keys that are specified in the path are concatenated onto the name of
+// the path element using the format [name=value]. If the path specifies both pre-
+// and post-0.4.0 paths, the pre-0.4.0 version is returned.
+func PathToString(path *gnmipb.Path) string {
```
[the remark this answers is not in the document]
robshakir: Agreed -- seems I'd missed a few of these cases. Changed this to return `string, error` such that these can be detected.

ygot/pathstrings.go
robshakir: I'm not clear that the function that you propose is actually more robust. How does it handle `/is/i[name=one\=two]` which is one of the test cases -- since in this case, what it'll do is @ line 39 it will split at the first "=". I think that you always need to have the previous character as you're parsing it to determine what the right thing to append is.

This approach if doing character by character parsing and keeping state isn't ideal, but with all the escaping cases that one needs to handle, I think you end up with something akin to this.

ygot/pathstrings.go
robshakir: Sure - `util.Errors` implements `error`.

ygot/pathstrings.go
ostromart: Yeah, I realized some of these issues after seeing a few more possible string examples. However, per offline conversation, I still recommend splitting the processing into multiple steps to simplify the state machine e.g. https://goplay.googleplex.com/p/t2uMm9uyun

robshakir:
Hi @ostromart - PTAL!

ygot/pathstrings.go
```diff
@@ -0,0 +1,268 @@
...
+		return nil, errs
+	}
+
+	return pmsg, nil
+}
+
+// StringToStringSlicePath takes a string representing a path, and converts it into a
+// gnmi.Path. For example, if the Path "/a/b[c=d]/e" is input, it is converted
+// to a gnmi.Path{Element: []string{"a", "b[c=d]", "e"}} which is returned.
+func StringToStringSlicePath(path string) (*gnmipb.Path, error) {
```
ostromart: Mention the special treatment of quotes etc. - e.g. link to the appropriate section in the docs.

ygot/pathstrings.go
```diff
@@ -0,0 +1,268 @@
...
+	parts := pathStringToElements(path)
+	for _, p := range parts {
```
ostromart: Is it possible to put this check inside pathStringToElements and return error there?

ygot/pathstrings.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return gpath, nil
+}
+
+// pathStringToElements splits the string s, which represents a gNMI string
+// path into its constituent elements. It does not parse keys, which are left
+// unchanged within the path - but removes escape characters from element
+// names. The path returned omits any leading empty elements when splitting
+// on the / character.
+func pathStringToElements(s string) []string {
```
ostromart: I think just stripping \ could be dangerous e.g. what to do with \123.

ygot/pathstrings.go
```diff
@@ -0,0 +1,268 @@
...
+func extractKV(in string) (string, map[string]string, error) {
+	var inEscape, inKey, inValue bool
+	var name, currentKey string
+	var buf bytes.Buffer
+	keys := map[string]string{}
+
+	for _, ch := range in {
+		switch {
+		case ch == '[' && !inEscape && !inKey:
+			inKey = true
```
ostromart: Since it would cost little and add a layer of verification I suggest some additional simple checks e.g. here, if inKey -> error. I could be reading this wrong but it looks like the current code would allow [[.

ygot/pathstrings.go
```diff
@@ -0,0 +1,268 @@
...
+		// In this case, we have trailing garbage following the key.
+		return "", nil, fmt.Errorf("trailing garbage following keys in element %s, got: %v", name, buf.String())
+	}
+
+	return name, keys, nil
+}
+
+// addKey adds key k with value v to the key's map. The key, value pair is specified
+// to be for an element named e.
+func addKey(keys map[string]string, e, k, v string) error {
```
ostromart: Should this (or the caller) trim whitespace from k and v?

ygot/pathstrings_test.go
```diff
@@ -0,0 +1,352 @@
...
+		wantErr string
+	}{{
+		name: "root path",
+		in:   &gnmipb.Path{Element: []string{}},
+		want: "/",
+	}, {
+		name: "simple path parts",
+		in:   &gnmipb.Path{Element: []string{"a", "b", "c", "d"}},
+		want: "/a/b/c/d",
+	}, {
```
ostromart: how about no leading /, or trailing /? also empty string.

ygot/pathstrings_test.go
```diff
@@ -0,0 +1,352 @@
...
+		wantStructuredPath: &gnmipb.Path{
+			Elem: []*gnmipb.PathElem{
+				{Name: "foo"},
+				{Name: "bar", Key: map[string]string{"baz": "=bat"}},
+			},
+		},
+	}, {
+		name:              "error - unescaped ] within a key value",
+		in:                `/foo/bar[baz=]bat]`,
+		wantSliceErr:      "error parsing path /foo/bar[baz=]bat]: received null value for key baz of element bar",
```
ostromart: adding specific check in the code would lead to more meaningful error message here.

ygot/pathstrings.go
```diff
@@ -0,0 +1,268 @@
...
+// the path Path{Element: []string{"one", "two", "three"} is converted to the string
+// "/one/two/three" and returned. Both the pre-0.4.0 "element"-based paths, and the
+//  >0.4.0 paths based on "elem" are supported. In the case that post-0.4.0 paths are
+// specified, keys that are specified in the path are concatenated onto the name of
+// the path element using the format [name=value]. If the path specifies both pre-
+// and post-0.4.0 paths, the pre-0.4.0 version is returned.
+func PathToString(path *gnmipb.Path) (string, error) {
+	p := []string{"/"}
+	if path.Element != nil {
+
```
ostromart: remove blank line

ygot/pathstrings.go
```diff
@@ -0,0 +1,268 @@
...
+			for _, k := range keys {
+				elem = fmt.Sprintf("%s[%s=%s]", elem, k, e.Key[k])
+			}
+		}
+		p = append(p, elem)
+	}
+	return filepath.Join(p...), nil
+}
+
+// PathType is used to indicate a gNMI path type.
```
ostromart: suggest moving type to top of file.

ygot/pathstrings.go
```diff
@@ -0,0 +1,268 @@
...
+
+		buf.WriteRune(ch)
+		inEscape = false
+	}
+
+	if buf.Len() != 0 {
+		parts = append(parts, buf.String())
+	}
+
+	if len(parts) > 0 && parts[0] == "" {
```
ostromart: Was reading over gnmi spec and it looks like no trailing / should be an error? Anyway, seems like outputs should somehow be different for with/without leading /.

ygot/pathstrings_test.go
```diff
@@ -0,0 +1,352 @@
...
+		want: "/a/b/c/d",
+	}, {
+		name:    "empty path segment",
+		in:      &gnmipb.Path{Element: []string{"x", "", "y", "z"}},
+		want:    "/x//y/z",
+		wantErr: "nil element at index 1 in [x  y z]",
+	}, {
+		name: "path with attributes",
+		in:   &gnmipb.Path{Element: []string{"q", "r[s=t]", "u"}},
+		want: "/q/r[s=t]/u",
```
ostromart: case of whitespace around = ?

ygot/pathstrings.go
robshakir: I think that `/foo/bar` is completely legal - since this would refer to a leaf. It doesn't make sense to me that `/foo/container/` and `/foo/container` should have different outputs, since in all gNMI operations they mean the same thing. Could you explain why they might/should?

ygot/pathstrings.go
robshakir: I'm not sure that this is a problem if you have `s := "\123gef"` then `range s` will resolve the control characters like this. See: https://play.golang.org/p/2KGZSMpRzH

ygot/pathstrings.go
robshakir: Done.

ygot/pathstrings.go
robshakir: Done.

ygot/pathstrings.go
robshakir: Done -- linked the entire legacy document here.

ygot/pathstrings.go
robshakir: It's possible to, yes, but it means that if `pathStringToElements` is run then `extractKV` (as is done below) then you range the string three times rather than two that we have currently.

ygot/pathstrings.go
robshakir: Added this check and a test.

ygot/pathstrings.go
robshakir: It can't trim from `v`, since if we have `[foo= bar]` then this is literally `foo` being set to ` bar`, and hence trimming from there is an issue.

Since spaces are not valid YANG identifiers then yes, we can check for spaces there.

ygot/pathstrings.go
robshakir: Added a check for unescaped ] in an element name, are there others that you think are missing here? (This is getting very esoteric - since the specification is very loose here.)

ygot/pathstrings.go
ostromart: i was referring to the case of a/b vs /a/b. sorry for the confusion, first trailing should be leading in my comment.

ygot/pathstrings_test.go
robshakir: No leading `/` is covered in "relative path" above.

robshakir: Added trailing /, added empty string.

ygot/pathstrings_test.go
robshakir: Done - added various whitespace tests.

ygot/pathstrings_test.go
robshakir: What is the specific check here though?

1) We are in a key, so ] is valid unescaped because it means the end of the key.
2) We are in the value of a key, where we expect the ] character.
3) We might have additional characters remaining if the next character is [.

So, we need to refactor entirely here again to be able to do this since we need to check that the next character is not a [. Is this worth doing just for a slightly better error message?

robshakir:
Thanks for the review. Submitting now! :tada:

awly:
So, I just realized that this converts to and from plain strings, while gNMI client lib works with []string (https://github.com/openconfig/gnmi/blob/master/client/values.go#L50).
For string -> proto that's not a huge deal, we can just pass strings.Join(p, "/"). But for proto -> string, we can't simply strings.Split it.

Do you think we could add versions of these funcs with []string? I'm happy to do the work, just making sure we agree on the idea.

PR #111, opened 2017-11-08
Add generated protobuf code to repo.

robshakir:
This CL adds generated code to the ygot repo. The current approach has requirements that cannot be met by external build systems (e.g., bazel).


ostromart:
What's the process for updating these and can they drift out of sync with the code?

robshakir:
They're updated by `update.sh` in the same directory. They could drift out of sync, but in testing we run `update.sh` to make sure that they are up-to-date at least during tests. We could create a pre-commit hook that ensures that they are regenerated if you feel it's useful.

I think this is a general issue with generated protos, but unfortunately not generating the code results in a bunch of issues, we can sync up tomorrow IYL.

PR #117, opened 2017-11-16
Add leafref data tree validation.

ostromart:
Previously, leafref validation simply validated that the referring node conformed to the schema at the leafref path, without attempting to compare with the referred-to value. 
This change adds a data tree verification at the root node that compares the actual leafref values.
To accomplish this, a mirror data tree is built from the existing gostruct tree with additional information at each node to allow traversing back up to the parent, which is not possible with gostructs alone and which is required to follow relative leafref paths. 
Then, for each leafref encountered in the traversal, the new tree is passed in which allows traversal back up to the root of the leafref. From this point, a new function traverses down the data tree and returns all data nodes that match the specified key (which selects the entire list when empty).
Currently, while the node selection components supports partial keys, the leafref path parsing does not. Therefore, leafref paths like "current()/../a[k1 = ../path1 and k2 = ../path2]" will lead to a parse error.


util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+			}
 		}
 
-	case IsValueSlice(ni.FieldValue):
-		for i := 0; i < ni.FieldValue.Len(); i++ {
-			nn := *ni
-			nn.FieldValue = ni.FieldValue.Index(i)
-			errs = AppendErrs(errs, forEachFieldInternal(&nn, in, out, iterFunction))
+	case IsTypeSlice(t):
+		schema := *(ni.Schema)
```
robshakir: Consider doing this outside of the `switch` given that it's used in multiple `case` statements.

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+				Schema:         &schema,
+				FieldValue:     reflect.Zero(t.Elem()),
+			}
+			errs = AppendErrs(errs, forEachFieldInternal(nn, in, out, iterFunction))
+		} else {
+			for _, key := range ni.FieldValue.MapKeys() {
+				nn := *ni
+				nn.Schema = &schema
+				nn.Parent = ni
+				nn.PathFromParent = []string{schema.Name}
```
robshakir: Consider making the `PathFromParent` more explicit. Is there a reason that it needs to be `[]string` rather than `[]*gnmipb.PathElem`, since this would allow `FieldKey` to be specified within the actual path specified.

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+				nn.FieldValue = ni.FieldValue.MapIndex(key)
+				nn.FieldKey = key
+				nn.FieldKeys = ni.FieldValue.MapKeys()
```
robshakir: It isn't clear to me why all keys of the map need to be carried in `NodeInfo` here. It doesn't seem like it's ever used:

```
[14:11] github.com/openconfig/ygot   ostromart_dev ✔                     3d ◒  
▶ grep -r "FieldKeys" *
util/reflect.go:	// FieldKeys is the slice of keys in the map being traversed. nil if type
util/reflect.go:	FieldKeys []reflect.Value
util/reflect.go:				nn.FieldKeys = ni.FieldValue.MapKeys()       
```

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+			}
 		}
 	}
 
-	return nil
+	return errs
+}
+
+// GetNodes returns the nodes in the data tree at the indicated path, relative
+// to the supplied root. If for a list node is missing, all values in the list
```
robshakir: If not a list node <X> is missing? I guess you want to say "if none of the keys of a list node are supplied".

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+	}
+	if len(path.GetElem()) == 0 {
+		ResetIndent()
+		return []interface{}{root}, []*yang.Entry{schema}, nil
+	}
+	if schema == nil {
+		return nil, nil, fmt.Errorf("nil schema for data element type %T, remaining path %v", root, path)
+	}
+	// Strip off the absolute path prefix since the relative and absolute paths
+	// are assumed to be equal.
```
robshakir: I don't think this is correct. According to the path specification the root node is encoded as the empty set of elements. Instead, this should be an error if the name of the first element is "".

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
 
-	return nil
+	return errs
+}
+
+// GetNodes returns the nodes in the data tree at the indicated path, relative
+// to the supplied root. If for a list node is missing, all values in the list
+// are returned. If the key is partial, all nodes matching the values present
+// in the key are returned.
+// If the root is the tree root, the path may be absolute.
```
robshakir: Aren't all paths meant to be relative, even if they are relative to the root?

util/reflect.go
robshakir: (NB: goyang's paths don't meet the requirements of the `PathElem` gNMI paths, but they shouldn't ever be supplied as `[]string`, and the way that they schema tree is rebuilt in the supplied code, I don't think that you should ever see `[]string{"", "root-elem"}`.

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+	DbgPrint("GetNode next path %v, value %v", path.GetElem()[0], ValueStr(root))
+
+	switch {
+	case schema.IsContainer() || (schema.IsList() && IsTypeStructPtr(reflect.TypeOf(root))):
+		// Either a container or list schema with struct data node (which could
+		// be an element of a list).
+		return getNodesContainer(schema, root, path)
+	case schema.IsList():
+		// A list schema with the list data node. Must find the element selected
+		// by the path.
```
robshakir: By the keys supplied to the list, I think we always specify the node with the path.

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+		f := v.Field(i)
+		ft := v.Type().Field(i)
+		cschema, err := FieldSchema(schema, ft)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error for schema for type %T, field name %s: %s", root, ft.Name, err)
+		}
+		if cschema == nil {
+			return nil, nil, fmt.Errorf("could not find schema for type %T, field name %s", root, ft.Name)
+		}
+		cschema, err = ResolveLeafRef(cschema)
```
robshakir: It's counter-intuitive to me that you call this here even if the field isn't a leafref -- isn't the pattern of:

```
switch {
   case cschema == nil: 
     // return err
   case isLeafref(cschema):
     var err error
     cschema, err = ResolveLeafRef(cschema)
 }
```

make more sense?

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+	v := rv.Elem()
+
+	for i := 0; i < v.NumField(); i++ {
+		f := v.Field(i)
+		ft := v.Type().Field(i)
+		cschema, err := FieldSchema(schema, ft)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error for schema for type %T, field name %s: %s", root, ft.Name, err)
+		}
+		if cschema == nil {
```
robshakir: Here, you return the target leaf's schema of a leafref -- this seems like it doesn't meet the functionality that one would expect here -- since it means that we might get back a schema that describes a `leaf-list` even if the node that was asked for is a `leaf` (not `leaf-list`) with a leafref.

If it's expected that we get back the target of a leafref schema then we should say so in the function comment - otherwise I think we end up with some unexpected client behaviour.

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+// GetNodes returns an error if the path is not found in the tree, or an element
+// along the path is nil.
+func GetNodes(schema *yang.Entry, root interface{}, path *gpb.Path) ([]interface{}, []*yang.Entry, error) {
+	return getNodesInternal(schema, root, path)
+}
+
+// getNodesInternal is the internal implementation of GetNode. In addition to
+// GetNode functionality, it can accept non GoStruct types e.g. map for a keyed
+// list, or a leaf.
+func getNodesInternal(schema *yang.Entry, root interface{}, path *gpb.Path) ([]interface{}, []*yang.Entry, error) {
```
robshakir: Suggest to add something to the function comment that tells us what the return values are -- I guess the `[]interface{}` is the set of values that the path matches, and `[]*yang.Entry` is the schema(s).

See the comment below about `leafref`. It seems to me that you can return a single value and state that it is a `leaf-list` even though the input schema is a `leaf` of type `leafref`. This might mean that marshalling at the client is wrong (i.e., it should not marshal to a JSON array, but a scalar).

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+}
+
+// GetNodes returns the nodes in the data tree at the indicated path, relative
+// to the supplied root. If for a list node is missing, all values in the list
+// are returned. If the key is partial, all nodes matching the values present
+// in the key are returned.
+// If the root is the tree root, the path may be absolute.
+// GetNodes returns an error if the path is not found in the tree, or an element
+// along the path is nil.
+func GetNodes(schema *yang.Entry, root interface{}, path *gpb.Path) ([]interface{}, []*yang.Entry, error) {
```
robshakir: Another comment on this function signature -- are elements in the`[]interface{}` meant to correspond to those in `[]*yang.Entry`. What happens in the case that we have a path that specifies a path to a `union` of `leafref` (allowable in Y1.1, not Y1.0). In this case, I think I can get multiple `[]*yang.Entry` schemas, do these correspond 1:1 to the entries in `[]interface{}`?

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+				return getNodesInternal(cschema, f.Interface(), trimGNMIPathPrefix(path, p[0:to]))
+			}
+		}
+	}
+
+	return nil, nil, DbgErr(fmt.Errorf("could not find path in tree beyond schema node %s, (type %T), remaining path %v", schema.Name, root, path))
+}
+
+// getNodesList traverses the list root, which must be a map of struct
+// type and matches each map key against the first path element in path. If the
```
robshakir: Against the keys specified in the first `PathElem` of the `Path`?

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+	var matchNodes []interface{}
+	var matchSchemas []*yang.Entry
+
+	// Iterate through all the map keys to see if any match the path.
+	for _, k := range rv.MapKeys() {
+		ev := rv.MapIndex(k)
+		DbgPrint("checking key %v, value %v", k.Interface(), ValueStr(ev.Interface()))
+		match := true
+		if !emptyKey { // empty key matches everything.
+			if !IsValueStruct(k) {
```
robshakir: In the generated code there is a function on each `GoStruct` which represents a list member called [`ΛListKeyMap`](https://github.com/openconfig/ygot/blob/master/ygen/gogen.go#L511) -- this returns `map[string]interface{}` of the keys that are specified within the struct (keyed by schema name). It seems like you might be able to simplify here by calling that function and then just checking map equality (or partial match).

The `Sprintf` below is probably still required however.

util/reflect.go
```diff
@@ -367,61 +413,368 @@ type NodeInfo struct {
...
+		}
+	}
+
+	if len(matchNodes) == 0 {
+		return nil, nil, nil
+	}
+	return matchNodes, matchSchemas, nil
+}
+
+// pathStructTagKey returns the string label of the struct field sf when it is
```
robshakir: This is only true when compression is enabled -- this code is robust to both cases, but the comment isn't entirely true.

util/schema.go
```diff
@@ -0,0 +1,326 @@
...
+// operates on data tree paths.
+func IsChoiceOrCase(e *yang.Entry) bool {
+	if e == nil {
+		return false
+	}
+	return e.IsChoice() || e.IsCase()
+}
+
+// SchemaPaths returns all the paths in the path tag.
+func SchemaPaths(f reflect.StructField) ([][]string, error) {
```
robshakir: We have multiple implementations of this function. I wonder whether we can make them common. We should create an issue.

util/schema.go
```diff
@@ -0,0 +1,326 @@
...
+		m := MatchingNonChoiceCaseSchema(e, path)
+		if m != nil {
+			return m
+		}
+	}
+
+	return nil
+}
+
+// FindFirstNonChoiceOrCase recursively traverses the schema tree and returns a
```
robshakir: Ditto, now this is public, we have multiple implementations of this function (the other is in `ygen`), let's create an issue/TODO to fix that.

util/schema.go
```diff
@@ -0,0 +1,326 @@
...
+	return s
+}
+
+// ResolveLeafRef returns a ptr to the schema pointed to by the provided
+// leaf-ref schema. It returns schema itself if schema is not a leaf-ref.
+func ResolveLeafRef(schema *yang.Entry) (*yang.Entry, error) {
+	if schema == nil {
+		return nil, nil
+	}
+	// TODO(mostrowski): this should only be possible in fakeroot. Add an
```
robshakir: There is already an annotation in the schema for this, see [here](https://github.com/openconfig/ygot/blob/master/ygen/schemaparse.go#L58-L60).

ytypes/util_schema.go
```diff
@@ -228,6 +216,21 @@ func childSchema(schema *yang.Entry, f reflect.StructField) (*yang.Entry, error)
 	return nil, nil
 }
 
+// findFirstNonChoiceOrCase recursively traverses the schema tree and populates
```
robshakir: I think you have two implementations of this even in this code?

util/reflect.go
ostromart: This is something that I find confusing in string to gnmi path conversion (the paths here may be coming from leafref paths that may be absolute). What does /a/b/c vs. a/b/c convert to? If they are the same, how can we recognize an absolute path?

util/reflect.go
ostromart: I think this comment was confusing. Just trying to say that the function we are calling still needs to find the list element we want, using the supplied path (with keys).

util/reflect.go
ostromart: The ResolveLeafRef function description states "It returns schema itself if schema is not a leaf-ref." but I agree this is a bit confusing because at first read it implies we think we have a leafref when calling the function. 
At the same time I don't want to make the caller check if it's a leafref every time - I could maybe rename this to ResolveIfLeafRef, or otherwise what would you suggest?

util/reflect.go
ostromart: sg, thanks.

util/reflect.go
ostromart: If path compression not enabled then I think the element that would've been compressed (e.g. config/name) cannot be part of the key anymore because it's now a container or list. In that case pathToSchema would just return "name" (if that is part of the key). I may be missing your point - let's talk about it live.

util/schema.go
ostromart: https://github.com/openconfig/ygot/issues/120

util/schema.go
ostromart: Grouped in comment with the same issue since these are related.

util/schema.go
ostromart: Done.

ytypes/util_schema.go
ostromart: Done.

util/reflect.go
ostromart: Per discussion, I've updated the comment and removed resolving the leafref. For the case of returning a leaf-list, the list itself is considered a single node and the caller must be prepared to iterate through it. The schema at the same index will be the leaf-ref schema (which is shared with the leaf-list entries).

util/reflect.go
ostromart: After our discussion and sleeping on this a bit I'm wondering if we should be insisting that gnmi path spec be followed rigidly here (because we are using it for convenience, rather than in its original domain). Maybe we could state somewhere that we expect both absolute and relative paths as gpb.Path and the former start with ""?

util/reflect.go
ostromart: Removed.

util/reflect.go
ostromart: Changed to ResolveIfLeafRef per offline discussion.

util/reflect.go
ostromart: Comparing the code required to reflect into ΛListKeyMap it doesn't look much better than what's here and it probably makes the logic even harder to understand for a casual reader (they would have to know what ΛListKeyMap contains etc.)

util/reflect.go
robshakir: I just wonder whether this will cause confusion elsewhere -- since for `GetNode` etc. you'd expect that these paths that are prefixed with "" are not valid, I think. I'm fine with not requiring the path specification to be followed rigidly, if there's a good reason to do this and we can make it clear that these paths are only internal.

Looking at this again, I wasn't (and probably forgot from our previous discussion) entirely clear where the absolute vs. relative case is interesting in this fn?

util/reflect.go
robshakir: Sure - your choice :-)

util/reflect.go
robshakir: Sorry, yes this was a nit that I was just pointing out. Feel free to ignore.

robshakir:
@ostromart (not sure whether you meant to change to @maoghub :-)) -- is this ready for me to review the test code? Did we say we'd fix the coverage % in a subsequent CL?

util/reflect.go
ostromart: We are converting a leafref path to gnmi and this can be absolute or relative. Since absolute == relative if root is the tree base the code/comment here is just saying if you're passing in an absolute path, I hope you are passing the root element along with it.

util/debug_test.go
```diff
@@ -177,3 +177,68 @@ func TestSchemaTreeString(t *testing.T) {
...
+	}
+
+	got := DataSchemaTreesString(containerWithListSchema, container)
+	want := `     [container (container)]
+    KeyList [key-list (list)]
+    keyval1
+      Key : "keyval1" [key (leaf)]
+      LeafField : 42 [leaf-field (leaf)]
+`
+	if got != want {
```
robshakir: Consider using `pretty.Compare` or similar to generate a diff here? Seems like it might be easier to spot what the diff is in the case that the tests fail.

util/reflect_test.go
```diff
@@ -771,31 +849,64 @@ func TestForEachField(t *testing.T) {
 }
 
 func TestUpdateFieldUsingForEachField(t *testing.T) {
+	containerSchema := &yang.Entry{
```
robshakir: Can this schema be made general for the file such that each test function doesn't need to define the same schema and the associated types that represent its generated code?

util/reflect_test.go
```diff
@@ -847,3 +958,399 @@ func TestStructValueHasNFields(t *testing.T) {
...
+type ListElemStruct1 struct {
+	Key1   *string              `path:"key1"`
+	Outer  *OuterContainerType1 `path:"outer"`
+	Outer2 *OuterContainerType1 `path:"outer2"`
+}
+type ContainerStruct1 struct {
+	StructKeyList map[string]*ListElemStruct1 `path:"config/simple-key-list"`
+}
+
+func (*InnerContainerType1) IsYANGGoStruct() {}
```
robshakir: I think conventionally these should follow the definition of the type - rather than be in a block like this?

util/reflect_test.go
```diff
@@ -847,3 +958,399 @@ func TestStructValueHasNFields(t *testing.T) {
...
+						Name: "leaf-field",
+					},
+				},
+			},
+			want: []interface{}(nil),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
```
robshakir: For my education -- it seems like here you could run the anonymous function here directly within the test - with the only difference that it wouldn't run in its own goroutine. Is there another motivating factor for using `t.Run` here?

util/reflect_test.go
```diff
@@ -847,3 +958,399 @@ func TestStructValueHasNFields(t *testing.T) {
...
+type ListElemStruct2 struct {
+	Key1  *string              `path:"key1"`
+	Key2  *int32               `path:"key2"`
+	Outer *OuterContainerType2 `path:"outer"`
+}
+type ContainerStruct2 struct {
+	StructKeyList map[KeyStruct2]*ListElemStruct2 `path:"struct-key-list"`
+}
+
+func (*InnerContainerType2) IsYANGGoStruct() {}
```
robshakir: Same comment re: ordering of these functions with their receiver type.

util/reflect_test.go
```diff
@@ -847,3 +958,399 @@ func TestStructValueHasNFields(t *testing.T) {
...
+						Name: "outer",
+					},
+					{
+						Name: "inner",
+					},
+				},
+			},
+			want: []interface{}{c1.StructKeyList[KeyStruct2{"forty-two", 42}].Outer.Inner},
+		},
+	}
```
robshakir: Would be good to test some of the failure cases here -- i.e., mismatched keys within the list, partial key matches, leaves that do not exist.

util/reflect_test.go
```diff
@@ -847,3 +958,399 @@ func TestStructValueHasNFields(t *testing.T) {
...
+						Name: "inner",
+					},
+					{
+						Name: "leaf-field",
+					},
+				},
+			},
+			want: []interface{}(nil),
+		},
+	}
```
robshakir: Test the case of a key that is not present within the map?

util/reflect_test.go
```diff
@@ -847,3 +958,399 @@ func TestStructValueHasNFields(t *testing.T) {
...
+									"inner": {
+										Name: "inner",
+										Kind: yang.DirectoryEntry,
+										Dir: map[string]*yang.Entry{
+											"leaf-field": {
+												Name: "leaf-field",
+												Kind: yang.LeafEntry,
+												Type: &yang.YangType{
+													Kind: yang.Yleafref,
+													Path: "../../config/inner/leaf-field",
```
robshakir: Consider also testing absolute paths here.

ytypes/leafref_test.go
```diff
@@ -0,0 +1,432 @@
...
+						},
+					},
+					"int32-ref-to-list": {
+						Name: "int32-ref-to-list",
+						Kind: yang.LeafEntry,
+						Type: &yang.YangType{
+							Kind: yang.Yleafref,
+							Path: "../../list[key = current()/../../key]/int32",
+						},
+					},
```
robshakir: Consider adding test cases for:

* `leaf-list` containing `leafref`.
* `typedef` that defines a `leafref`.
* (for Y1.1) `union` that defines a `leafref`.

ytypes/leafref_test.go
```diff
@@ -0,0 +1,432 @@
...
+					1: {Int32(1), Int32(42)},
+					2: {Int32(2), Int32(43)},
+				},
+				Key:        Int32(3),
+				Container2: &Container2{LeafRefToList: Int32(43)},
+			},
+			wantErr: `pointed-to value with path ../../list[key = current()/../../key]/int32 from field LeafRefToList value 43 (int32 ptr) schema /int32-ref-to-list is empty set`,
+		},
+	}
+	for _, tt := range tests {
```
robshakir: Linebreak here (nit).

ytypes/leafref_test.go
```diff
@@ -0,0 +1,432 @@
...
+		{
+			desc:       "quotes",
+			in:         `b[key="[=value=]"]`,
+			wantPrefix: "b",
+			wantKey:    "key",
+			wantValue:  `"[=value=]"`,
+		},
+		{
+			desc:       "path",
+			in:         "b[key = ../a/b/c]",
```
robshakir: Unquoted keys aren't valid XPATH - so wouldn't ever be seen here, IMHO, this should return an error.

ytypes/schema_tests/testdata/bgp-example.json
```diff
@@ -1,212 +1,248 @@
...
-         }
+        {
+          "config": {
+            "name": "LOOPBACK-TO-BGP"
+          },
+          "name": "LOOPBACK-TO-BGP"
+        },
+        {
+          "config": {
+            "name": "METRO-AR"
```
robshakir: Please anonymise these policy names a bit :-)

util/debug_test.go
ostromart: For small protos I find dumping out the whole thing makes the diffs easier to spot. I agree though once it's more than a dozen lines or so, pretty's the way to go.

util/reflect_test.go
ostromart: Done. I've put it above the tests to keep some locality, otherwise the tests become very hard to understand.

util/reflect_test.go
ostromart: Done.

ytypes/schema_tests/testdata/bgp-example.json
ostromart: Done.

ytypes/leafref_test.go
ostromart: Done.

ytypes/leafref_test.go
ostromart: Created https://github.com/openconfig/ygot/issues/124

ytypes/leafref_test.go
ostromart: Done.

util/reflect_test.go
ostromart: Done.

robshakir:
Apologies for the delay. LGTM. Let's fix the coverage so we get our green logo back!

PR #121, opened 2017-11-22
Add ygot.PathToStrings

awly:
This is same as PathToString, but returns separated path elements in
a slice.


awly:
I also switched from path/filepath to path, since filepath produces OS-specific results (and is meant for files).

PR #122, opened 2017-11-27
Remove leading "/" from PathToStrings

awly:
The "/" is only needed in formatted string from PathToString.

Also add a short sanity unit test for PathToStrings, more detailed cases
are already covered via TestPathToString.

### DIFF
--- a/ygot/pathstrings.go
+++ b/ygot/pathstrings.go
@@ -43,7 +43,7 @@ const (
 // representing the path.
 func PathToString(path *gnmipb.Path) (string, error) {
 	s, err := PathToStrings(path)
-	return stdpath.Join(s...), err
+	return "/" + stdpath.Join(s...), err
 }
 
 // PathToStrings takes a gNMI Path and provides its string representation. For example,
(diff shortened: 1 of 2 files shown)


ygot/pathstrings.go
```diff
@@ -43,7 +43,7 @@ const (
 // representing the path.
```
robshakir: Suggest that we add a function comment here that says that the supplied `path` is always parsed as absolute and therefore should be the concatenation of the `prefix` plus `path` if a `prefix` exists.

ygot/pathstrings_test.go
```diff
@@ -109,6 +110,22 @@ func TestPathToString(t *testing.T) {
 	}
 }
 
+func TestPathToStrings(t *testing.T) {
+	in := &gnmipb.Path{Elem: []*gnmipb.PathElem{
```
robshakir: Can you make this the standard `tests := []struct{ ... }; for _, tt := range tests { ... }` layout please? It just makes things easier to extend when/if there are issues.

awly: I usually insist on the opposite in reviews: only add extra abstraction when it's actually needed.
Given the code sharing between PathToStrings and PathToString, more extra cases can be to TestPathToString.

ygot/pathstrings.go
awly: Clarified that path is absolute.
`prefix` is a property of Notification, and is not something this package deals with directly.

awly:
ping @robshakir

robshakir:
Sorry for the delay, LGTM.

PR #128, opened 2017-12-07
Continue validation on schema error.

ostromart:
Validation should not fail due to schema error, rather log error message.


robshakir:
If there is a schema error and the data is valid, we'll return nil errors right? Shouldn't this be an error case - i.e., we can't really validate this because we have a bad schema? Or should we assume here that goyang has validated the schema for us such that this can't actually occur?

ostromart:
Yes, it's an error case (separately we have to talk about the item that I saw triggering it). I think it comes down to how severely we want to react to such cases - I'd vote for just logging an error but still proceeding so that the schema can be fixed without being completely blocking.

PR #129, opened 2017-12-08
Beef up coverage.

ostromart:
Restore good coverage level for util/ directory. ytypes/ to come in separate PR.

### DIFF
--- a/util/common.go
+++ b/util/common.go
@@ -37,22 +37,15 @@ func stringMapKeys(m map[string]*yang.Entry) []string {
 	return out
 }
 
-// stringMapSetToSlice converts a string set expressed as a map m, into a slice
-// of strings.
-func stringMapSetToSlice(m map[string]interface{}) []string {
-	var out []string
-	for k := range m {
-		out = append(out, k)
-	}
-	return out
-}
-
 // TODO(mostrowski): move below functions into path package.
 
 // pathMatchesPrefix reports whether prefix is a prefix of path.
 func pathMatchesPrefix(path *gpb.Path, prefix []string) bool {
 	if len(path.GetElem()) < len(prefix) {
 		return false
+	}
+	for len(prefix) != 0 && prefix[len(prefix)-1] == "" {
+		prefix = prefix[:len(prefix)-1]
 	}
 	for i := range prefix {
 		if prefix[i] != path.GetElem()[i].GetName() {
@@ -66,6 +59,9 @@ func pathMatchesPrefix(path *gpb.Path, prefix []string) bool {
 // trimGNMIPathPrefix returns path with the prefix trimmed. It returns the
 // original path if the prefix does not fully match.
 func trimGNMIPathPrefix(path *gpb.Path, prefix []string) *gpb.Path {
+	for len(prefix) != 0 && prefix[len(prefix)-1] == "" {
+		prefix = prefix[:len(prefix)-1]
+	}
 	if !pathMatchesPrefix(path, prefix) {
 		return path
 	}
--- a/util/common_test.go
+++ b/util/common_test.go
@@ -1,0 +1,228 @@
+// Copyright 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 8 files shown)


util/common.go
```diff
@@ -37,23 +37,16 @@ func stringMapKeys(m map[string]*yang.Entry) []string {
...
-}
-
 // TODO(mostrowski): move below functions into path package.
 
 // pathMatchesPrefix reports whether prefix is a prefix of path.
 func pathMatchesPrefix(path *gpb.Path, prefix []string) bool {
 	if len(path.GetElem()) < len(prefix) {
 		return false
 	}
+	for len(prefix) != 0 && prefix[len(prefix)-1] == "" {
```
robshakir: Should we treat this case as an error -- i.e., I read this to be `prefix := []string{"one", "" , ""}` -- isn't the case that we have "" in the path something erroneous?

util/common.go
```diff
@@ -66,6 +59,9 @@ func pathMatchesPrefix(path *gpb.Path, prefix []string) bool {
 // trimGNMIPathPrefix returns path with the prefix trimmed. It returns the
 // original path if the prefix does not fully match.
 func trimGNMIPathPrefix(path *gpb.Path, prefix []string) *gpb.Path {
+	for len(prefix) != 0 && prefix[len(prefix)-1] == "" {
```
robshakir: Same comment as above.

util/common_test.go
```diff
@@ -0,0 +1,228 @@
...
+package util
+
+import (
+	"strings"
+	"testing"
+
+	gpb "github.com/openconfig/gnmi/proto/gnmi"
+)
+
+// pathNoKeysToGNMIPath conerts the supplied path, which may not contain any
```
robshakir: s/conerts/converts/

util/common_test.go
```diff
@@ -0,0 +1,228 @@
...
+// keys, into a GNMI path.
+func pathNoKeysToGNMIPath(path string) *gpb.Path {
+	out := &gpb.Path{}
+	for _, p := range strings.Split(path, "/") {
+		out.Elem = append(out.Elem, &gpb.PathElem{Name: p})
+	}
+	return out
+}
+
+// gnmiPathNoKeysToPath conerts the supplied GNMI path, which may not contain
```
robshakir: s/conerts/converts/

util/common_test.go
```diff
@@ -0,0 +1,228 @@
...
+		{
+			desc:   "not equal",
+			path:   "a/b/c",
+			prefix: "a/d",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			if got, want := pathMatchesPrefix(pathNoKeysToGNMIPath(tt.path), strings.Split(tt.prefix, "/")), tt.want; got != want {
```
robshakir: I see this is how you end up with ""  in the paths, but this isn't really safe to use in general right (since  `\/` will be split), so do we want to have the functions that are called handle this, or return error? I'd prefer the latter.

util/common.go
ostromart: I'm not sure. I haven't seen anything to suggest that // is illegal in schema paths so I've stuck with the common accepted path behavior. It sounds like you're interested in detecting errors with this - but perhaps those errors would be caught anyway at another layer without introducing this constraint here? Is there a specific scenario you can think of maybe?

util/common_test.go
ostromart: Done.

util/common_test.go
ostromart: Done.

util/common_test.go
ostromart: Yes, in general we wouldn't want to do this. The only place I think it's ok to use split("/") would be schema paths and tests.

util/common.go
robshakir: But isn't the only case that this check is really required based on the fact that you're using `strings.Split` on the input from tests? I'm just not sure that this mutation of the path is really the right thing to be doing regardless of whether this is an error or not.

util/common_test.go
robshakir: OK, SGTM.

util/common.go
ostromart: It's true that the reason trailing / appear is because they are coming from the tests, but that's exactly the point of the tests - I was to make sure that introducing a trailing / in the prefix is not going to lead to pathMatchesPrefix("a/b/c", "a/") failing. Although strictly speaking the trailing / should not be there I like having the extra robustness since trailing / often find their way into paths. 
I'm not sure if this is what you meant - if not let's chat live.

robshakir: No, I missed the point here -- thanks for clarifying for me.

PR #133, opened 2017-12-13
Handle escape characters on string->Path.Element conversion

awly:
In order to produce paths identical to Path.Elem, take the unescaped
paths and strigify them, instead of using raw strings originally
provided.

Note: this broke more tests than I was expecting. Please take a look at those to make sure I'm not actually breaking things.

### DIFF
--- a/ygot/pathstrings.go
+++ b/ygot/pathstrings.go
@@ -79,6 +79,8 @@ func PathToStrings(path *gnmipb.Path) ([]string, error) {
 	return p, nil
 }
 
+// elemToString returns a formatted string representation of a single Path.Elem
+// item. name and kv correspond to PathElem.Name and PathElem.Key.
 func elemToString(name string, kv map[string]string) (string, error) {
 	if name == "" {
 		return "", errors.New("empty name for PathElem")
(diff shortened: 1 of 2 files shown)


ygot/pathstrings.go
```diff
@@ -67,29 +67,43 @@ func PathToStrings(path *gnmipb.Path) ([]string, error) {
...
+	for k, v := range kv {
+		if k == "" {
+			return "", fmt.Errorf("empty key name (value: %s) in element %s", v, name)
+		}
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	for _, k := range keys {
+		v := strings.Replace(kv[k], `=`, `\=`, -1)
```
robshakir: I think the previous version of the specification said that `]` also needed to be escaped, see here: https://github.com/openconfig/reference/blob/792cdfbbe13cc8fbd3a20171671c95fe32e10fdb/rpc/gnmi/gnmi-path-conventions.md#paths-referencing-list-elements

ygot/pathstrings.go
```diff
@@ -67,29 +67,43 @@ func PathToStrings(path *gnmipb.Path) ([]string, error) {
...
+		elem, err := elemToString(e.Name, e.Key)
+		if err != nil {
+			return nil, fmt.Errorf("failed formatting PathElem at index %d: %v", i, err)
 		}
 		p = append(p, elem)
 	}
 	return p, nil
 }
 
+func elemToString(name string, kv map[string]string) (string, error) {
```
robshakir: Please add a function comment.

awly: Done

ygot/pathstrings.go
awly: Done

awly:
PTAL

PR #136, opened 2017-12-13
Add validation for leafref from leaf-list to another leaf-list.

ostromart:
This commit adds logic to ensure that a leaf-list data node that has a leafref to another leaf-list is a subset of the node it's pointing to.

### DIFF
--- a/ytypes/leafref.go
+++ b/ytypes/leafref.go
@@ -194,12 +194,12 @@ func matchesNodes(ni *util.NodeInfo, matchNodes []interface{}) (bool, error) {
 	// Check if any of the matching data nodes is equal to the referring
 	// node value. In the case that the referring node is a list, check that
 	// each node in the list is also in the target list.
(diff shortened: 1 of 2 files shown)


ytypes/leafref.go
```diff
@@ -192,35 +192,46 @@ func matchesNodes(ni *util.NodeInfo, matchNodes []interface{}) (bool, error) {
...
-		switch {
-		case util.IsValueScalar(ov):
-			util.DbgPrint("comparing leafref values %s vs %s", util.ValueStr(nii), util.ValueStr(other))
-			if util.DeepEqualDerefPtrs(nii, other) {
-				util.DbgPrint("values are equal")
-				match = true
-				break
+	// node value. In the case that the referring node is a list, check that
+	// each node in the list is also in the target list.
+	souceNodes := []interface{}{nii}
```
robshakir: `sourceNodes`?

PR #139, opened 2017-12-20
Resolve issue with duplicate field names, and missing protobuf types.

robshakir:
```
 * (M) ygen/codegen_test.go
   - Add new tests for protobufs with leafref->leafref->union keys.
 * (M) ygen/protoelements_test.go
   - Add new test cases for leafref type resolution for protobuf.
 * (M) ygen/protogen.go
   - Handle bug whereby if there is a list that has a key that has
     the same name as the list, then duplicate field names were
     used in the repeated message used for the key. This issue is
     resolved by appending _key onto the field name to indicate
     that this is the key.
   - Fix bug whereby if a leafref to a leafref to a union existed,
     then the code that was output would not include a type since
     it was not determined to be a union type (isUnionType() ->
     false). Since unionTypes can only ever be populated when there
     is a union, then this check was removed in both list key and
     normal message computation.
  * (M) ygen/protogen_test.go
   - Add test cases for list generation with identically named keys.
  * (A) testdata/proto/union-list-key*
   - Add test cases for the leafref->leafref->union case.
```


ygen/protogen.go
```diff
@@ -61,6 +61,34 @@ const (
 	// protoSchemaAnnotationOption specifies the name of the FieldOption used to annotate
 	// schemapaths into a protobuf message.
 	protoSchemaAnnotationOption = "(yext.schemapath)"
+	// protoMatchingListNameKeySuffix defines the suffix that should be added to a list
+	// key's name in the case that it matches the name of the list itself. This is requried
```
ostromart: s/requried/required

robshakir:
Thanks for the review! :-)

PR #140, opened 2017-12-21
Fix a number of bugs related to enumerations within unions.

robshakir:
This PR fixes a few issues related to enumerated types within unions, and recursively defined unions that were causing build failures for the generated OpenConfig protobuf files.


robshakir:
Thanks for the review!

PR #141, opened 2017-12-21
Add handling for partial data tree validation for leafrefs.

robshakir:
```
 * (M) ytypes/validate.go
   - Add a configuration struct to the ytypes library that allows
     some parameters relating to validation to be globally set for
     a caller. Particularly, this allows for enabling/disabling
     errors when a leafref points to a target leaf that does not
     exist. This is required for partial schema population..
 * (M) ytypes/leafref.go
 * (M) ytypes/leafref_test.go
   - Add functions to check whether errors should be thrown for
     leafref data errors. If data errors are not to be returned as
     errors, log this error occurring. Add test cases to check this.
 * (M) ytypes/schema_tests/validate_test.go
   - Add a test case that covers ignoring leafref data errors.
```


ytypes/validate.go
```diff
@@ -22,6 +22,26 @@ import (
...
+	// is populated, but validation is required - for example, configuration for
+	// a protocol within OpenConfig references an interface, but the schema being
+	// validated does not contain the interface definitions.
+	IgnoreMissingLeafrefData bool
+}
+
+var (
+	// Config is used to control the behaviour of the Validate functionality
+	// called against a generated struct.
+	Config *ValidationConfig
```
ostromart: From past experience, setting config through a global var may lead to trouble down the road. I recommend passing config around explicitly to Validate rather than setting the global var (even though it will clutter up the Validate API). It is also better to expose this to the users in this manner. I can take this work item if the changes prove to be extensive.

robshakir: Yes, I pondered the same. I wonder whether there's a good way to figure out how to make this API stable going forward - it seems like some interface like gRPC uses might be reasonable:

```golang
type ValidationOption interface {
  IsValidationOption()
}

type LeafrefOptions struct {
  DataMustExist bool
}
func (LeafrefOptions) IsValidationOption() { }

func Validate(opts ...ValidationOption) util.Error { ... }
```

ostromart: sgtm

robshakir:
OK! Let's try this for size, I created an interface that is implemented by the options -- this is defined in `ygot` so that we don't get a circular dependency between `ytypes` and `ygot`. PTAL!

The use of `...ygot.ValidationOptions` in the `Validate` API also means that this change should be backwards compatible for existing callers :+1:.

robshakir:
@ostromart PTAL, thank you!

ytypes/validate.go
```diff
@@ -22,9 +22,28 @@ import (
...
+// data types.
+type LeafrefOptions struct {
+	// IgnoreMissingData determines whether leafrefs that target a node
+	// that does not exist should return an error to the calling application. When
+	// set to true, no error is returned.
+	//
+	// This functionality is typically used where a partial set of schema information
+	// is populated, but validation is required - for example, configuration for
+	// a protocol within OpenConfig references an interface, but the schema being
+	// validated does not contain the interface definitions.
```
ostromart: Suggest a change to the wording to something like "that target a node that either doesn't exist or is not in the subtree of the element being validated"
Maybe mention that this could also be used in the case of validating a subtree rather than the root. In that case it would probably always be appropriate since leafrefs often require the tree root to navigate to.

ytypes/validate.go
```diff
@@ -22,9 +22,28 @@ import (
...
+}
+
+// IsValidationOption ensures that LeafrefOptions implements the ValidationOption
+// interface.
+func (LeafrefOptions) IsValidationOption() {}
+
 // Validate recursively validates the value of the given data tree struct
-// against the given schema.
-func Validate(schema *yang.Entry, value interface{}) util.Errors {
+// against the given schema. The supplied options (opts) are used to determine
```
ostromart: nit: the description of opts doesn't provide a lot of information in this case - suggest dropping that.

ytypes/validate.go
```diff
@@ -33,11 +52,19 @@ func Validate(schema *yang.Entry, value interface{}) util.Errors {
 		return util.NewErrs(fmt.Errorf("nil schema for type %T, value %v", value, value))
 	}
 
+	var leafrefOpt *LeafrefOptions
+	for _, o := range opts {
```
ostromart: Recommend making this a utility function (say in util/common.go). One possible way would be to add a constructor like NewLeafrefOptions(lro ...ygot.ValidationOptions). Then you could do ValidateLeafRefData(schema, value, NewLeafrefOptions(lro)). Just an idea, feel free to take it or leave it ;-)

robshakir:
Just looking at the code, we have the following places where we can return an error during `leafref` validation:

 * schema was `nil` for a particular element: returns error
 * `path` within the leafref was not valid: returns error
 * `dataNodesAtPath` - i.e., data tree node doesn't exist: returns error
 * `matchesNodes` returns an error -- this function never returns an error other than when the pointed to set of values is the empty set. I thought about changing the API here so that there was an explicit "some error has occurred that means we couldn't even check" signal vs. simply a "not found" one. This is relatively easy to do, but the function today never uses the `error` it returns as an actual error.
  * `matchesNodes` returns a `false` -- this is the case that we find a non-empty target but the value doesn't match.

So the change that we'd make here would be to switch the `error` return for the nil set to false, with no error, or change `matchesNodes` to have three return parameters:
 - Whether there was a match.
 - A verbose "no match" error message.
 - A fatal error that meant that the match wasn't really compared.

I'm happy to make this change -- thoughts?

ytypes/validate.go
robshakir: Per the analysis below, I'm not sure that this "is not in the subtree" is actually an error case that is ignored here? Can you see if you agree.

ytypes/validate.go
robshakir: SGTM!

ytypes/validate.go
robshakir: I can see the value of a utility function here. Just not sure I understand the semantics. You're saying here that this would just return the first `LeafrefOptions` from the `ValidationOptions`?

ostromart: Yes, the first or maybe all?

ostromart:
dataNodesAtPath will return an error if there's a data path problem to get to the root OR schema path problem from the root. It will return an error if there's a schema path problem but will return an empty node set if any node on the path from the leafref root is nil (it's super convoluted but I thought I was taking the least worst option here). That's the part that will get masked by the change, so I think it's the only place where we'd need to detect whether the data node can validly be nil (due to being defined in some other module or whatever the reason).
btw. to convince yourself that's the behavior, try setting your override flag and remove a leafref destination from a test (e.g. one in the schema_tests dir), the error should be suppressed.

ytypes/validate.go
robshakir: The only issue with this is that as we add new different types of options then we will end up walking the list of options multiple times, since we'll have NewLeafrefOpts, NewFoobarOpts, etc. I'll add a TODO here.

robshakir:
Right -- but with `dataNodesAtPath`, the override does not actually suppress this error -- it's only `matchesNodes` that has suppressed errors -- [see here](https://github.com/openconfig/ygot/blob/leafref_validation/ytypes/leafref.go#L61-L64)

The two calls that we have to `leafrefErrOrLog` is [here](https://github.com/openconfig/ygot/blob/leafref_validation/ytypes/leafref.go#L69-L78).

ostromart:
This change will suppress _all_ errors in the case where nodes should be in the data tree but are not populated. That's because if the target node is not in the tree is not necessarily reported as an error in dataNodesAtPath. It may just result in an empty set in which case the test is actually done in matchesNodes.

robshakir:
Sorry -- I think I misunderstood the comment above. If there is a "schema path problem from the root" this isn't suppressed, but yes, any element being null in the data tree in the path to the entity is suppressed. This is by design as otherwise, partial configuration instances can't be created.

From your initial question: "It should be possible to detect the case where we cannot reach a node vs. node is really not defined". I think that this is already the case. The node not being defined case is the only thing that can be overridden. 

I think I might be missing something here. We can try and talk realtime if I am.

robshakir:
I added such a check in the schema tests already -- https://github.com/openconfig/ygot/pull/141/files#diff-fdf9e1e7c0f2b4b11a260662c1c3522eR796 :-)

PR #143, opened 2017-12-29
Add option framework for MergeStructs.

robshakir:
```
 * (M) ygot/struct_validation_map.go:
 * (M) ygot/struct_validation_map_test.go:
    - Add an opts slice to MergeStructs which allows options to be
      specified when merging structs. Particularly, allow overwrite of a
      pointer field that is specified in both structs. In the case that
      the IfEqual bool is set true, merging is only allowed when the two
      values are equal. This is particularly aimed to handle merging
      lists which always have equal keys.
    - Change default behaviour for merging map structs to be allowing
      leaf overwrites.
```

### DIFF
--- a/ygot/struct_validation_map.go
+++ b/ygot/struct_validation_map.go
@@ -203,22 +203,12 @@ type EmitJSONConfig struct {
 	// Indent is the string used for indentation within the JSON output. The
 	// default value is three spaces.
 	Indent string
-	// ValidationOpts is the set of options that should be used to determine how
-	// the schema should be validated. This allows fine-grained control of particular
-	// validation rules in the case that a partially populated data instance is
-	// to be emitted.
-	ValidationOpts []ValidationOption
 }
 
 // EmitJSON takes an input ValidatedGoStruct (produced by ygen with validation enabled)
 // and serialises it to a JSON string. By default, produces the Internal format JSON.
 func EmitJSON(s ValidatedGoStruct, opts *EmitJSONConfig) (string, error) {
-	var vopts []ValidationOption
-	if opts != nil {
-		vopts = opts.ValidationOpts
-	}
-
-	if err := s.Validate(vopts...); err != nil {
+	if err := s.Validate(); err != nil {
 		return "", fmt.Errorf("validation err: %v", err)
 	}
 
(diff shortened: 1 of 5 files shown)


robshakir:
Ping @ostromart -- could PTAL if you have a moment please? :-)

ygot/struct_validation_map.go
```diff
@@ -332,18 +332,35 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
 
+// MergeOption is an interface that is implemented by all parameters which specify a merge
+// behaviour when merging GoStructs.
+type MergeOption interface {
+	// IsMergeOption is a marker method for these options.
+	IsMergeOption()
+}
+
+// AllowLeafOverwrite allows scalar (leaf) fields to be overwritten in the destination
+// strucft when specified.
```
ostromart: s/strucft/struct/

ygot/struct_validation_map.go
```diff
@@ -332,18 +332,35 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
-// identical members will be added in future iterations of this code.
-//
-// TODO(robjs): Fix the unimplemented test cases where two structs of
-// the same type have slices or maps that are already populated.
-// See https://github.com/openconfig/ygot/issues/74.
-func MergeStructs(a, b ValidatedGoStruct) (ValidatedGoStruct, error) {
+// Where two structs contain maps or slices that are populated in both a and b
+// their contents are merged. If a leaf is populated in both a and b, an error
+// is returned unless the AllowLeafOverwrite MergeOption is supplied.
+func MergeStructs(a, b ValidatedGoStruct, opts ...MergeOption) (ValidatedGoStruct, error) {
```
ostromart: b overwrites a should be explicitly stated.

ygot/struct_validation_map.go
```diff
@@ -445,8 +478,15 @@ func copyPtrField(dstField, srcField reflect.Value) error {
 	}
 
 	if !util.IsNilOrInvalidValue(dstField) {
-		// Return an error when we are overwriting fields in the destination.
-		return fmt.Errorf("destination value was set when merging, src: %v, dst: %v", srcField.Elem().Interface(), dstField.Elem().Interface())
+		if ao == nil {
```
ostromart: Per earlier comment, seems like this should check if values are equal.

ygot/struct_validation_map_test.go
```diff
@@ -1397,10 +1398,28 @@ func TestMergeStructs(t *testing.T) {
...
+		inB:    &validatedMergeTest{String: String("deschutes-pacific-wonderland")},
+		inOpts: []MergeOption{NewAllowLeafOverwrite(false)},
+		want:   &validatedMergeTest{String: String("deschutes-pacific-wonderland")},
+	}, {
+		name:    "error - merge leaf overwrite but not equal",
+		inA:     &validatedMergeTest{String: String("schneider-weisse-hopfenweisse")},
+		inB:     &validatedMergeTest{String: String("deschutes-jubelale")},
+		inOpts:  []MergeOption{NewAllowLeafOverwrite(true)},
+		wantErr: "error merging b to new struct: destination value was not equal to source when merging with override, src: deschutes-jubelale, dst: schneider-weisse-hopfenweisse",
 	}}
```
ostromart: Missing case of no override but values are equal. I think this should pass but if not I'd suggest putting a comment why not since that is not intuitive.

ygot/struct_validation_map.go
```diff
@@ -332,18 +332,35 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
 	return o, nil
 }
 
+// MergeOption is an interface that is implemented by all parameters which specify a merge
```
ostromart: Unless there's a clear need to pass in options through the public API, I'd recommend keeping the override option internal and making it just a boolean. I may be missing the big picture but from this code alone it seems like the option is being added just for the sake of testing. 
https://big.corp.google.com/~jmcmaster/testing/2016/08/episode-438-eliminate-yagni-smells.html

ygot/struct_validation_map.go
```diff
@@ -332,18 +332,35 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+type MergeOption interface {
+	// IsMergeOption is a marker method for these options.
+	IsMergeOption()
+}
+
+// AllowLeafOverwrite allows scalar (leaf) fields to be overwritten in the destination
+// strucft when specified.
+type AllowLeafOverwrite struct {
+	IfEqual bool // IfEqual allows leaves to be overwritten only if their value is equal.
+}
```
ostromart: This is a bit confusing. Why would you not allow an override if values are equal?

ygot/struct_validation_map.go
robshakir: The issue here is that if a user has a struct that has a list populated in it, then the key is always populated in both structs, i.e.,

```
type Device struct {
   Interface map[string]*Interface
}

type Interface struct {
  Name string
}
```

Since `Name` is the key of the `Interface` list, then it is populated in both structs - so we wanted a way to be able to allow this merge to happen.

ygot/struct_validation_map.go

ygot/struct_validation_map.go
robshakir: It's the other way around -- explicitly overwriting is disabled if `a` and `b` both have a populated field, but that field is not equal (i.e., there are two list entries that have the same map key, but their `Name` field - using the above example is not equal) then this throws an error.

ygot/struct_validation_map.go

ygot/struct_validation_map.go
robshakir: In this case, we return an error if fields are populated in both since the `AllowLeafOverwrite` option isn't set. Maybe I misinterpreted your comment - do you mean here that we should always just allow merging if the two values are equal?

ygot/struct_validation_map_test.go
robshakir: Today it won't -- because the default behaviour is to never allow override.

robshakir:
Thanks for the review -- it seems like there's a design question to ask here. I think your assertion is that we should always allow `a` and `b` to be merged if the fields in them are set as long as the contents are equal. This meets the use case that I'm fixing here -- so maybe it's just better to make this the default option - which would remove the whole need for the `MergeOption`. 

Thinking about this more, I think that this does sound more suitable (and was the original plan). So I'll update this PR just to do that. Please let me know if you think this isn't right.

PR #145, opened 2018-01-17
Fix ValidationOpts interface implementation for LeafrefOptions.

robshakir:
```
 * (M) ytypes/validate.go
  - LeafrefOptions was previously implementing ValidationOpts as
    a non-pointer receiver. The handling code expected the
    implementation to be a pointer - this CL changes the receiver to
    be a pointer receiver.
```


robshakir:
Thank you for the approval @ostromart ! :-)

PR #147, opened 2018-01-19
Add a method to rename a list entry, and associated testing.

robshakir:
This CL adds a generated method for each list which allows a list entry to be renamed from an old key to a new key. The method is supported for both single and compound key lists.

Tests are added for the generated code, along with a new `schema_tests` directory and set of tests which use the `exampleoc` schema to check functionality of the new `RenameXXX` methods.

### DIFF
--- a/ygen/codegen_test.go
+++ b/ygen/codegen_test.go
@@ -392,9 +392,14 @@ func TestSimpleStructs(t *testing.T) {
 		inFiles:             []string{filepath.Join(TestRoot, "testdata/structs/openconfig-simple.yang")},
 		wantStructsCodeFile: filepath.Join(TestRoot, "testdata/structs/openconfig-simple-no-compress.formatted-txt"),
 	}, {
-		name:                "simple openconfig test, with a list",
-		inFiles:             []string{filepath.Join(TestRoot, "testdata/structs/openconfig-withlist.yang")},
-		inConfig:            GeneratorConfig{CompressOCPaths: true},
+		name:    "simple openconfig test, with a list",
+		inFiles: []string{filepath.Join(TestRoot, "testdata/structs/openconfig-withlist.yang")},
+		inConfig: GeneratorConfig{
+			CompressOCPaths: true,
+			GoOptions: GoOpts{
+				GenerateRenameMethod: true,
+			},
+		},
 		wantStructsCodeFile: filepath.Join(TestRoot, "testdata/structs/openconfig-withlist.formatted-txt"),
 	}, {
 		name:                "simple openconfig test, with a list that has an enumeration key",
--- a/ygen/gogen.go
+++ b/ygen/gogen.go
@@ -504,6 +504,43 @@ func (t *{{ .Receiver }}) New{{ .ListName }}(
 }
 `
 
+	// goListMemberRenameTemplate provides a template for a function which renames
+	// an entry within a list. It is used to generate functions for each list within
+	// a generated Go struct.
+	goListMemberRenameTemplate = `
+// Rename{{ .ListName }} renames an entry in the list {{ .ListName }} within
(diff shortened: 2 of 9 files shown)


ygen/gogen.go
```diff
@@ -502,6 +502,59 @@ func (t *{{ .Receiver }}) New{{ .ListName }}(
...
 	return t.{{ .ListName }}[key], nil
 }
+`
+
+	// goListMemberRenameTemplate provides a template for a function which renames
+	// an entry within a list. It is used to generate functions for each list within
+	// a generated Go struct.
+	goListMemberRenameTemplate = `
+// Rename{{ .ListName }} renames an entry in the list {{ .ListName }} within
+// the {{ .Receiver }} struct. The entry with key oldk is renamed to newk updating
```
mum4k: Should the arguments be named using mixedCaps, i.e. oldK and newK ?

ygen/gogen.go
```diff
@@ -502,6 +502,59 @@ func (t *{{ .Receiver }}) New{{ .ListName }}(
...
+	oldk, newk {{ $key.Type -}}
+	{{- end -}}
+	{{- end -}}
+) error {
+	e, ok := t.{{ .ListName }}[oldk]
+	if !ok {
+		return fmt.Errorf("key %v not found in {{ .ListName }}", oldk)
+	}
+
+	ns, err := ygot.DeepCopy(e)
```
mum4k: Does this have to perform a copy?

We could as well get a ref to the existing value in the map, change it's name, delete and re-insert the same object.

ygen/gogen.go
```diff
@@ -502,6 +502,59 @@ func (t *{{ .Receiver }}) New{{ .ListName }}(
...
+	if err != nil {
+		return fmt.Errorf("cannot DeepCopy entry %v, got error: %v", oldk, err)
+	}
+	n, ok := ns.(*{{ .ListType }})
+	if !ok {
+		return fmt.Errorf("wrong type returned in list, got error: %v", err)
+	}
+	{{- if ne .KeyStruct "" -}}
+	{{- range $key := .Keys -}}
```
mum4k: Looking at the diff in the exampleoc/oc.go file, none of the changes exercises this code branch (range of key renames). Would it be useful to expand the example?

The tests in ygen/schema_tests/schema_test.go indicate that maybe they exercise these other code paths?

ygen/codegen_test.go
```diff
@@ -392,9 +392,14 @@ func TestSimpleStructs(t *testing.T) {
 		inFiles:             []string{filepath.Join(TestRoot, "testdata/structs/openconfig-simple.yang")},
 		wantStructsCodeFile: filepath.Join(TestRoot, "testdata/structs/openconfig-simple-no-compress.formatted-txt"),
 	}, {
-		name:                "simple openconfig test, with a list",
-		inFiles:             []string{filepath.Join(TestRoot, "testdata/structs/openconfig-withlist.yang")},
-		inConfig:            GeneratorConfig{CompressOCPaths: true},
+		name:    "simple openconfig test, with a list",
```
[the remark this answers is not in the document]
robshakir: It's a test that covers all methods for lists -- updated the name to reflect this.

ygen/gogen.go
```diff
@@ -502,6 +502,59 @@ func (t *{{ .Receiver }}) New{{ .ListName }}(
...
+	{{- if $key.IsScalarField }}
+	n.{{ $key.Name }} = &newk.{{ $key.Name }}
+	{{- else }}
+	n.{{ $key.Name }} = newk.{{ $key.Name }}
+	{{- end -}}
+	{{- end -}}
+	{{ else -}}
+	{{- range $key := .Keys -}}
```
robshakir: Since in this case, `len(Keys) == 1` (since we have a keyed list, and we know that there is not a composite key), changed this to `index`.

PR #148, opened 2018-01-24
Implement initial framework for struct diff [1/2]

robshakir:
This PR adds functionality to the `util` package which allows traversal of the structs solely based on their contents (i.e., the data tree) as a `ForEachDataField`. This is very similar to the `ForEachField` implementation, which co-traverses the data and schema trees.

The implementation here walks the data tree, and runs an `iterFunc` on each non-nil leaf - subsequently, that function can choose how to handle those values.

Apologies for the PR length, due to the length of the test cases, this ended up being >500 LOC. Please let me know if you'd like me to split it up further. The actual logic here is around 120 LOC.


util/reflect.go
```diff
@@ -545,6 +548,108 @@ func forEachFieldInternal(ni *NodeInfo, in, out interface{}, iterFunction FieldI
 	return errs
 }
 
+// ForEachDataField iterates the value supplied and iterates iterFunction for
+// each data tree node found in the function. No schema information is required
```
mum4k: (nit) "... found in the value."?

util/reflect.go
```diff
@@ -545,6 +548,108 @@ func forEachFieldInternal(ni *NodeInfo, in, out interface{}, iterFunction FieldI
...
+// to perform the iteration.
```
mum4k: Can we also say something about the in and out arguments?

util/reflect.go
```diff
@@ -545,6 +548,108 @@ func forEachFieldInternal(ni *NodeInfo, in, out interface{}, iterFunction FieldI
...
+	if IsValueNil(value) {
+		return nil
+	}
+
+	return forEachDataFieldInternal(&NodeInfo{FieldValue: reflect.ValueOf(value)}, in, out, iterFunction)
+}
+
+func forEachDataFieldInternal(ni *NodeInfo, in, out interface{}, iterFunction FieldIteratorFunc) (errs Errors) {
+	if IsValueNil(ni) {
+		return
```
mum4k: (optional) Naked returns tend to be confusing, unless this is part of the local style I would recommend:

return nil

Here and in a few other locations.

util/reflect.go
```diff
@@ -545,6 +548,108 @@ func forEachFieldInternal(ni *NodeInfo, in, out interface{}, iterFunction FieldI
 	return errs
 }
 
+// ForEachDataField iterates the value supplied and iterates iterFunction for
+// each data tree node found in the function. No schema information is required
+// to perform the iteration.
+func ForEachDataField(value, in, out interface{}, iterFunction FieldIteratorFunc) Errors {
+	if IsValueNil(value) {
```
mum4k: (optional) Is this check redundant? The forEachDataFieldInternal() func seems to start with the same check.

util/reflect.go
```diff
@@ -545,6 +548,108 @@ func forEachFieldInternal(ni *NodeInfo, in, out interface{}, iterFunction FieldI
...
+			}
+
+			nn.FieldValue = v.Field(i)
+			ps, err := SchemaPaths(nn.StructField)
+			if err != nil {
+				return NewErrs(err)
+			}
+			// In the case that the field expands to >1 different data tree path,
+			// i.e., SchemaPaths above returns more than one path, then we run
+			// this function for each different path. This ensures that the iterator
```
mum4k: (optional) I would specifically mention the term "recursively" to make it clear which function is "this function".

util/reflect.go
```diff
@@ -545,6 +548,108 @@ func forEachFieldInternal(ni *NodeInfo, in, out interface{}, iterFunction FieldI
...
+			// function runs for all expansions of the data tree as well as the GoStruct
+			// fields.
+			for _, p := range ps {
+				nn.PathFromParent = p
+				if IsTypeSlice(sf.Type) || IsTypeMap(sf.Type) {
+					nn.PathFromParent = p[0:1]
```
mum4k: Can you help me understand this indexed slice?

util/reflect_test.go
```diff
@@ -753,7 +766,25 @@ func TestForEachField(t *testing.T) {
...
+
+		fn, err := SchemaPaths(ni.StructField)
+		if err != nil {
+			errs = append(errs, err)
+		}
+		if l := len(fn); l != 1 {
+			errs = append(errs, fmt.Errorf("invalid schema path length %d for %v", l, ni.StructField.Name))
+		}
+
+		outs := out.(*string)
```
mum4k: (optional) Probably pointless since this is only test code, but even that might still serve as an example.

Would we be better off with out being a bytes.Buffer instead of doing the string appends?

Applies below too.

util/reflect_test.go
```diff
@@ -766,9 +797,33 @@ func TestForEachField(t *testing.T) {
...
+		}
+		if l := len(fn); l != 1 {
+			errs = append(errs, fmt.Errorf("invalid schema path length %d for %v", l, ni.StructField.Name))
+		}
 
+		*outs += fmt.Sprintf("%s/%s : \n%s\n, ", ValueStr(ni.FieldKey.Interface()), fn[0][0], ValueStr(s))
+		return
+	}
+
+	basicStruct1 = BasicStruct{Int32Field: int32(42), StringField: "forty two", Int32PtrField: toInt32Ptr(4242), StringPtrField: toStringPtr("forty two ptr")}
```
mum4k: (optional) Prefer to keep each part of the type literal on its own line. That way future diffs become more readable, since they will be line addition or deletion instead of inline changes.

Example:
&someType{
  Field1: val1,
  Field2: val2,
}

util/reflect_test.go
```diff
@@ -856,6 +911,103 @@ func TestForEachField(t *testing.T) {
...
+			iterFunc:     printSchemaAnnotationFieldsIterFunc,
+			wantOut: `int32 : 42, string : "forty two", int32ptr : 4242, stringptr : "forty two ptr", ` +
+				`int32 : 43, string : "forty three", int32ptr : 4343, stringptr : "forty three ptr", `,
+		},
+		{
+			desc:         "struct with string leaf-list",
+			parentStruct: &BasicSliceStruct{StringSlice: []string{"one", "two"}},
+			in:           nil,
+			iterFunc:     printSchemaAnnotationFieldsIterFunc,
+			wantOut:      ``,
```
mum4k: Why do we want an empty output here?

If that is intentional, can we update the test case description to make it more obvious?

util/reflect_test.go
```diff
@@ -856,6 +911,103 @@ func TestForEachField(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		outStr := ""
+		var errs Errors
+		errs = ForEachDataField(tt.parentStruct, tt.in, &outStr, tt.iterFunc)
+		if got, want := errs.String(), tt.wantErr; got != want {
+			t.Errorf("%s: ForEachDataField(%v, %#v, ...): did not get expected error, got: %s, want: %s", tt.desc, tt.parentStruct, tt.in, got, want)
+		}
+		if errs == nil {
```
mum4k: If Errors is a list, we should use len() to check that it is empty, since a non-nil list can still be empty.

if len(errs) == 0 { ... }

On the other hand, in Go we usually indent the error-flow. So let's consider the opposite logic:

if len(errs) > 0 {
  continue
}

if got, want := ...

ygot/diff.go
```diff
@@ -0,0 +1,115 @@
...
+func joingNMIPaths(parent *gnmipb.Path, child *gnmipb.Path) *gnmipb.Path {
+	p := proto.Clone(parent).(*gnmipb.Path)
+	for _, e := range child.Elem {
+		p.Elem = append(p.Elem, e)
+	}
+	return p
+}
+
+// pathSpec is a wrapper type used to store a gNMI path for use as a map key.
+type pathSpec struct {
```
mum4k: I don't think I follow this comment. How does this struct help the gNMI path to become a map key? It still only contains a slice, i.e. not something truly hashable.

Or maybe I am misunderstanding it altogether?

ygot/diff_test.go
```diff
@@ -0,0 +1,263 @@
...
+package ygot
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/golang/protobuf/proto"
+	"github.com/kylelemons/godebug/pretty"
+	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
```
mum4k: (optional) In the other files the renamed imports seem to be in their own group. Not sure if we want to follow that or just leave this up to gofmt.

ygot/diff_test.go
```diff
@@ -0,0 +1,263 @@
...
+			}, {
+				Name: "two",
+			}, {
+				Name: "three",
+			}},
+		},
+	}}
+
+	for _, tt := range tests {
+		if got := schemaPathTogNMIPath(tt.in); !proto.Equal(got, tt.want) {
```
mum4k: (optional) Do we usually test unexported functions? Would it be better from maintenance perspective to test all the behaviors via the public API? If an unexported function is complex enough to need its own test, maybe we should consider placing it in its own library.

ygot/diff_test.go
```diff
@@ -0,0 +1,263 @@
...
+				Key:  map[string]string{"two": "three"},
+			}, {
+				Name: "four",
+			}},
+		},
+	}}
+
+	for _, tt := range tests {
+		if got := joingNMIPaths(tt.inParent, tt.inChild); !proto.Equal(got, tt.want) {
+			diff := pretty.Compare(got, tt.want)
```
mum4k: (optional) Inside Google we usually pass want as the first argument to comparators like pretty to give the opposite direction (-want +got). Feel free to ignore, its just something I got used to.

ygot/diff_test.go
```diff
@@ -0,0 +1,263 @@
...
+				Annotation: []interface{}{},
+			},
+		},
+		inSchemaPaths: [][]string{{"foo", "bar"}, {"baz"}},
+		wantErr:       "could not find path specification annotation",
+	}}
+
+	for _, tt := range tests {
+		got, err := nodeValuePath(tt.inNI, tt.inSchemaPaths)
+		if err != nil && err.Error() != tt.wantErr {
```
mum4k: (optional) Should we prefer to test error semantics rather than the error value? Testing the error value can turn the test into a change detector where with a spelling change in the error message the author also needs to update tests.

util/reflect.go
robshakir: No - the check here and the check in the `forEachDataFieldInternal` are actually for different things -- the check in the `forEachDataFieldInternal` checks whether the supplied `NodeInfo` is nil - rather than whether the input value is `nil`. I think it's worth having both since we'll get either a panic (`reflect.ValueOf` on a nil value causes a panic) or a NPE without both these checks.

util/reflect.go
robshakir: Agreed -- have fixed this to returning `nil` or `errs` explicitly where required. I can't remember why we made this choice before, so I also fixed the `forEachFieldInternal` implementation as well. Thanks.

util/reflect.go
robshakir: Done.

util/reflect.go
robshakir: In the library, we sometimes compress elements out of the data tree that are within the YANG schema when `compress_paths` is set to true to the Go generator binary. This allows us to do things like remove the "config" container which isn't really required for folks just writing config generation from the output structs. We also remove containers that are only there to surround lists, since they are only required in vendor's implementations.

In these compression cases, then we end up with a path that looks like `config/bar` or `interfaces/interface` in the schema annotation. This just ensures that if the path were something like `config/bar/` and we had ended up with a length of greater than 2, then we sanitise the input. 

Strictly, this isn't really required, since we can never generate a three element long path in these cases from the `ygen` library. So it could be removed with no negative impact - it seemed worth being consistent though (`forEachFieldInternal` does this).

util/reflect_test.go
robshakir: It certainly could -- I was trying to keep consistency with the test cases that we had for `forEachFieldInternal`. I added a TODO to refactor this at some point.

util/reflect_test.go
robshakir: Done.

util/reflect_test.go
robshakir: This is an error -- and a bug -- thanks for catching it. The error is just in the test iteration function since `IsValueScalar` check was wrong, the overall implementation was calling the right thing.

util/reflect_test.go
robshakir: Done, thanks.

ygot/diff_test.go
robshakir: This needs to be fixed since outside of Google this isn't properly handled, done.

ygot/diff.go
robshakir: Yes, apologies -- this comment is wrong. The current design idea I have is that we'll have the map that is returned use pointers to this struct type for the key of the map. The `findSetLeaves` function in the other PR makes this clearer perhaps. I can update the comment there.

ygot/diff_test.go
robshakir: Throughout this library we do a lot of testing of unexported functions. The issue is that we end up with quite a lot of complexity that hides behind very few public APIs -- so we'd rather have some tests here that allow more easy isolation of issues for debugging. 

The logic we applied as to whether something is its own library is whether we really expect to call it from somewhere else -- `util` is generally a collection of things that fall into that category.

I'd prefer to keep the testing like this if we can -- it's a tradeoff between some duplication of testing, but more precise error localisation when we end up with regression issues IMHO.

ygot/diff_test.go
robshakir: Yeah - unfortunately we're the other way around in this code. The "got before want" rule and the approach in `pretty.Compare` are the same throughout. At some point I can go back and fix this.

ygot/diff_test.go
robshakir: This is the pattern that we've used elsewhere. The semantics are difficult to test -- since in general we're just returning an `error` type. We want to make sure that the error that we catch is the one that exercises the specific error handling. How do you suggest we check that without checking the contents without defining new types for every error handling branch (which seems maintenance overhead)?

util/reflect.go
mum4k: (optional) Thanks that makes sense, what do you think, does this warrant for an inline comment of some sort?

robshakir: Yep, added.

PR #149, opened 2018-01-24
Implement initial framework for struct diff [2/2]

robshakir:
This PR has a base of the `structdiff` branch - please feel free to not review it until `structdiff` is merged. I opened it since I think it gives some context as to what the implementation in the `structdiff` branch is intending to do.

The design for the diff functions is to use the `ForEachDataField` functionality implemented in `structdiff` to walk `original` and `modified`. The return value of the iteration function supplied is a map, keyed by a path identifier, with the set value. The second-stage of the diffing will then take these two outputs and return the overall diff as a gNMI `Notification` as we've discussed.

The implementation in this CL completes the walk started in `structdiff` and builds the function to actually return the map of changed leaves.


ygot/diff_test.go
```diff
@@ -249,6 +251,130 @@ func TestNodeValuePath(t *testing.T) {
...
+					gNMIPaths: []*gnmipb.Path{{
+						Elem: []*gnmipb.PathElem{{
+							Name: "a-list",
+						}},
+					}},
+				}},
+			},
+			FieldValue: reflect.ValueOf(&badListKeyType{Value: &cmplx}),
+		},
+		wantErr: "cannot convert keys to map[string]string: cannot convert type complex128 to a string for use in a key: (1+2i)",
```
mum4k: (optional) Back to our discussion about matching on error messages - consider at least making the matched error shorter and use a substring match. In this case this test actually verifies the error content that is produced by a function different than the one under test. Maybe just matching on "cannot convert keys" would be enough here and be less "change-detectory" at the same time?

robshakir: SGTM -- made this a substring match of the relevant error from this function.

Thinking about it, I wonder whether using something like https://godoc.org/google.golang.org/grpc/status could help here -- since this gives a structured code/error. I'll investigate doing this going forward.

PR #152, opened 2018-01-30
Add ygot.Diff public API to diff two GoStructs

robshakir:
This CL adds the public API for diffing two `GoStructs`. It returns a `gnmi.Notification` message with the difference between two structs.

Schema-based tests rather than mock inputs will be added in a subsequent CL.

### DIFF
--- a/ygot/diff.go
+++ b/ygot/diff.go
@@ -19,6 +19,7 @@ import (
 	"reflect"
 
 	"github.com/golang/protobuf/proto"
+	"github.com/kylelemons/godebug/pretty"
 	"github.com/openconfig/ygot/util"
 
 	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
@@ -234,7 +235,23 @@ func findSetLeaves(s GoStruct) (map[*pathSpec]interface{}, error) {
 		return nil, fmt.Errorf("error from ForEachDataField iteration: %v", errs)
 	}
 
-	return out, nil
+	uout := map[*pathSpec]interface{}{}
+	// Deduplicate the list, since the iteration function will be called
+	// multiple times for path tags that have >1 element.
+	for ok, ov := range out {
+		var skip bool
+		for uk := range uout {
+			if ok.Equal(uk) {
+				// This is a duplicate path, so we do not need to append it to the list.
+				skip = true
+			}
+		}
+		if !skip {
+			uout[ok] = ov
+		}
+	}
+
+	return uout, nil
 }
 
 // togNMIValue returns the GoStruct field v as a gNMI TypedValue message. It
@@ -245,22 +262,33 @@ func findSetLeaves(s GoStruct) (map[*pathSpec]interface{}, error) {
 //  - Union values - which are extracted from the interface type that is used
 //    to represent them.
 func togNMIValue(v interface{}) (*gnmipb.TypedValue, error) {
-	if _, ok := v.(GoEnum); ok {
-		estr, _, err := enumFieldToString(reflect.ValueOf(v), false)
-		if err != nil {
-			return nil, fmt.Errorf("cannot convert enum field to string: %v", err)
+	val := reflect.ValueOf(v)
+	switch {
+	case val.Type().Kind() == reflect.Int64:
+		if _, ok := v.(GoEnum); ok {
+			estr, _, err := enumFieldToString(reflect.ValueOf(v), false)
+			if err != nil {
+				return nil, fmt.Errorf("cannot convert enum field to string: %v", err)
+			}
+
+			return &gnmipb.TypedValue{
+				Value: &gnmipb.TypedValue_StringVal{estr},
+			}, nil
 		}
-
-		return &gnmipb.TypedValue{
-			Value: &gnmipb.TypedValue_StringVal{estr},
-		}, nil
-	}
-
-	if util.IsValueInterfaceToStructPtr(reflect.ValueOf(v)) {
+		// A non-enum implementing int64 is an invalid type, since fields should
+		// be pointers.
+		return nil, fmt.Errorf("invalid enum type")
+	case util.IsValueStructPtr(val):
 		var err error
-		if v, err = unionInterfaceValue(reflect.ValueOf(v), false); err != nil {
+		if v, err = unionInterfaceValue(val, false); err != nil {
 			return nil, fmt.Errorf("cannot resolve union field value: %v", err)
 		}
+	case util.IsValuePtr(val):
+		v = val.Elem().Interface()
+	case val.Type().Name() == EmptyTypeName:
+		v = val.Bool()
+	case val.Type().Name() == BinaryTypeName:
+		v = val.Bytes()
 	}
 
 	return value.FromScalar(v)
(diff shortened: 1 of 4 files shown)


robshakir:
Sorry -- the length of the struct literals in the tests made this CL too long. I thought it was better to have well-tested vs. incomplete testing. PTAL and let me know if you'd like me to split it.

Thanks!

ygot/diff.go
```diff
@@ -47,12 +49,35 @@ func joingNMIPaths(parent *gnmipb.Path, child *gnmipb.Path) *gnmipb.Path {
...
 
-// pathSpec is a wrapper type used to store a gNMI path for use as a map key.
+// pathSpec is a wrapper type used to store a set of gNMI paths to which
+// a value within a GoStruct corresponds to.
 type pathSpec struct {
 	// gNMIPaths is the set of gNMI paths that the path represents.
 	gNMIPaths []*gnmipb.Path
 }
 
+// Equal compares two pathSpecs, returning true all paths within the pathSpec
```
mum4k: Typo: "... true if all paths ...".

ygot/diff.go
```diff
@@ -210,5 +235,155 @@ func findSetLeaves(s GoStruct) (map[*pathSpec]interface{}, error) {
 		return nil, fmt.Errorf("error from ForEachDataField iteration: %v", errs)
 	}
 
-	return out, nil
+	uout := map[*pathSpec]interface{}{}
```
mum4k: (nit) The var should be named uOut.

ygot/diff.go
```diff
@@ -210,5 +235,155 @@ func findSetLeaves(s GoStruct) (map[*pathSpec]interface{}, error) {
...
+//  - Enumerated values (YANG enum or identityref leaves) which are returned as
+//    their string value, per the gNMI spec.
+//  - Union values - which are extracted from the interface type that is used
+//    to represent them.
+func togNMIValue(v interface{}) (*gnmipb.TypedValue, error) {
+	val := reflect.ValueOf(v)
+	switch {
+	case val.Type().Kind() == reflect.Int64:
+		if _, ok := v.(GoEnum); ok {
+			estr, _, err := enumFieldToString(reflect.ValueOf(v), false)
```
mum4k: Should be eStr.

ygot/diff.go
```diff
@@ -210,5 +235,155 @@ func findSetLeaves(s GoStruct) (map[*pathSpec]interface{}, error) {
...
+
+	return value.FromScalar(v)
+}
+
+// appendUpdate adds an update to the supplied gNMI Notification message corresponding
+// to the path and value supplied.
+func appendUpdate(n *gnmipb.Notification, path *pathSpec, val interface{}) error {
+	v, err := togNMIValue(val)
+	if err != nil {
+		s := pretty.Sprint(path)
```
mum4k: I wonder - would this and the following lines look better as a method on the pathSpec type?

Maybe we can even have it implement fmt.Stringer and then just use %v here in the fnt.Errorf() call.

ygot/diff.go
```diff
@@ -210,5 +235,155 @@ func findSetLeaves(s GoStruct) (map[*pathSpec]interface{}, error) {
...
+	for _, p := range path.gNMIPaths {
+		n.Update = append(n.Update, &gnmipb.Update{
+			Path: p,
+			Val:  v,
+		})
+	}
+	return nil
+}
+
+// Diff takes an original and modified GoStruct, which must of the same type
```
mum4k: ... which must be ...

ygot/diff.go
```diff
@@ -210,5 +235,155 @@ func findSetLeaves(s GoStruct) (map[*pathSpec]interface{}, error) {
...
+// and returns a gNMI Notification that contains the diff between them. The
+// GoStructs provided are recursively traversed. The original struct is
```
mum4k: The fact that they are recursively traversed is probably an implementation detail that doesn't need to be in the generated godoc?

ygot/diff_test.go
```diff
@@ -503,3 +487,553 @@ func TestFindSetLeaves(t *testing.T) {
...
+	tests := []struct {
+		desc          string
+		inOrig, inMod GoStruct
+		want          *gnmipb.Notification
+		wantErr       *string
+	}{{
+		desc:   "single path addition in modified",
+		inOrig: &renderExample{},
+		inMod: &renderExample{
+			Str: String("cabernet-sauvignon"),
```
mum4k: I like your selec-erm-collection of testdata ;)

ygot/diff_test.go
```diff
@@ -503,3 +487,553 @@ func TestFindSetLeaves(t *testing.T) {
...
+	}, {
+		desc:    "different types",
+		inOrig:  &renderExample{},
+		inMod:   &pathElemExample{},
+		wantErr: String("cannot diff structs of different types"),
+	}}
+
+	for _, tt := range tests {
+		got, err := Diff(tt.inOrig, tt.inMod)
+		if tt.wantErr != nil && err == nil || tt.wantErr != nil && !strings.Contains(err.Error(), *tt.wantErr) || tt.wantErr == nil && err != nil {
```
mum4k: (optional) Since we're doing a substring match now, consider calling the field wantErrSubStr or similar.

mum4k: (optional) This is now fairly hard to process. I wonder if a helper library similar to what we internally have in the errdiff package would help.

Alternatively consider splitting it:
if (err != nil) != tt.ErrSubstr != "" { // unexpected error}
if err != nil { substring match && continue}

ygot/render_test.go
```diff
@@ -1312,14 +1314,23 @@ func notificationSetEqual(a, b []*gnmipb.Notification) bool {
 
 	matchall := true
 	for _, aelem := range a {
-		var matched bool
+		var updateMatched, deleteMatched bool
 		for _, belem := range b {
+			if aelem == nil && belem != nil || belem == nil && aelem != nil || aelem == nil && belem == nil {
```
mum4k: (pre-existing, optional) Names here and below should be aElem, bElem...

ygot/render_test.go
```diff
@@ -1348,9 +1365,66 @@ func updateSetEqual(a, b []*gnmipb.Update) bool {
...
+func pathSliceEqual(a, b []*gnmipb.Path) bool {
+	pathIsLess := func(a, b *gnmipb.Path) bool {
+		for _, a := range a.Elem {
+			for _, b := range b.Elem {
+				if a.Name != b.Name {
+					return a.Name < b.Name
+				}
+				// If the element names are not equal then we consider the keys in
+				// alphabetical order.
+				akeys := stringKeys(a.Key)
```
mum4k: aKeys and bKeys.

ygot/diff.go
robshakir: Done.

ygot/diff.go
robshakir: Done.

ygot/diff.go
robshakir: Done.

ygot/diff.go
robshakir: SGTM! Implemented `String()` with a `pathSpec` receiver.

ygot/diff.go
robshakir: Done.

ygot/diff.go
robshakir: True, removed this sentence.

ygot/diff_test.go
robshakir: ;-)

ygot/diff_test.go
robshakir: Done.

robshakir: Agreed -- actually I noted that we had `errdiff` open sourced in openconfig/gnmi which is already a dep for this repo. Switched to using that library.

ygot/render_test.go
robshakir: Fixed.

ygot/render_test.go
robshakir: Done.

PR #153, opened 2018-01-30
Add annotation fields to the generated output.

robshakir:
This PR adds the option to generate structs with annotation fields. These fields are defined to be a slice of types that implement `ygot.Annotation`. This CL adds the code generation for the annotation fields, as well as the testing associated with it.

A subsequent CL will add schema tests, and JSON marshalling/unmarshalling support.


robshakir:
Apologies -- this isn't ready for review just yet. I hadn't checked the tests in ytypes.

mum4k:
No worries, please comment once it is ready.

robshakir:
Hi @mum4k -- this is ready for review now. Thanks! The `exampleoc` file is autogenerated so please feel free to ignore it :-)

util/reflect.go
```diff
@@ -444,6 +444,12 @@ func forEachFieldInternal(ni *NodeInfo, in, out interface{}, iterFunction FieldI
 		return nil
 	}
 
+	// If the field is an annotation, then we do not process it any further, including
+	// skipping running the iterFunction.
+	if IsYgotAnnotation(ni.StructField) {
```
mum4k: Should we say this in the doc comment on this function? Since it no longer does its work "ForEachField".

util/reflect_test.go
```diff
@@ -844,6 +844,26 @@ var (
 	}
 )
 
+type annotatedStruct struct {
+	FieldA     *string `path:"field-a"`
+	Annotation *string `path:"@field-a" ygotAnnotation:"true"`
+}
+
+var (
```
mum4k: (nit, optional) No need for a var group for just one var.

ygen/codegen.go
```diff
@@ -102,6 +102,14 @@ type GoOpts struct {
 	// GenerateRenameMethod specifies whether methods for renaming list entries
 	// should be generated in the output Go code.
 	GenerateRenameMethod bool
+	// AddAnnotationFields specifies whether annotation fields should be added to
+	// the generated structs. When set to true, a metadata field is added for each
+	// struct, and for each field of each struct. Metadata field's names are
+	// prefixed by the string specified in the AnnotationPrefix argument.
+	AddAnnotationFields bool
+	// AnnotationPrefix specifies the string which is prefixed to the name of
+	// annotation fields. It defaults to Λ.
```
mum4k: The part that says what it defaults to is likely to get out-of-sync with the default value in the annotation_prefix flag.

We can either remove that note from here or define a const like DefaultAnnotationFieldPrefix and reference that both from here and from the flag.

ygen/gogen.go
```diff
@@ -46,6 +46,12 @@ const (
 	// DefaultGoyangImportPath is the default path for the goyang/pkg/yang library that
 	// is used in the generated code.
 	DefaultGoyangImportPath string = "github.com/openconfig/goyang/pkg/yang"
+	// defaultMetadataPrefix is the default string that is used to prefix the name
+	// of metadata fields in the output Go structs.
+	defaultAnnotationPrefix string = "Λ"
```
mum4k: Ah we even have such constant - so if we make this public, we could use it in the flag, instead of having the literal in multiple locations.

ygot/diff.go
```diff
@@ -215,6 +215,12 @@ func findSetLeaves(s GoStruct) (map[*pathSpec]interface{}, error) {
 			return
 		}
 
+		// Handle the case of having an annotated struct - in the diff case we
```
mum4k: Should we mention this in the comment on the public func? (Diff)

util/reflect.go
robshakir: Added a comment here and a shorter one in `forEachInternal` to cover this. Thanks.

util/reflect_test.go
robshakir: ACK -- moved this definition into the test where it is the only place it is used.

ygen/codegen.go
robshakir: Yes, sorry -- this was the pattern that I used elsewhere, not sure why I didn't here. Fixing this.

ygen/gogen.go
robshakir: Done :-)

ygot/diff.go
robshakir: Yes, SGTM. Done.

PR #155, opened 2018-02-02
Add option to EmitJSONConfig to skip validation.

robshakir:
```
   - Add an option to skip validating a struct in EmitJSON. The
     default is to validate, but this allows a user who explicitly
     wants to skip all validation to force JSON marshalling.
```


ygot/struct_validation_map.go
```diff
@@ -214,11 +218,13 @@ type EmitJSONConfig struct {
 // and serialises it to a JSON string. By default, produces the Internal format JSON.
 func EmitJSON(s ValidatedGoStruct, opts *EmitJSONConfig) (string, error) {
 	var vopts []ValidationOption
+	var skipValidation bool
```
mum4k: (nit, optional) Consider a var group.

robshakir: Done, thanks.

PR #156, opened 2018-02-08
Implement a function to remove empty branches from a GoStruct.

robshakir:
```
  - When a struct is initialised with BuildEmptyTree, many of its
    branches may not be populated - but then have a nil value, such
    that comparisons with a struct literal are not equal. This
    results in complexities in testing and equality within consuming
    code. This commit implements a pruning function that removes any
    struct pointer field of a GoStruct which is found to be empty
    recursively.
```

### DIFF
--- a/ygot/struct_validation_map.go
+++ b/ygot/struct_validation_map.go
@@ -150,6 +150,54 @@ func initialiseTree(t reflect.Type, v reflect.Value) {
 	}
 }
 
+// PruneEmptyBranches removes branches that have no populated children from the
+// GoStruct s in-place. This allows a YANG container hierarchy that has been
+// initialised with BuildEmptyTree to have those branches that were not populated
+// removed from the tree. All subtrees rooted at the supplied GoStruct are traversed
+// and any encountered GoStruct pointer fields are removed if they equate to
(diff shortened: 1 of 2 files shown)


ygot/struct_validation_map.go
```diff
@@ -150,6 +150,54 @@ func initialiseTree(t reflect.Type, v reflect.Value) {
...
+// pointer fields are examined, since these are subtrees within the generated GoStruct
+// types.
+func pruneBranchesInternal(t reflect.Type, v reflect.Value) {
+	for i := 0; i < v.NumField(); i++ {
+		fVal := v.Field(i)
+		fType := t.Field(i)
+		if util.IsTypeStructPtr(fType.Type) {
+			// Create an empty version of the struct that is within the struct pointer.
+			// We can safely call Elem() here since we verified above that this type
+			// as a struct pointer.
```
mum4k: Typo: "is a struct ...".

robshakir:
No, thank you for the swift review! :-)

ygot/struct_validation_map.go
robshakir: Fixed, thanks.

PR #159, opened 2018-02-08
Exclude read only leaves from code generation.

### LINKED ISSUE #158: Add flag to allow read-only fields to be skipped in code generation.
For some use cases (e.g., configuration generation) - there is no requirement for `config false` fields to be included in generated structs. To keep the generated code size down, and aid comprehension by developers working with the generated code, we should implement an option to skip generation of fields in `ygen`.

This should be agnostic to the output format.

robshakir:
Fixes #158.


ygen/genstate.go
```diff
@@ -199,7 +206,7 @@ func (s *genState) buildDirectoryDefinitions(entries map[string]*yang.Entry, com
 
 			// Find the elements that should be rooted on this particular entity.
 			var fieldErr []error
-			elem.fields, fieldErr = findAllChildren(e, compressPaths)
+			elem.fields, fieldErr = findAllChildren(e, compressPaths, excludeConfigFalse)
```
mum4k: (optional) I am wondering if "config false" is the best possible terminology we can choose. Is it a well known term in this codebase? generally it is better to describe a thing by saying what it is rather than what it isn't. I feel that excludeState or excludeReadOnly would be more obvious.

Up to you, feels like this would be a large change to ask for.

ygen/goelements_test.go
```diff
@@ -29,6 +29,7 @@ func TestFindChildren(t *testing.T) {
 	tests := []struct {
 		name             string
 		inElement        *yang.Entry
+		inExcludeState   bool
```
mum4k: We probably should use one terminology - i.e. ExcludeConfigFalse or ExcludeState.

ygen/genstate.go
robshakir: RFC6020 Section 4.2.3 is where this is defined: https://tools.ietf.org/html/rfc6020#section-4.2.3  -- I've tried to be as consistent as possible with the RFC, but I agree with this comment. Refactored throughout.

ygen/goelements_test.go
robshakir: Agreed. Adopted 'state' throughout.

PR #160, opened 2018-02-09
Add support for marshalling annotations to JSON

robshakir:
This CL adds support for marshalling `[]ygot.Annotation` fields to JSON.  An example is included within `demo/device` to provide users an example and act as a test within the repo for the end to end `ygen` / `ygot` and `ytypes` libraries.

CL length is generally in the generated code, rather than being in library logic. PTAL! Thanks :+1:


demo/device/devicedemo.go
```diff
@@ -44,6 +45,23 @@ func main() {
 	fmt.Println(ietfjson)
 }
 
+// ExampleAnnotation is used to demonstrate the ygot.Annotation interface,
+// and the ability for ygot to add annotations to generated structs.
+type ExampleAnnotation struct {
+	GeneratorSource string `json:"gen-source"`
```
mum4k: (optional) Since this serves as a public example, maybe we shouldn't introduce the Generator term here.

How about a ConfigSource ?

ygot/render.go
```diff
@@ -839,7 +840,14 @@ func constructJSON(s GoStruct, parentMod string, args jsonOutputConfig) (map[str
 			continue
 		}
 
-		value, err := constructJSONValue(field, pmod, args)
+		var value interface{}
+
+		if util.IsYgotAnnotation(fType) {
+			value, err = constructJSONAnnotationSlice(field)
```
mum4k: (optional) The word "construct" feels redundant in these names. Here and below.

demo/device/devicedemo.go
robshakir: SGTM, done.

ygot/render.go
robshakir: True -- refactored all the `construct*` names other than the public API functions.

PR #162, opened 2018-02-09
 Fix PruneEmptyBranches. 

robshakir:
When `ygot.BuildEmptyTree` is run on a struct, then the struct fields that are
struct pointers are recursively initialised. `ygot.PruneEmptyBranches` checked
whether a struct was non-nil, and removed it if not. However, this only worked
where the struct had only non-struct ptr children, since BuildEmptyTree had
already initialised struct ptr fields. The test cases that were used for
PruneEmptyBranches did not reflect this completely.

Additionally, `PruneEmptyBranches` did not iterate through maps and recursively
remove branches lower down in the tree. This made it less effective at removing
initialised structs that are empty than it should have been.

This CL fixes `PruneEmptyBranches`, adds better testing, and tests against the
exampleoc schema.


robshakir:
No worries -- I should have used the schema for the test case too, since I hadn't realised that the test case was overly simplistic.

robshakir:
Thanks for the review! :-)

PR #163, opened 2018-02-10
Resolve protobuf generation issues.

robshakir:
This CL resolves a small number of issues with protobuf and Go generation:

- When excluding a module, this was done based on the module that included the YANG statements *not* the module that instantiated the module. This resulted in a configuration that excluded `openconfig-interfaces` not being able to resolve containers within the `interface-ref` grouping as they were excluded themselves.
- In addition to the above, the schema tree would only be built based on included modules - which resulted in leafrefs not being able to be resolved if they pointed to an excluded module. The `schematree` is now resolved based on all modules, regardless of whether they are output.
- When multiple YANG modules defined entities at the root, and fakeroot a fakeroot was generated, they would end up in the same directory.
- When there was a list within a surrounding container at the root, and fakeroot and compress paths was enabled,  then this list would sometimes be placed in a separate package due to incorrect handling of package name generation.

Test cases for the above issues have been added.


ygen/codegen.go
```diff
@@ -464,7 +465,6 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
 // It returns a GeneratedProto3 struct containing the messages that are to be
 // output, along with any associated values (e.g., enumerations).
 func (cg *YANGCodeGenerator) GenerateProto3(yangFiles, includePaths []string) (*GeneratedProto3, *YANGCodeGeneratorError) {
-	// TODO(github.com/openconfig/ygot/issues/20): Handle enumerated types in proto messages.
 	mdef, errs := mappedDefinitions(yangFiles, includePaths, &cg.Config)
```
ostromart: Why not use one of the already existing error list definitions? e.g. util/errs.go, the one that pictor uses, or https://github.com/hashicorp/go-multierror is a good choice.

ygen/codegen.go
```diff
@@ -712,26 +711,31 @@ func mappedDefinitions(yangFiles, includePaths []string, cfg *GeneratorConfig) (
...
 	for _, module := range modules {
-		findMappableEntities(module, dirs, enums, cfg.ExcludeModules, cfg.CompressOCPaths)
+		errs = append(errs, findMappableEntities(module, dirs, enums, cfg.ExcludeModules, cfg.CompressOCPaths, modules)...)
 		if module == nil {
 			errs = append(errs, errors.New("found a nil module in the returned module set"))
 			continue
 		}
 		// Ensure that we do not try and traverse an empty module.
-		if module.Dir != nil && !excluded[module.Name] {
+		if module.Dir != nil {
```
ostromart: optional: len(module.Dir) would be more robust if the semantics of that field are not explicitly defined.

ygen/codegen.go
```diff
@@ -796,29 +800,27 @@ func mappableLeaf(e *yang.Entry) *yang.Entry {
...
-		if s == pp[1] {
-			return
+		for _, m := range modules {
+			if m.Name == s && m.Namespace().Name == e.Namespace().Name {
+				return nil
+			}
 		}
 	}
 
+	var errs []error
```
ostromart: Since util.Errors is being returned, it's better to use that type and associated functions, which take care of those annoying nil errors appearing in error lists among other things.

ygen/codegen.go
```diff
@@ -832,12 +834,12 @@ func findMappableEntities(e *yang.Entry, dirs map[string]*yang.Entry, enums map[
 			// If this is a config or state container and we are compressing paths
 			// then we do not want to map this container - but we do want to map its
 			// children.
-			findMappableEntities(ch, dirs, enums, excludeModules, compressPaths)
+			errs = append(errs, findMappableEntities(ch, dirs, enums, excludeModules, compressPaths, modules)...)
```
ostromart: Use AppendErrs, here and elsewhere.

ygen/protogen.go
```diff
@@ -338,14 +338,32 @@ type protoMsgConfig struct {
...
 	return writeProto3MsgSingleMsg(msg, msgs, state, cfg)
 }
 
+// isValidNestedMessage determines whether the message represented by the supplied
+// yangDirectory is a message that should be output when nested messages are being
+// created. The compressPaths argument specifies whether path compression is enabled.
+// Valid messages are those that are direct children of a module, or become a direct
+// child when path compression is enabled (i.e., lists that have their parent
+// surrounding container removed).
+func isValidNestedMessage(msg *yangDirectory, compressPaths bool) bool {
```
ostromart: nit: "invalid" suggests some error - consider s/isValidNestedMessage/outputNestedMessage/.

ygen/codegen.go
robshakir: Done -- and throughout this file.

ygen/codegen.go

ygen/codegen.go
robshakir: Yes, I think this was just for consistency, but I've moved everything to util.Errors now. Would prefer to avoid an external dependency.

ygen/codegen.go
robshakir: I think this check can actually be entirely removed, since we'd expect that the map in `Dir` is nil and range will still not panic with NPE here. Removing.

ygen/protogen.go

ygen/codegen.go
```diff
@@ -363,13 +333,13 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
...
 	enumTypeMap := map[string][]string{}
-	codegenErr := NewYANGCodeGeneratorError()
+	var codegenErr util.Errors
 	var structSnippets []string
 	for _, structName := range orderedStructNames {
 		structOut, errs := writeGoStruct(structNameMap[structName], goStructs, cg.state,
 			cg.Config.CompressOCPaths, cg.Config.GenerateJSONSchema, cg.Config.GoOptions)
 		if errs != nil {
-			codegenErr.Errors = append(codegenErr.Errors, errs...)
+			util.AppendErrs(codegenErr, errs)
```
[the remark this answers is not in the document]
ostromart: That also raises the question of whether some of these at least could be covered in tests.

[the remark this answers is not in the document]
robshakir: Thanks for catching this -- fix applied.

In terms of testing -- the issue is that we're trying to catch errors that are not possible to inject in this function since they depend on something being wrong in the input YANG or the previous processing stage. I added tests that catch all these top-level errors, but the testing that the right error is caught if a previous stage gives bad input is on the per function unit tests.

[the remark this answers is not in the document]
robshakir: The `nil` check can't get removed in a bunch of cases, since we also do a `continue`.

ygen/codegen.go
```diff
@@ -616,7 +616,7 @@ func processModules(yangFiles, includePaths []string, options yang.Options) ([]*
 	var errs util.Errors
 	for _, name := range yangFiles {
 		if err := moduleSet.Read(name); err != nil {
```
ostromart: optional: here you could do errs = util.AppendErr(errs, moduleSet.Read(name))

robshakir: Thanks, done.

PR #164, opened 2018-02-10
Flag support to sort Protobuf fields by type, then name.

### LINKED ISSUE #116: Protobuf fields could be grouped by type.
In order that fields that have their type defined by the same module (e.g., `openconfig.enums.FOO` and `openconfig.enums.BAR`)  are grouped together, field sorting could by via type rather than name. This needs post-processing of the fields since the type is not known at creation time. It could be done in `genProto3MsgCode`.

robshakir:
Fixes #116.


robshakir:
This was an aesthetic PR and hasn't come up again for the last 6 years, so I'm going to go ahead and close it.

PR #165, opened 2018-02-12
Add support for Getter and Append generation

### LINKED ISSUE #161: Automatically generate a FindOrNew function for every list
I see a bit of an anti pattern forming in my ygot code.  One function that works with a port calls NewInterface and future functions must check for the existence of the list element.

Example:
port, err := dev.NewInterface(portName)
if err != nil {
	return err
}

Next function that works with this interface:
port := dev.Interface[portName]
if port == nil {
	port, err = dev.NewInterface(portName)
        if err != nil {
          return err
       }
}

It would be nice if there was a helper function that would allow each caller to do:

port := dev.FindOrNewInterface(portName)

This is very similar to the Google map_util.h helper functions.
https://github.com/google/protobuf/blob/master/src/google/protobuf/stubs/map_util.h

If I have time, I might try adding this......

robshakir:
This CL adds:

* A flag (`GenerateGetters`) to `ygen` that creates `GetOrCreateXXX()` methods for structs, and `GetOrCreateXXX(<list key spec>)` methods for lists within the output code.
* A flag (`GenerateAppendMethod`) to `ygen` that creates `AppendXXX(<list member type>)` for lists. This function takes a list member that has been created outside of the structure, extracts its keys and appends it to the list.
* Unit tests for the above in `ygen`.
* Schema-aware tests in `ygen/schema_tests` for these method's functionality.

Fixes #161 :+1: :ship:


ygen/gogen.go
```diff
@@ -1151,6 +1284,76 @@ func generateValidator(buf *bytes.Buffer, structDef generatedGoStruct) error {
...
+	return nil
+}
+
+// generateGetOrCreateList generates a getter function similar to that created
+// by the generateGetOrCreateStruct function for maps within the generated Go
+// code (which represent YANG lists). It handles both simple and composite key
+// lists.
+//
+// If the list described has a single key, the argument to the function is the
+// non-pointer key value. If the list has a complex type, it is a instance of
```
mum4k: (nit) "... an instance ..." ?

ygen/gogen.go
```diff
@@ -1151,6 +1284,76 @@ func generateValidator(buf *bytes.Buffer, structDef generatedGoStruct) error {
...
+// the generated key type for the list.
+//
+// The generated function returns the existing value if the key exists in the
+// list, or creates a new value using the NewXXX method if it does not exist.
+// The generated function is written to the supplied buffer, using the method
+// argument to determine the list's characteristics in the template.
+func generateGetOrCreateList(buf *bytes.Buffer, method *generatedGoListMethod) error {
+	if err := goTemplates["getOrCreateList"].Execute(buf, method); err != nil {
```
mum4k: Can this and the following function just return directly?

return gotemplates[...]...

ygen/schema_tests/schema_test.go
```diff
@@ -80,3 +80,67 @@ func TestMultiKeyListRename(t *testing.T) {
...
+	}
+
+	if _, ok := in.NetworkInstance["DEFAULT"]; !ok {
+		t.Errorf("AppendNetworkInstance(%v): did not find element after append, got: %v, want: true", ni, ok)
+	}
+}
+
+func TestMultiKeyAppend(t *testing.T) {
+	in := &exampleoc.Device{}
+	_, err := in.NewNetworkInstance("DEFAULT")
```
mum4k: (nit) Can we minimize the scope of err here?

if _, err := in.NewNetworkInstance(); err != nil {...}

ygen/gogen.go
```diff
@@ -508,6 +518,103 @@ func (t *{{ .Receiver }}) New{{ .ListName }}(
...
+	key := {{ $key.Name }}
+	{{- end -}}
+	{{- end }}
+
+	if v, ok := t.{{ .ListName }}[key]; ok {
+		return v
+	}
+	// Safely discard the error from New since we check for the existence
+	// of the list key above. This allows chaining of GetOrCreate methods.
+	v, _ := t.New{{ .ListName }}(
```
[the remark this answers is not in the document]
robshakir: Sure -- changed this to panic if the err ever returns true. I didn't consider the case that someone made the `NewXXX` return an error for some other reason.

ygen/gogen.go
robshakir: Thanks, fixed.

ygen/gogen.go
robshakir: Yes, not sure why I wrote this like this -- fixed.

ygen/schema_tests/schema_test.go
robshakir: ACK, done, thanks.

PR #166, opened 2018-02-16
Ensure that -exclude_state removes surrounding containers from lists.

robshakir:
```
 * (M) ygen/*
   - This CL fixes a bug whereby -exclude_state would still attempt to
     include the surrounding container of a read-only list such that
     code generation would fail when searching for the grandchild of
     the container (i.e., the list).
```


robshakir:
Thanks for the reviews :-)

PR #173, opened 2018-03-13
Generate getters that consider schema defaults.

robshakir:
```
 * (M) ygen/*
   - Add code to generate a 'GetXXX' function that retrieves a
     field's value directly. If the field has a default within the
     YANG schema and the field is unset, the default value is
     returned. Methods are generated for lists, containers and
     scalar fields.
 * (M) demo/getters/*
   - Add example code that shows use of the getter functions and
     makes up an integration test for generated code that contains
     the getters. This code is used in preference to the exampleoc
     schema to aid in debugging.
 * (M) demo/{protobuf_getting_started, uncompressed}:
   - Ensure checked-in generated code is up to datee.
 * (M) exampleoc/oc.go
   - Regenerate exampleoc using the new getters.
```

This PR looks huge due to changes in the generated code - please feel free to ignore changes in `exampleoc/` and the generated protobuf files.


mum4k:
Rob, can you confirm if this should be reviewed now that we decided not to generate the leaf getters?

robshakir:
hi @mum4k -- it doesn't, I'll close it and open a new one for the `map`/`struct` field getters.

robshakir:
Thanks for following up on this!

PR #174, opened 2018-03-17
Getters for containers and lists.

robshakir:
This CL adds support for generating getters for `container` and `list` nodes that exist as struct fields within the generated structs. Per our discussions, we do not generate getters for `leaf` or `leaf-list` entities at this time.


ygen/gogen.go
```diff
@@ -382,6 +386,19 @@ func (s *{{ .StructName }}) Validate(opts ...ygot.ValidationOption) error {
 	}
 	return nil
 }
+`
+
+	// goContainerGetterTemplate defines a template that generates a getter function
+	// for the field of a generated struct. It is generated only for YANG containers.
+	goContainerGetterTemplate = `
+// Get{{ .Field.Name }} returns the value of the {{ .Field.Name }} struct pointer
```
mum4k: Should we be more specific and mention that this also happens to return nil in some cases?

ygen/gogen_test.go
```diff
@@ -1242,6 +1242,21 @@ func (t *Tstruct) GetOrCreateListWithKey(KeyLeafOne string, KeyLeafTwo int8) (*L
...
+// GetListWithKey retrieves the value with the specified key from
+// the ListWithKey map field of Tstruct.
+func (t *Tstruct) GetListWithKey(KeyLeafOne string, KeyLeafTwo int8) (*ListWithKey){
+
+  key := Tstruct_ListWithKey_Key{
+		KeyLeafOne: KeyLeafOne,
+		KeyLeafTwo: KeyLeafTwo,
+	}
+
+  if lm, ok := t.ListWithKey[key]; ok {
```
mum4k: Should the GetList getters also be nil receiver resilient? So that we can chain through empty lists.

ygen/schema_tests/schema_test.go
```diff
@@ -151,3 +151,17 @@ func TestGetOrCreateMultiKeyList(t *testing.T) {
 		t.Errorf("GetOrCreateNetworkInstance('DEFAULT').GetOrCreateProtocol(ISIS, '0').GetOrCreateGlobal().MaxEcmpPaths: got incorrect return value, got: %v, want: %v", got, want)
 	}
 }
+
+func TestGetters(t *testing.T) {
```
mum4k: (optional) Consider something like TestGetterChaining or similar.

ygen/schema_tests/schema_test.go
```diff
@@ -151,3 +151,17 @@ func TestGetOrCreateMultiKeyList(t *testing.T) {
...
+	d := &exampleoc.Device{}
+	if got := d.GetSystem().GetAaa(); got != nil {
+		t.Errorf("chained getters: GetSystem().GetAaa() did not return nil, got: %v, want: nil", got)
+	}
+
+	intName := "eth0"
```
mum4k: (optional) Consider calling this "want" instead of intName.

ygen/testdata/structs/openconfig-list-enum-key.getters-append.formatted-txt
```diff
@@ -150,6 +177,21 @@ func (t *OpenconfigListEnumKey_Top_MultiKey) GetOrCreateEkm(K1 E_OpenconfigListE
...
+// GetEkm retrieves the value with the specified key from
+// the Ekm map field of OpenconfigListEnumKey_Top_MultiKey.
+func (t *OpenconfigListEnumKey_Top_MultiKey) GetEkm(K1 E_OpenconfigListEnumKey_Top_MultiKey_Ekm_Config_K1, K2 E_OpenconfigListEnumKey_FooIdentity) (*OpenconfigListEnumKey_Top_MultiKey_Ekm){
+
+  key := OpenconfigListEnumKey_Top_MultiKey_Ekm_Key{
+		K1: K1,
+		K2: K2,
+	}
+
+  if lm, ok := t.Ekm[key]; ok {
```
mum4k: Similarly here, looks like this wouldn't handle a nil receiver.

ygen/gogen_test.go
robshakir: Good catch, thank you. Done.

ygen/schema_tests/schema_test.go
robshakir: Done.

ygen/schema_tests/schema_test.go
robshakir: Done.

ygen/testdata/structs/openconfig-list-enum-key.getters-append.formatted-txt
robshakir: Fixed :-)

robshakir:
Thanks for the review -- PTAL at the updated code.

ygen/gogen.go
robshakir: This was done in the last commit, thanks.

PR #177, opened 2018-04-10
Add merge of slices fields to ygot.

### LINKED ISSUE #175: Merging structs that contain leaf-lists of unions fails
This is currently unimplemented as it is a `[]struct { .. }` that is being merged.

robshakir:
Fixes #175.

This PR adds the ability to merge two slice fields within a struct when the destination field is set. Previously, this would throw an error. The slices are checked for uniqueness so as not to violate RFC6020 constraints.


robshakir:
Thanks for the review! Merging :-)

[This testdata is all beer related, since at least there are enough of them that I don't really run out of material ;-)]

Issue #176, opened 2018-04-09
Allow splitting of output Go code into multiple files.


robshakir:
This issue tracks implementing an approach that splits the Go code produced by ygot into multiple files to ease consumption by code indexing tools.

wenovus:
#409 implements this with a flag.

PR #187, opened 2018-04-25
Address #176 - split Go code into multiple input files.

robshakir:
```
* (M) ygen/*
 * (M) generator/*
  - When large models are generated by ygot, the output code file can
    become unwieldy. This PR addresses this, as described in #176, by
    splitting various entities into their own source files. These
    files remain within a single package.
  - Structs within the output are sorted into a series of alphabetically
    named files in order to keep the struct filesize down.
 * (M) exampleoc/*
 * (M) demo/uncompressed/*
  - Ensure that the -output_dir function is leveraged in test cases
    such that breakage and build is detected. Currently, this output
    mode needs goimports to be run against the created files.
```


demo/uncompressed/uncompressed.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 // Generate rule to create the example structs:
-//go:generate go run ../../generator/generator.go -path=yang -output_file=pkg/demo/uncompressed.go -package_name=demo -generate_fakeroot -fakeroot_name=root yang/example.yang
+//go:generate go run ../../generator/generator.go -path=yang -output_file=pkg/demo/uncompressed.go -package_name=demo -generate_fakeroot -fakeroot_name=root yang/example.yang
```
mum4k: Was this change intentional?

generator/generator.go
```diff
@@ -30,12 +31,31 @@ import (
...
 
+const (
+	// enumMapFn is the filename to be used for the enum map when Go code is output to a directory.
+	enumMapFn = "enum_map.go"
+	// enumFn is the filename to be used for the enum code when Go code is output to a directory.
+	enumFn = "enum.go"
+	// schemaFn is the filename to be used for the schema code when outputting to a directory.
+	schemaFn = "schema.go"
+	// interfaceFn is the filename to be used for interface code when outputting to a directory.
+	interfaceFn = "union.go"
```
mum4k: Why is the file that contains interface code called union? Some terminology I am not familiar with maybe?

generator/generator.go
```diff
@@ -82,6 +105,88 @@ func writeGoCode(w io.Writer, goCode *ygen.GeneratedGoCode) error {
...
+	contents string
+	// oneoffHeader indicates whether the one-off header should be included in this
+	// file.
+	oneoffHeader bool
+}
+
+// makeOutputSpec generates a map, keyed by filename, to a codeOut struct containing
+// the code to be output to that filename. It allows division of a ygen.GeneratedGoCode
+// struct into a set of source files. It divides the methods, interfaces, and enumeration
+// code snippets into their own files. Structs are output into files dependent on the
```
mum4k: Is this a good split? (Based on the first name). How does the distribution look like, do we have a good spread?

generator/generator.go
robshakir: When ygen generates any interface it is because it is for a YANG union - a bit like protobuf oneof. So it seemed like a reasonable split here was to say that this file was named after the YANG construct that it contained. I'm agnostic on this.

generator/generator.go
robshakir: `N` continues to be the largest for OpenConfig, since it contains all the `NetworkInstance...` code. An alternate splitting strategy would be to do it at the next-level down from `NetworkInstance`, but `NetworkInstance_Protocol` would still end up being quite big -- and this would be quite a high number of files. Here's the breakdown of the `exampleoc` package (which contains almost the whole schema, and has`-generate_getters`, `-generate_rename` and `-generate_append`. It does not have `-exclude_state`.

```
[8:10] openconfig/ygot/splitoc   file-split ✔    9h46m ◒  
▶ du -sch pkg/oc/*
480K	pkg/oc/enum.go
160K	pkg/oc/enum_map.go
2.1M	pkg/oc/methods.go
2.5M	pkg/oc/schema.go
 24K	pkg/oc/structs_A.go
152K	pkg/oc/structs_B.go
 48K	pkg/oc/structs_C.go
4.0K	pkg/oc/structs_D.go
 80K	pkg/oc/structs_I.go
 24K	pkg/oc/structs_L.go
 88K	pkg/oc/structs_M.go
828K	pkg/oc/structs_N.go
 28K	pkg/oc/structs_O.go
 20K	pkg/oc/structs_R.go
 76K	pkg/oc/structs_S.go
 20K	pkg/oc/structs_T.go
276K	pkg/oc/union.go
6.8M	total
```

demo/uncompressed/uncompressed.go
robshakir: It wasn't -- reverted, thanks

PR #190, opened 2018-04-27
Add support for extracting the schemapath from a gNMI Path message.

robshakir:
```
 * (M) ygot/pathstrings*
  - Add support for PathToSchemaPath which extracts only element
    names, ignoring keys from the supplied path.
```

### DIFF
--- a/ygot/pathstrings.go
+++ b/ygot/pathstrings.go
@@ -46,6 +46,33 @@ func PathToString(path *gnmipb.Path) (string, error) {
 	return "/" + stdpath.Join(s...), err
 }
 
+// PathToSchemaPath returns the supplied Path as its corresponding schema path.
+// The YANG schema path removes any keys (i.e., predicates) from the path, using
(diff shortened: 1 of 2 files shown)


ygot/pathstrings.go
```diff
@@ -46,6 +46,33 @@ func PathToString(path *gnmipb.Path) (string, error) {
 	return "/" + stdpath.Join(s...), err
 }
 
+// PathToSchemaPath returns the supplied Path as its corresponding schema path.
+// The YANG schema path removes any keys (i.e., predicates) from the path, using
+// only the name.
+func PathToSchemaPath(path *gnmipb.Path) (string, error) {
```
mum4k: This is a great way of making this change as it is backwards compatible with all pre-existing code.

Thank you for looking into this.

robshakir: No worries! Thanks.

PR #191, opened 2018-04-30
Fix broken test in HEAD@master.

robshakir:
When protobufs include unknown field tags, then struct literals must use field names, rather than anonymous initialisers - this PR fixes the tests.


robshakir:
Force merging to fix head.

PR #197, opened 2018-06-07
Add an option to ignore additional fields in input JSON.

robshakir:
Where devices have augmentation to the YANG schema, additional fields may be present in the JSON supplied to `Unmarshal`. Prior to this PR, this would result in errors being thrown relating to those missing fields. This change adds a new `UnmarshalOpt` interface which can be optionally supplied to `Unmarshal` in the generated code. When supplied it ensures that an error is not thrown for such additional fields.


robshakir:
ygot misses your contributions too! :-)

PR #198, opened 2018-06-08
Add error handling for nil paths.

robshakir:
Fixes a silly NPE in the public API :-(

```
  * (M) ygot/pathstrings.go
  * (M) ygot/pathstrings_test.go
    - Add implementation and tests for a nil path being passed to
      PathToX methods.
```


robshakir:
Thanks for the review!

PR #199, opened 2018-06-12
Add helper to initialize pointer and composite struct fields

g-yusufsn:
This change is part of a bigger change towards unmarshalling a gNMI notification into a GoStruct. This helper will be used initialize fields of GoStruct corresponding to the path element while walking on the GoStruct to validate given gNMI path.

### DIFF
--- a/util/reflect.go
+++ b/util/reflect.go
@@ -15,6 +15,7 @@
 package util
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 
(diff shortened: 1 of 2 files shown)


g-yusufsn:
Trigger CLA bot.

robshakir:
Don't worry about CLA bot -- the issue is that the author on your git install is set to `yusufsn@yusufsn.svl.corp.google.com` rather than the email address associated with your GitHub repo. 

You can just do `git config --global user.email "yusufsn@google.com"` to fix this in the future.

PR #201, opened 2018-06-19
Add StringToType helper to use with gNMI notification unmarshalling

g-yusufsn:
gNMI PathElem carries keys in Key field. For struct key types, the Key field carries more than one <key, value> pairs. Source type is just string, but destination field in struct can be string, enum etc. This helper will be used to convert given string to given type while unmarshalling a gNMI notification into GoStruct.


robshakir:
Thanks for the work here!

g-yusufsn:
Thanks for reviewing!

PR #202, opened 2018-06-21
Add ygot helper to create map element value

g-yusufsn:
makeKeyForInsert function ,which already exists in list.go, have capability to create a map key given a map value. This change adds another helper to create a map value given a map[string]string assuming from the gNMI PathElem. After this change, it will be possible to create a map entry given a gNMI Path with Key field populated.

### DIFF
--- a/ytypes/common_test.go
+++ b/ytypes/common_test.go
@@ -46,6 +46,8 @@ func (EnumType) ΛMap() map[string]map[int64]ygot.EnumDefinition {
 	return globalEnumMap
 }
 
+func (EnumType) IsYANGGoEnum() {}
+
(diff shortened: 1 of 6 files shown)


robshakir:
Looks good, thanks.

We should come back after this series of CLs and add test coverage for some of the error cases that we're adding here. For example, [here](https://coveralls.io/builds/17618670/source?filename=ytypes%2Flist.go#L372) we're missing coverage for some of the errors, and in the `getKeyVal` coverage for one of the non-error return cases: https://coveralls.io/builds/17618670/source?filename=ytypes%2Flist.go#L513

PR #203, opened 2018-07-02
Add utility to validate given gNMI Path against given schema 

g-yusufsn:
- Added an internal function to use  while unmarshalling a gNMI notification into GoStruct
- Behavior of the internal function can be modified with the flags provided
- Exposed GetOrCreateNode function to use while retrieving a node corresponding to given gNMI path

This change will be followed by a set of changes that use the internal function mentioned above with different combination of flags. We aim to consolidate existing similar functions listed below;

See **NewNode** and **GetNode** in _experimental/ygotutils/getnode.go_
See **GetNodes** in _util/reflect.go_

### DIFF
--- a/util/common.go
+++ b/util/common.go
@@ -39,8 +39,8 @@ func stringMapKeys(m map[string]*yang.Entry) []string {
 
 // TODO(mostrowski): move below functions into path package.
 
-// pathMatchesPrefix reports whether prefix is a prefix of path.
-func pathMatchesPrefix(path *gpb.Path, prefix []string) bool {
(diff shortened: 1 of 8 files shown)


robshakir:
Thanks Yusuf -- can you run `gofmt` on this code please? Looks like the CI is failing because of this.

robshakir:
Oh, you just did! Thanks.

g-yusufsn:
Thanks, Rob for approving PR. Also, thanks for opening the issue. I will work on that.

PR #205, opened 2018-07-03
Fix the panic occurs when key and value fields don't match in makeKeyForInsert

g-yusufsn:
- Added utility to compare types of given reflect.Values
- Fixed the panic that occurs while setting unmatching types in makeKeyForInsert
- Add tests for TestInsertAndGetKey


robshakir:
Thanks for the work -- merged!

g-yusufsn:
Thanks for the approval!

PR #206, opened 2018-07-10
Add utility to translate given string slice as gNMI PathElem in a schema aware way

g-yusufsn:
With gNMI Path 0.4, PathElem is used to describe a path. It discriminates name and key fields. With the utility added in this change, it will be possible to translate given string slice into PathElem slice in a schema aware way.

### DIFF
--- a/ygot/pathtranslate/pathtranslate.go
+++ b/ygot/pathtranslate/pathtranslate.go
@@ -1,0 +1,118 @@
+// Copyright 2018 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


ygot/pathtranslate/pathtranslate.go
```diff
@@ -0,0 +1,118 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/openconfig/goyang/pkg/yang"
+
+	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
+)
+
+var (
+	seperator = "/"
```
robshakir: Sorry, just spotted this nit.

Should we make this a `const`, and also `s/seperator/separator/g` in this file.

g-yusufsn:
Thanks for catching those!

ygot/pathtranslate/pathtranslate.go
g-yusufsn: Thanks for catching this. Addressed in the next commits.

robshakir:
Thanks very much for fixing this, and all the work here!

PR #209, opened 2018-07-19
Add specific tests for list key mismatches.

robshakir:
These test cases verify that a map's key matches the key fields of the struct that is the map value. This functionality was already implemented, but here we add specific tests.


robshakir:
@keadwen -- PTAL. Ross was asking to verify that we make sure that a list key is the same as the value under `config` or `state` such that this could be assumed when processing the output config. If you have a moment a review here would be great.

PR #213, opened 2018-08-08
Fix panic when unmarshalling leaf-list of single-type union.

robshakir:
```
 * (M) exampleoc/...
   - Update exampleoc to include openconfig-relay-agent for use in
     ytypes schema tests.
 * (M) ytypes/leaf(_test)?.go
   - Avoid panic that was experienced when a leaf-list of a union
     that only had a single type specified in it was encountered.
   - Add testing for this case.
 * (M) ytypes/schema_tests/...
   - Add integration test for panic case.
```


robshakir:
@g-yusufsn -- would appreciate a review on this PR when you have a moment. It fixes a panic in `Unmarshal`.

robshakir:
@g-yusufsn Thanks!

PR #214, opened 2018-08-09
Issue 185: Relative leafref that traverses list boundary in an uncompressed schema fails validation

robshakir:
```
 * (M) util/reflect.go
   - Fix an issue whereby in uncompressed schemas, relative references
     would fail if they traversed a list boundary, since the path
     traversal would always try and skip the parent surrounding
     container since it assumed compression was enabled. The fix is
     to read the annotation in the schema to determine whether we
     are dealing with a compressed schema, and only make this skip
     during cases where path compression is enabled.
 * (M) util/reflect_test.go
   - Unit tests for reading the schema annotation.
 * (M) ygen/schemaparse.go
   - Switch to using a common const for the schema compression
     annotation.
 * (M) integration_tests/uncompressed/...
   - Add a specific integration test for ygen, ytypes, and util to
     cover this case. Fake input data provides partial coverage for
     this scenario, however, since it relies on generated code plus
     these being read in util, this test integrates the three.
```


robshakir:
@ostromart -- do you mind looking at this one? It's a nuance of the leafref traversal code that you originally authored - so might have the most state on :-) LMK if you don't have time, and I can look for other reviewers.

PR #217, opened 2018-08-10
Fix enumeration flake

robshakir:
This PR fixes a flake whereby an enum that was within a typedef, that is in a grouping that is used in more than one non-config/state location could be assigned a different name during code generation. The fix here is to detect such cases, and ensure that de-duplication is minimised for them. Added a specific test case to cover this.

The `codegen_test` cases have also been extended to do flake-detection.


g-yusufsn:
Thanks for the change!

robshakir:
Oops -- this had a DIFFBASE branch. Let me merge it back to `master`.

PR #225, opened 2018-09-06
Update retrieveNode to return paths.

robshakir:
```
 * (M) ygot/render.go
   - Make PathKeyFromStruct a public function. This returns the
     map[string]string required in a gNMI Notification to an
     external caller.
 * (M) ytypes/node.go
   - Update retrieveNode (and hence Get/GetOrCreateNode) to store
     the path to a retrieved node when it returns the value. This
     increases the utility of the function when in use within a
     gNMI implementation.
 * (M) ytypes/.../*test.go
   - Add test coverage for the retrieveNode function against
     crafted and generated schemas.
```


g-yusufsn:
Thanks for the change!

robshakir:
Thanks for the review!

PR #227, opened 2018-09-09
Fix go lint error, missing exported constant name.

robshakir:


g-yusufsn:
Thanks for fixing this.

PR #232, opened 2018-09-20
fix having an enum key type in a simple keyed list

g-yusufsn:
In a simple keyed list(non struct key type), key type can be an enum. During retrieveNode, we need to be using string value of the enum key while comparing against the value in gNMI path.

### DIFF
--- a/ytypes/node.go
+++ b/ytypes/node.go
@@ -15,7 +15,6 @@
 package ytypes
 
 import (
-	"fmt"
 	"reflect"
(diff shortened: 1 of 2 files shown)


robshakir:
Can you fix the merge conflict here and we'll merge to master. Thanks!

g-yusufsn:
Thanks for review, Rob! I leave the merge to you in case you want to look at the last commit.

PR #235, opened 2018-09-25
Add support for leaf getters.

robshakir:
```
* (M) ygen/*
   - Add support for leaf getters which enable the value of a leaf to be retrieved
     using Get<LeafName> methods. These getters return the Go zero value if a leaf
     is unset in the schema. Some caution must therefore be exercised such that
     applications that use them do not assume that a leaf is set simply because
     GetXXX returns a value.
   - A subsequent change will implement YANG default lookups such that GetXXX
     returns the YANG default for unset leaves.
 * (M) ygen/testdata/*
   - Mechanical change to make the receiver name for a generated Go struct
     consistent per style rules. We had a mix of s and t prior to this change.
 * (M) ygen/schema_tests/schema_test.go
   - Add tests against the generated code to ensure operation of leaf getters.
 * (M) generator/generator.go
   - Add support for the -generate_leaf_getters flag.
 * (M) exampleoc/*
   - Regenerate with leaf getters enabled.
```


robshakir:
Sorry -- this change looks huge, but is mainly regenerating the example package.

demo/gnmi_telemetry/gnmi.go
```diff
@@ -104,15 +105,30 @@ func CreateAFTInstance() (*oc.Device, error) {
...
+	nh.PushedMplsLabelStack = []oc.NetworkInstance_Afts_NextHop_PushedMplsLabelStack_Union{
+		&oc.NetworkInstance_Afts_NextHop_PushedMplsLabelStack_Union_Uint32{42},
+		&oc.NetworkInstance_Afts_NextHop_PushedMplsLabelStack_Union_Uint32{84},
 		expNull,
 	}
 
 	return d, nil
 }
+
+// addUnaryNextHop adds a next-hop-group and next-hop with the specified ID to the
```
mum4k: Are these changes related to the leaf getters?

ygen/gogen.go
```diff
@@ -637,6 +655,20 @@ func (t *{{ .Receiver }}) GetOrCreate{{ .ListName }}(
...
 	return v
 }
+`
+
+	// goLeafGetterTemplate defines a template for a function that, for a
+	// particular leaf, generates a getter method.
+	goLeafGetterTemplate = `
+// Get{{ .Name }} retrieves the value of the leaf {{ .Name }} from the {{ .Receiver }}
+// struct. Caution should be exercised whilst using this method since it will return
+// the Go zero value if the field is explicitly unset.
```
mum4k: (optional) Do we need to worry about getting out of sync with the "Go zero values" since we aren't consulting Go when setting them, instead we are creating a maps of zero values maintained within this package.

Go shouldn't really change this because of their compatibility guarantee, just something to think about.

ygen/gogen.go
```diff
@@ -637,6 +655,20 @@ func (t *{{ .Receiver }}) GetOrCreate{{ .ListName }}(
...
+func (t *{{ .Receiver }}) Get{{ .Name }}() {{ .Type }} {
```
mum4k: Would it be worth also documenting an example pattern for the client, i.e. how to check if it is set and then get it?

This is fairly obvious to any existing proto users, but might not be so for other users.

1) To check if it is set one *doesn't* use a getter.
if foo.Bar != nil ...

2) To just read regardless, one uses the getter to prevent nil panics.
foo.GetBar()

ytypes/schema_tests/validate_test.go
```diff
@@ -800,12 +802,12 @@ func TestLeafrefCurrent(t *testing.T) {
...
 
-	ygot.BuildEmptyTree(dev)
-	mi, err := dev.Mpls.Global.NewInterface("eth0.0")
+	ygot.BuildEmptyTree(ni)
+	mi, err := ni.Mpls.Global.NewInterface("eth0.0")
 	if err != nil {
 		t.Fatalf("TestLeafrefCurrent: could not add new MPLS interface, got: %v, want error: nil", err)
 	}
-	mi.InterfaceRef = &oc.Mpls_Global_Interface_InterfaceRef{
+	mi.InterfaceRef = &oc.NetworkInstance_Mpls_Global_Interface_InterfaceRef{
```
mum4k: Also unsure if this is related.

demo/gnmi_telemetry/gnmi.go
robshakir: These changes are made because the schema changed for the public OpenConfig, and to demo the leaf getters I regenerated our example schema -- which then mandated me to make these changes. I can revert the schema tests and make this separate commits if you'd prefer.

ygen/gogen.go
robshakir: Good idea. I've added a couple of sentences to this comment that covers this.

ytypes/schema_tests/validate_test.go
robshakir: Per other change, this is a schema change (we removed /mpls).

demo/gnmi_telemetry/gnmi.go
mum4k: Understood, that is fine, thanks for explaining.

PR #237, opened 2018-09-27
Add support for defaults in leaf getters.

robshakir:
This PR adds support for leaf getters to return defaults that are specified in the YANG schema. Thus, for a schema of the form:

```yang
container a-container { 
  leaf a-bool {
     type boolean;
     default true;
  }

  leaf a-string {
    type string;
    default "aardvark";
  }
}
```

then `&AContainer{}.GetABool()` will return `true` rather than `false`, and `&AContainer{}.GetAString()` will return `"aardvark`" rather than `""`.


ygen/gogen.go
```diff
@@ -1973,3 +1982,35 @@ func writeIfNotEmpty(b *bytes.Buffer, s string) {
...
+
+// quoteDefault adds quotation marks to the value string if the goType specified
+// is a string, and hence requires quoting.
+func quoteDefault(value *string, goType string) *string {
+	if value == nil {
+		return nil
+	}
+
+	if goType == "string" {
+		return ygot.String(fmt.Sprintf(`"%s"`, *value))
```
mum4k: (optional) We could replace this with %q:

return ygot.String(fmt.Sprintf("%q", *value))

ygen/yang_helpers.go
```diff
@@ -408,3 +410,30 @@ func isYANGBaseType(t *yang.YangType) bool {
...
+
+// enumDefaultValue sanitises a default value specified for an enumeration
+// which can be specified as prefix:value in the YANG schema. The baseName
+// is used as the generated enumeration name stripping any prefix specified,
+// (allowing removal of the enumeration type prefix if required). The default
+// value in the form <sanitised_baseName>_<sanitised_defVal> is returned as
+// a pointer.
+func enumDefaultValue(baseName, defVal, prefix string) *string {
+	if strings.Contains(defVal, ":") {
+		defVal = strings.SplitN(defVal, ":", -1)[1]
```
mum4k: (optional) If we are passing -1 as the split count, we could as well use strings.Split() instead of strings.SplitN().

ygen/gogen.go
robshakir: Oh, neat! I didn't know about %q. Fixed.

ygen/yang_helpers.go
robshakir: Done.

PR #238, opened 2018-09-27
Update scripts/coverage.sh to exclude tests.

robshakir:
```
 * (M) scripts/coverage.sh
  - Avoid testing coverage for test packages that are themselves
    tests.
```

This is just some internal housekeeping, since we track the coverage and report on it in PRs where it decreases significantly (as a blocking check), then we want to avoid checking files where we don't expect coverage.


g-yusufsn:
Thanks for the change!

PR #250, opened 2018-10-31
testutil enhancements; render panic fixes; error updates.

robshakir:
This PR updates the behaviour of the return arguments for some retrieval
functions to allow Subscribe implementation, adds further test util
helpers and fixes a panic in findUpdatedLeaves.

```
 * (M) testutil/testutil(_test)?.go
  - Add support for checking the equality of two slices of
    SubscribeResponse messages, and individual SubscribeResponse
    messages.
 * (M) ygot/render(_test)?.go
  - Fix a panic which occurred when a nil GoStruct was provided to
    TogNMINotifications.
 * (M) ytypes/node.go
  - Ensure that we comply with the gNMI specification as to when to
    return NotFound vs. InvalidArgument.
```

### DIFF
--- a/testutil/testutil.go
+++ b/testutil/testutil.go
@@ -35,6 +35,37 @@ func GetResponseEqual(a, b *gnmipb.GetResponse) bool {
 	return NotificationSetEqual(a.Notification, b.Notification)
 }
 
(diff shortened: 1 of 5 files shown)


g-yusufsn:
Thanks for this change!

PR #252, opened 2018-11-02
add support for unmarshalling gNMI TypedValue

g-yusufsn:
Change includes;

- Adding unmarshalling support for gNMI TypedValue
- Modifying retriveNode to use unmarshalGeneric function to unmarshal gNMI TypedValues
- Fixing a bug where we should return an error in SetNode if path points to a non-leaf schema


g-yusufsn:
I guess we are broken due to https://github.com/grpc/grpc-go/pull/2428. We need to remove it from the list of go versions we are building against.

robshakir:
Yep - I'll make this change. We should deprecate 1.8 at this point. It
seems fair enough to only handle head - 2.

Will make this change and review your PR when I am at a computer in a
couple of hours.

Thanks!
r.

g-yusufsn:
I just created PR and sent to you. Feel free to review whenever you have time. Thanks for the reply!

https://github.com/openconfig/ygot/pull/254

g-yusufsn:
Thanks for the review!

PR #255, opened 2018-11-06
Resolve a number of issues related to unions in uncompressed structs.

robshakir:
```
 * (A) uexampleeoc/...
   - Add an uncompressed example generated schema for integration tests.
 * (M) util/reflect*.go
   - Expose IsValueTypeCompatible for callers - add direct unit testing.
 * (M) ygen/*
   - Modify code generation to produce To_XXX methods for unions where
     the struct contains a union that had been previously output (i.e.,
     de-duplicated).
 * (M) ytypes/*
   - Add testing for lists that are keyed on unions and fix an issue
     whereby the key value would not be set correctly when the key
     type is a multi-type union (i.e., interface).
   - Add testing for leafrefs that point to multi-type unions where
     the values were not compared correctly.
   - Add new integration tests for an user-identified case.
```


robshakir:
Thanks for the review!

PR #257, opened 2018-12-05
 Add testutil options for JSON, and ignoring fields.

robshakir:
```
* (M) testutil/testutil.go
  - Add an options framework to testutil's comparisons which allows
    for specific fields to be diff'd using custom comparers, and
    for fields to be ignored.
  - Add a diff mechanism that compares IETF JSON values by unmarshalling
    the JSON into map[string]interface{} which is used by default to
    ensure that syntatic differences in valid but equal JSON do not
    cause the value to be considered invalid.
  - Add a diff mechanism that is schema aware in IETF JSON values, which
    unmarshals JSON into a schema tree root, and then diffs the
    resulting roots. This allows for comparisons between semantically
    equal but syntactically different JSON contents.
  - Add an option to ignore the timestamp field of a Notification.
 * (M) testutil/testutil_test.go
  - Add unit tests for above functionality.
```


robshakir:
Hi @g-yusufsn -- can you PTAL at this again? I refactored to deal with an import cycle loop that I'd caused.

Thanks!
r.

PR #260, opened 2018-12-14
Fix bug in SetNode for leafref when source leaf name != target leaf name.

robshakir:
```
Add integration test for leafref lookup.
 * (M) uexampleoc/*
  - Add platform and terminal optics to uncompressed schema.
 * (M) ytypes/*
  - Add integration test that ensures that where SetNode was used for
    a leafref where the target leaf's name didn't equal the source leaf
    name, then the target leaf is set successfully.
```


g-yusufsn:
Thanks for fixing this case!

robshakir:
Added negative tests based on our discussion

PR #261, opened 2018-12-14
Fix error formatting in Validate.

robshakir:
```
 * (M) util/errs*
  - Add functions to deduplicate util.Errors and prefix all errors in a
    supplied util.Errors.
 * (M) ytypes/{validate,container}*
  - Ensure that when an error is returned by Validate() each path
    element is not a separate error in the slice -- there was some
    re-use of the errors slice that caused this to be the case. This
    matters only to consumers that parse ytypes' return value as a
    util.Errors, string users were unaffected.
  - Ensure that where duplicate errors occur (same regexp, same schema
    path) only one error is returned. In the future, we should track the
    datatree path, but this is currently not possible without more major
    refactoring of Validate().
 * (M) ytypes/schema_tests/validate_test.go
  - Add test cases checking error output.
```

### DIFF
--- a/util/errs.go
+++ b/util/errs.go
@@ -14,6 +14,10 @@
 
 package util
 
+import (
+	"fmt"
+)
+
 // Errors is a slice of error.
 type Errors []error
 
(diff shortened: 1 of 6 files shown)


g-yusufsn:
Thanks for the work! Deduplicating errors will be really helpful for reporting.

PR #264, opened 2018-12-26
Enhance test coverage for node functions.

robshakir:
Add additional coverage for some of the functions within the `ytypes/node.go` file.


g-yusufsn:
Thanks for the change!

PR #267, opened 2019-02-06
Pin protobuf to v1.2.0

bovenyan:
Hey ygot developers. We are trying to deploy ygot as a dependency for our infra management system at scale. We would appreciate if we could pin a stable version for protobuf (v.1.2.0) instead of using the master branch.  I have tested that no test is broken with the stable protobuf version. 
Hope this can get accepted. 

Thanks!
Bo Yan

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,67 +3,209 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:107b233e45174dbab5b1324201d092ea9448e58243ab9f039e4c0f332e121e3a"
   name = "github.com/golang/glog"
   packages = ["."]
+  pruneopts = ""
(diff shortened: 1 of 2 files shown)


SVenky:
I signed it!

bovenyan:
I signed it!

robshakir:
Apologies for the delay replying. I'd prefer here to try and keep our dependencies against the master branch of these repos, such that we don't get out of sync. I think you should be able to override this within your project using the `[[override]]` directive in a local `Gopkg.toml`. Is this possible for you?

Thanks,
r.

bovenyan:
Hey @robshakir, 
Yes. That works for us. Thanks for the suggestion!

Best!
Bo Yan

PR #268, opened 2019-02-08
Implementation of getNodes, to handle also wildcards in the path.

idefixcert:
I need to get the nodes for path with wildcards as also the gnmi spec states.
e.g. interfaces/interface[name=*]
Therefore I implemented the piece of code.


idefixcert:
will come up with a new pull request

PR #270, opened 2019-02-11
Implementation of getNodes, to handle also wildcards in the path.

idefixcert:
I need to get the nodes for path with wildcards as also the gnmi spec states.
e.g. interfaces/interface[name=*]
Therefore I implemented the piece of code.


robshakir:
Thanks for the pull request!

We're not updating the package in `experimental` since we added the same functionality to the `ytypes` package. We have some handling there for wildcards, but through omission of a key. It should be relatively trivial to add support for the `*` wildcard.

Implementation should be a matter of:

 * Adding support [here](https://github.com/openconfig/ygot/blob/master/ytypes/node.go#L191) for the case where there is a single key specified but its value is explicitly "*". The same code can be used here.
 * Adding support [here](https://github.com/openconfig/ygot/blob/master/ytypes/node.go#L245) for the case where there is >1 key, but the key value is "*".

I'd propose to add an argument to [`retrieveNodeArgs`](https://github.com/openconfig/ygot/blob/master/ytypes/node.go#L32) which specifies whether to handle "*" keys, such that existing consumers who are expecting to be able to do literal matches against "*" can control this behaviour. 

I also notice that you've implemented this using a callback, rather than just returning the error. Was there a reason to make this design change?

Thanks!
r.

idefixcert:
@robshakir 
Thanx for the reply.

I will have a look at the code parts you pointed on.
So I will implement a solution there and open another pull request.

The implementation with a callback was chosen because I thought it fits more to the stream model of gnmi. But thats no problem will handle in future like it is implemented in the code you pointed me.

So should I close this pull request?

robshakir:
You can either update this PR, or we can close it and re-review in another one.

The main advantage of the callback seems to be that you potentially can send Notifications sooner whilst processing a query, rather than waiting for the query to be complete -- this might have performance advantages iff the callback allows Notification generation to be handled asynchronously. 

I'd like to split a change to support such a callback from the handling of wildcard keys represented by `*` if possible please.

Thanks -- look forward to the PR.

Cheers,
r.

idefixcert:
@robshakir I implemented your suggestions.
The tests run green.
At the moment I don't need the implementation with the callback.

idefixcert:
There is an error when build with the master repository of go, but I think thats not a problem of my commit. Can someone check that?

ytypes/node_test.go
```diff
@@ -1101,11 +1125,68 @@ func TestGetNode(t *testing.T) {
 			},
 		},
 		inPath: mustPath("/multilist[keyone=1][keytwo=2]"),
+		inArgs: []GetNodeOpt{&GetHandleWildcards{}},
```
robshakir: This looks like it changes this test case to test a different behaviour - i.e., with the wildcard handling enabled or not -- can we have two tests:
1. Explicit key match with wildcard handling off
2. Explicit key match with wildcarding handling on

ytypes/node_test.go
```diff
@@ -1101,11 +1125,68 @@ func TestGetNode(t *testing.T) {
...
+		wantTreeNodes: []*TreeNode{{
+			Data:   &multiListEntry{Keyone: ygot.Uint32(1), Keytwo: ygot.Uint32(2)},
+			Schema: multiKeyListSchema,
+			Path:   mustPath("/multilist[keyone=1][keytwo=2]"),
+		}, {
+			Data:   &multiListEntry{Keyone: ygot.Uint32(3), Keytwo: ygot.Uint32(2)},
+			Schema: multiKeyListSchema,
+			Path:   mustPath("/multilist[keyone=3][keytwo=2]"),
+		}},
```
robshakir: I'd suggest to add a test case for an incorrectly specified key name being used in both single key and multi-key scenarios with wildcard handling on.

ytypes/node.go
```diff
@@ -218,8 +220,22 @@ func retrieveNodeList(schema *yang.Entry, root interface{}, path, traversedPath
...
+				for k, v := range pe.Key {
+					pathElem.GetKey()[k] = v
+				}
+				pathElem.GetKey()[schema.Key] = keyAsString
+				nodes, err := retrieveNode(schema, listElemV.Interface(), util.PopGNMIPath(path), appendElem(traversedPath, pathElem), args)
+				if err != nil {
+					return nil, err
+				}
+
+				matches = append(matches, nodes...)
```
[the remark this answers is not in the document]
idefixcert: I thought the nodes can't be nill, we have the same situation in Line 203.

idefixcert:
Hey rob I refactored the code to the comments you made.

ytypes/node.go
```diff
@@ -186,40 +188,47 @@ func retrieveNodeList(schema *yang.Entry, root interface{}, path, traversedPath
...
-				return nil, status.Errorf(codes.NotFound, "schema key %s is not found in gNMI path %v, root %T", schema.Key, path, root)
+				if (args.handleWildcards && pathKey == "*") || keyAsString == pathKey {
+					match = true
+				}
 			}
+			if match {
+				keys, err := ygot.PathKeyFromStruct(listElemV)
+				// if err != nil {
+				// 	return nil, status.Errorf(codes.Unknown, "could not get path keys at %v: %v", traversedPath, err)
+				// }
```
idefixcert: There is a Problem with this error check.
If it turned on, we get errors in other tests.
The problem ist, the code hasn't be there before refactoring.

ytypes/node.go
```diff
@@ -218,8 +220,22 @@ func retrieveNodeList(schema *yang.Entry, root interface{}, path, traversedPath
...
+				pe := path.GetElem()[0]
+				pathElem := &gpb.PathElem{
+					Key:  map[string]string{},
+					Name: pe.Name,
+				}
+				for k, v := range pe.Key {
+					pathElem.GetKey()[k] = v
+				}
+				pathElem.GetKey()[schema.Key] = keyAsString
+				nodes, err := retrieveNode(schema, listElemV.Interface(), util.PopGNMIPath(path), appendElem(traversedPath, pathElem), args)
```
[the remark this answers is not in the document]
idefixcert: done

ytypes/node.go
robshakir: Can we leave it here please -- we need to check against `master` to see if something has changed. I'll review this.

idefixcert:
I created a new branch in my repo, and fixed all errors.
I create a new pull request

PR #272, opened 2019-02-22
Honour enum package name.

### LINKED ISSUE #271: -enum_package_name is ignored?
When passing the option -enum_package_name to proto_generator it seems to be ignored, and all the enums end up in the same package with the default name. Furthermore the enumPackageName does not seem to be used.

Also, if you run proto_generator multiple times with different values for -enum_package_name, it will overwrite the previous enum package.

robshakir:
```
 * (M) proto_generator/protogenerator.go
  - Fix bug whereby we did not honour the enum package name supplied
    to us -- fixes #271.
```


robshakir:
Thanks @g-yusufsn!

PR #277, opened 2019-03-28
Optimisation of diff performace

idefixcert:
The performance of the ygot.diff.go is poor.
Therefore I tried to optimise it a bit.

### DIFF
--- a/ygot/diff.go
+++ b/ygot/diff.go
@@ -17,6 +17,8 @@ package ygot
 import (
 	"fmt"
 	"reflect"
+	"sort"
+	"strings"
 
 	"github.com/golang/protobuf/proto"
 	"github.com/kylelemons/godebug/pretty"
@@ -210,6 +212,7 @@ func getPathSpec(ni *util.NodeInfo) (*pathSpec, error) {
 // the walk.
 func findSetLeaves(s GoStruct, opts ...DiffOpt) (map[*pathSpec]interface{}, error) {
 	pathOpt := hasDiffPathOpt(opts)
+	preventDupicatePath := map[string]bool{}
 
 	findSetIterFunc := func(ni *util.NodeInfo, in, out interface{}) (errs util.Errors) {
 		if reflect.DeepEqual(ni.StructField, reflect.StructField{}) {
@@ -242,6 +245,19 @@ func findSetLeaves(s GoStruct, opts ...DiffOpt) (map[*pathSpec]interface{}, erro
 		if err != nil {
 			return util.NewErrs(err)
 		}
+		//prevent duplicate key
+		keys := make([]string, len(vp.gNMIPaths))
+		for i, paths := range vp.gNMIPaths {
+			s, _ := PathToString(paths)
+			keys[i] = s
+		}
+		sort.Strings(keys)
+		key := strings.Join(keys, "/")
+		if _, in := preventDupicatePath[key]; in == true {
+			return
+		}
+		preventDupicatePath[key] = true
+
 		ni.Annotation = []interface{}{vp}
 
 		if util.IsNilOrInvalidValue(ni.FieldValue) || util.IsValueStructPtr(ni.FieldValue) || util.IsValueMap(ni.FieldValue) {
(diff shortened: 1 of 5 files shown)


idefixcert:
cli solved

ygot/diff.go
```diff
@@ -242,6 +245,19 @@ func findSetLeaves(s GoStruct, opts ...DiffOpt) (map[*pathSpec]interface{}, erro
 		if err != nil {
 			return util.NewErrs(err)
 		}
+		//prevent duplicate key
+		keys := make([]string, len(vp.gNMIPaths))
+		for i, paths := range vp.gNMIPaths {
+			s, _ := PathToString(paths)
```
robshakir: Can we catch the error cleanly here? I think that if we find an error here we can just return that this wasn't possible -- rather than crashing or potentially silently failing.

ygot/diff.go
```diff
@@ -242,6 +245,19 @@ func findSetLeaves(s GoStruct, opts ...DiffOpt) (map[*pathSpec]interface{}, erro
...
+			keys[i] = s
+		}
+		sort.Strings(keys)
+		key := strings.Join(keys, "/")
+		if _, in := preventDupicatePath[key]; in == true {
```
robshakir: It's idiomatic to use `ok` rather than `in` here.

robshakir: I'd suggest, `if _, ok := preventDuplicatePath[key]; ok { ... } `

ygot/diff.go
```diff
@@ -210,6 +212,7 @@ func getPathSpec(ni *util.NodeInfo) (*pathSpec, error) {
 // the walk.
 func findSetLeaves(s GoStruct, opts ...DiffOpt) (map[*pathSpec]interface{}, error) {
 	pathOpt := hasDiffPathOpt(opts)
+	preventDupicatePath := map[string]bool{}
```
robshakir: Suggest `processedPaths` -- since this is what is being stored here.

idefixcert:
Thanks for your advice.
I am fairly new to go so please forgive my idiomatic errors.

There are lot of test cases that are walking paths twice, therefore I have not added more test cases.

For the Benchmark test, I will try to come up with one.

ytypes/schema_tests/diff_benchmark_test.go
```diff
@@ -0,0 +1,44 @@
...
+	jsonFileA := "interfaceBenchmarkA.json"
+	jsonFileB := "interfaceBenchmarkB.json"
+
+	jsonA, err := ioutil.ReadFile(filepath.Join(testRoot, "testdata", jsonFileA))
+	if err != nil {
+		b.Errorf("ioutil.ReadFile(%s): could not open file: %v", jsonFileA, err)
+		return
+	}
+	deviceA := &oc.Device{}
+	err = oc.Unmarshal(jsonA, deviceA)
```
robshakir: `if err := oc.Unmarshal(...); err != nil { ... }`

ytypes/schema_tests/diff_benchmark_test.go
```diff
@@ -0,0 +1,44 @@
...
+		return
+	}
+
+	jsonB, err := ioutil.ReadFile(filepath.Join(testRoot, "testdata", jsonFileB))
+	if err != nil {
+		b.Errorf("ioutil.ReadFile(%s): could not open file: %v", jsonFileB, err)
+		return
+	}
+	deviceB := &oc.Device{}
+	err = oc.Unmarshal(jsonB, deviceB)
```
robshakir: Same comment as above re: defining a scoped `err`: `if err := ...`

ytypes/schema_tests/diff_benchmark_test.go
```diff
@@ -0,0 +1,44 @@
+package validate
+
+import (
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+
+	"github.com/idefixcert/ygot/ygot"
```
robshakir: Can you update the import path here? Travis is failing based on this import being not found during the run.

idefixcert:
> I'm interested to hear what the results of the benchmark pre- and post- your change?

Hi rob,
forgot to mention that.
My example took first 14s and after that 7s.

before:
goos: darwin
goarch: amd64
pkg: github.com/idefixcert/ygot/ytypes/schema_tests
BenchmarkDiff-12    	       1	14438320395 ns/op	2103761712 B/op	96901288 allocs/op
PASS
ok  	github.com/idefixcert/ygot/ytypes/schema_tests	14.679s
Success: Benchmarks passed.

afterwards:
goos: darwin
goarch: amd64
pkg: github.com/idefixcert/ygot/ytypes/schema_tests
BenchmarkDiff-12    	       1	7346453059 ns/op	1073764224 B/op	48254076 allocs/op
PASS
ok  	github.com/idefixcert/ygot/ytypes/schema_tests	7.585s
Success: Benchmarks passed.

robshakir:
Thanks for the changes and details. Can you fix the syntax errors so that we can merge?

Thanks!
r.

robshakir:
Thanks for the change! Merging to master. 

r.

PR #278, opened 2019-03-29
ytypes get node wildcards

idefixcert:
I implemented the wildcard search for get node.
Please have a review.


ytypes/leaf.go
```diff
@@ -886,7 +886,7 @@ func sanitizeGNMI(parent interface{}, schema *yang.Entry, fieldName string, tv *
 			// represents the division. We don't want to fail unmarshalling as float64
 			// is the best type in ygot that can represent a decimal64. So, second
 			// return value is just ignored.
-			fv, _ := new(big.Rat).SetFrac(big.NewInt(v.DecimalVal.Digits), prec).Float64()
+			fv, _ := new(big.Rat).SetFrac(big.NewInt(int64(v.DecimalVal.Digits)), prec).Float64()
```
robshakir: Why does this change need to be made please?

idefixcert: There was a compile error in this line of code when I checked out the master branch.
No the compile error is gone, I revert to the original. Sorry

idefixcert:
In principle the idea of this change is the same as in #270, but there was one error in #270.
I had to handle the traversed path specifically for the wildcard search.

I'am sorry for closing the original Pull Request, I had to resetup the project on my account, because of the module dependency in our project, I rely on that changes in our project.

robshakir:
Sorry -- I'm confused by this change. The previous one added support for wildcards in both lists that have >1 key as well as single-key lists. This one only seems to handle the single key list -- did you mean to make this change?

It seems like the single key list change can be made trivially:

```diff
diff --git a/ytypes/node.go b/ytypes/node.go
index d603cf4..87d160b 100644
--- a/ytypes/node.go
+++ b/ytypes/node.go
@@ -184,11 +184,12 @@ func retrieveNodeList(schema *yang.Entry, root interface{}, path, traversedPath
        for _, k := range rv.MapKeys() {
                listElemV := rv.MapIndex(k)
 
+               hasWildcard := true // this is a kludge to avoid implementing hasWildcard args in this diff
                // Handle lists with a single key.
                if !util.IsValueStruct(k) {
                        // Handle the special case that we have zero keys specified only when we are handling lists
                        // with partial keys specified.
-                       if len(path.GetElem()[0].GetKey()) == 0 && args.partialKeyMatch {
+                       if len(path.GetElem()[0].GetKey()) == 0 && args.partialKeyMatch || hasWildcard && path.GetElem()[0].GetKey()[schema.Key] == "*" {
                                keys, err := ygot.PathKeyFromStruct(listElemV)
                                if err != nil {
                                        return nil, status.Errorf(codes.Unknown, "could not get path keys at %v: %v", traversedPath, err)
```

since the case of traversing a wildcard single key list is already implemented for zero keys. Was there a reason that this implementation didn't work?

ytypes/node.go
```diff
@@ -257,7 +278,7 @@ func retrieveNodeList(schema *yang.Entry, root interface{}, path, traversedPath
 			if err != nil {
 				return nil, status.Errorf(codes.Unknown, "failed to convert the field value to string, field %v: %v", fieldName, err)
 			}
-			if pathKey != keyAsString {
+			if !(args.handleWildcards && pathKey == "*") && pathKey != keyAsString {
```
idefixcert: That's the change for multiple keys!

idefixcert:
> Sorry -- I'm confused by this change. The previous one added support for wildcards in both lists that have >1 key as well as single-key lists. This one only seems to handle the single key list -- did you mean to make this change?

This is still there, on line 262!
There are also test cases which verifies that.

> since the case of traversing a wildcard single key list is already implemented for zero keys. Was there a reason that this implementation didn't work?

I retried that, and this works, don't know what the previous error was.
All test cases are green.

ytypes/node.go
robshakir: Ah, thanks -- I apologise, I'd missed that this wasn't reverted.

robshakir:
Thanks for the updates -- and apologies for missing the multiple case. If you now merge `master` back into your branch, then all the tests should pass and we can go ahead and merge this.

Thanks!
r.

robshakir:
Thanks for the change and addressing the review comments -- merging to master now.

PR #282, opened 2019-04-08
Revert changes to testdata based on upstream goyang change

robshakir:
The `uses` and `augment` changes in upstream goyang were reverted -- revert the testdata to ensure that ygot doesn't assume these are on by default.


robshakir:
@aashaikh Friendly ping.

PR #285, opened 2019-05-10
Enum dedup flag

robshakir:
```
 * (M) ygen/*
  - In some schemas, a single grouping with an enumerated leaf in
    it is widely used across the schema (e.g., OpenHLTest), with
    ygen's default behaviour this would result in a single enum
    being output with the path of the first alphabetical place that
    this was used in the schema being selected as the name for the
    leaf. This pattern works well with OpenConfig, where this means
    that enumerations that are in both config and state have the same
    type, as well as those that are in, say, /bgp and /network-instance,
    however, for other schema it doesn't work well -- with essentially
    arbitrary naming of the enum being used. Naming after the grouping
    seems suboptimal since in this case, we would imply that the
    downstream developer knows about the actual YANG files (a core
    principle of the project is that this ISN'T true!). So, add a
    flag to allow this behaviour to be controlled. With the new flag
    all enums that are from the same grouping can be selected to have
    their own type generated in the code.
  - Implement the above flag, and testcases.
  - Misc. cleanup of code to meet bestpractices where the change was
    in the general vicinity.
```


wenovus:
Rob's changes LGTM, Rob please take a look at my commit.

wenovus:
There are a couple places `buildListKey` and `buildDirectoryDefinitions` where this flag is not tested, but I had trouble understanding how `yang.Node` works. It's not clear to me what it should be and **why they would be the same between two usages of the same enum**. So, I don't feel comfortable adding tests for those. Could you explain that and I'll add the tests for those.

ygen/genstate.go
```diff
@@ -373,13 +376,11 @@ func (s *enumGenState) identityrefBaseTypeFromIdentity(i *yang.Identity, noUnder
...
-// output name.
-func (s *enumGenState) resolveEnumName(e *yang.Entry, compressPaths, noUnderscores bool) string {
+// enumIdentifier takes in an enum entry and returns a unique identifier for
+// that enum constructed using its path. This identifier would be the same for
+// an enum that's used in two different places in the schema.
+func enumIdentifier(e *yang.Entry, compressPaths bool) string {
+	definingModName := genutil.ParentModulePrettyName(e.Node)
 	// It is possible, given a particular enumerated leaf, for it to appear
 	// multiple times in the schema. For example, through being defined in
 	// a grouping which is instantiated in two places. In these cases, the
```
wenovus, line 386: As above, why is it two usages will have the same `yang.Node`? Thanks.

robshakir: When goyang is parsing a YANG schema, then it goes through and finds the nodes that are defined in the schema - which is in relation to what is actually written in the YANG file. When it converts this tree to entries, it then builds the tree such that the schema tree is expanded out -- so if we have:

```yang
grouping foo {
   leaf a { type enumeration { enum A; } }
}

container one { uses foo; }
container two { uses foo; }
```

then there is exactly one `yang.Node` which represents the `a` defined in the `foo` grouping, but there are two entries for `/one/a` and `/two/a`. 

Prior to this change we always de-duplicated these enumerated types, so that there is an enumeration that represents `/foo/a` -- this makes sense in OpenConfig for example, where we might have `/foo/state/an-enum` and `/foo/config/an-enum`, and we don't really want to refer to those with different types. In other schemas, like the example above, de-duplicating these means that we would choose one type for `/one/a` and `/two/a`. Since these types are named after the schema path, then this meant that we had very confusing naming for the enumerations.

robshakir:
Thanks for looking at this -- let me know if you'd prefer for me to look at the missing coverage.

ygen/genstate.go
wenovus: Thanks! I think I got it now. The most time-consuming parts were
1. Understanding the tests.
2. Finally deciding that some inputs are not worth testing since they're just pass-through parameters to lower-level functions.

ygen/enumgen.go
```diff
@@ -347,31 +347,81 @@ func (s *enumGenState) resolveEnumName(e *yang.Entry, compressPaths, noUnderscor
...
+		compressName = fmt.Sprintf("%s_%s_%s", yang.CamelCase(definingModName), yang.CamelCase(e.Parent.Parent.Name), yang.CamelCase(e.Name))
 		if noUnderscores {
-			name = strings.Replace(name, "_", "", -1)
+			compressName = strings.Replace(compressName, "_", "", -1)
 		}
-		uniqueName := genutil.MakeNameUnique(name, s.definedEnums)
-		s.uniqueEnumeratedLeafNames[identifierPath] = uniqueName
+
+		if skipDedup {
+			// If using compression and duplicating, then we add
```
robshakir, line 401: Clarified this wording here "and duplicating" seemed odd to me -- so I expanded it out to say that we're explicitly outputting a type per leaf.

ygen/enumgen_test.go
```diff
@@ -907,6 +911,341 @@ func TestFindEnumSet(t *testing.T) {
...
+				entry: &yang.Entry{
+					Name: "enumeration-leaf",
+					Type: &yang.YangType{
+						Enum: &yang.EnumType{},
+					},
+				},
+			},
+		},
+	}, {
+		name: "two enums with deduplication disabled, and where duplication occurs for both compressed and decompressed",
```
robshakir, line 999: (comment) This is an interesting corner-case, since in a compressed schema, this enum type would never actually be referenced. For example:

```yang
grouping foo-config {
   leaf a { type enumeration { enum ONE; } }
}

container foo {
   container config { uses foo-config; }
   contianer state { uses foo-state; }
}
```

In this case, compression will remove `/foo/state/a` but `/foo/state/a` will still create `Foo_State_A` in the output code.

I don't think we need to worry about this - because we're explicitly doing what the user asked, it just doesn't really make a huge amount of sense :-)

robshakir:
Merging this PR based on discussion with @wenovus and another look. Thanks for the collaboration here Wen.

Issue #291, opened 2019-07-07
Unable to unmarshal BoolVal to Yempty (GNMIEncoding)


cho-m:
Currently in `ytypes/leaf.go`, the `gNMIToYANGTypeMatches` and `sanitizeGNMI` functions do not support `yang.Yempty`.

A simple testcase to check this:
```diff
$ git diff
diff --git a/ytypes/leaf_test.go b/ytypes/leaf_test.go
index 8d0181a..00ce92d 100644
--- a/ytypes/leaf_test.go
+++ b/ytypes/leaf_test.go
@@ -1657,6 +1657,16 @@ func TestUnmarshalLeafGNMIEncoding(t *testing.T) {
                        },
                        wantVal: &LeafContainerStruct{BoolLeaf: ygot.Bool(true)},
                },
+               {
+                       desc:     "success gNMI BoolVal to Yempty",
+                       inSchema: typeToLeafSchema("empty-leaf", yang.Yempty),
+                       inVal: &gpb.TypedValue{
+                               Value: &gpb.TypedValue_BoolVal{
+                                       BoolVal: true,
+                               },
+                       },
+                       wantVal: &LeafContainerStruct{EmptyLeaf: YANGEmpty(true)},
+               },
                {
                        desc:     "success gNMI StringVal to Ystring",
                        inSchema: typeToLeafSchema("string-leaf", yang.Ystring),
```

Test result:
```
--- FAIL: TestUnmarshalLeafGNMIEncoding (0.00s)
    leaf_test.go:1848: success gNMI BoolVal to Yempty: unmarshalLeaf(&{<nil> <nil> empty-leaf    [] Leaf unset <nil> unset map[]  0xc00042efc0 [] <nil> <nil> [] [] [] [] map[] [] map[] map[] <nil>}, &{<nil> [] <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> [] <nil> <nil> 0 <nil> <nil> false [] []}, bool_val:true , GNMIEncoding): got err=failed to unmarshal &{true} into empty, want err=nil
E0706 17:29:38.442149   18766 util_types.go:170] unexpected type bits in yangBuiltinTypeToGoPtrType
FAIL
exit status 1
FAIL	github.com/openconfig/ygot/ytypes	0.063s
```

PR #292, opened 2019-07-07
Fix unmarshal of empty type leaf value for gNMI encoding

cho-m:
Possible fix for #291 

Need to confirm if any special handling of `bool_val:false` for empty type since YANG empty type really only has 2 possible states ([1]doesn't exist, or [2]true) while YANG boolean type has 3 possible states ([1]doesn't exist, [2]true, or [3]false)

### DIFF
--- a/ytypes/leaf.go
+++ b/ytypes/leaf.go
@@ -854,7 +854,7 @@ func sanitizeGNMI(parent interface{}, schema *yang.Entry, fieldName string, tv *
 	}
 
 	switch ykind {
-	case yang.Ybool:
+	case yang.Ybool, yang.Yempty:
 		return tv.GetBoolVal(), nil
 	case yang.Ystring:
 		return tv.GetStringVal(), nil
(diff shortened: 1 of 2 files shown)


ytypes/leaf.go
```diff
@@ -902,7 +902,7 @@ func sanitizeGNMI(parent interface{}, schema *yang.Entry, fieldName string, tv *
 func gNMIToYANGTypeMatches(ykind yang.TypeKind, tv *gpb.TypedValue) bool {
 	var ok bool
 	switch ykind {
-	case yang.Ybool:
+	case yang.Ybool, yang.Yempty:
```
robshakir: I suspect that we really need to make `yang.Yempty` always set the field to `true` if we receive this. Since `YANGEmpty` isn't a pointer in the generated code, then if we receive an empty field that has a bool value of `false` then this won't be something that the client application can tell was received. Could you add this handling as a special case please?

wenovus: Actually there already appears to be handling for this case:
https://github.com/openconfig/ygot/blob/f01913aa54675e818f97bf5a55692755e2b6ff10/util/reflect.go#L256-L262

When unmarshalling an empty leaf from a `TypedValue` in the current code, `unmarshalScalar` returns the non-pointer bool value, which `unmarshalLeaf` then takes and then feeds into `InsertIntoStruct` (directly or indirectly) to reach this special handling code.

This change doesn't support unmarshalling empty leafs (either JSON or gNMI) into a union type. However given its complexity I think it deserves to be addressed separately.

wenovus (approved):

ytypes/leaf.go
wenovus: I explored the question of unions containing empty in #582.

PR #294, opened 2019-07-07
Fix unmarshal/validate of binary type leaf value

### LINKED ISSUE #293: Failing to unmarshal Binary type & validation also broken
Currently ygot fails on unmarshaling binary type leafs due to assuming `[]byte` rather than derived type `Binary`.

Also, even if the unmarshaling is fixed, the validation is similarly broken for binary type.

### Example Data

The following is a basic YANG model to exhibit this:
```yang
module a {
  prefix a;
  namespace "urn:a";
  revision 2019-07-06 {
    description "Initial";
  }

  container top {
    leaf test {
      type binary {
        length 2;
      }
    }
  }
}
```

Using a sample JSON encoded data of:
```json
{
  "a:top": {
    "test": "NDI="
  }
}
```

### Unmarshal issue

Unmarshal currently fails with: **cannot assign value [52 50] (type []uint8) to struct field Test (type a.Binary) in struct \*a.A_Top**

The problem is in `reflect.InsertIntoStruct`.
Since `Binary` is derived type, a cast similar to `YANGEmpty` is required.

### Validation issue

Validate currently passes no matter the input length of binary data. For example, changing the value to JSON binary input to `"NDIgNDIgNDIgNDIgNDI="` will still pass.

The first problem here is in `validateLeaf` of leaf.go. The `rv` value passed to `return util.NewErrs(validateBinary(schema, rv))` is `nil` since there it is never set.

The second problem is `validateBinary` of `binary_type.go`. Here, the data is processed as if it was the base type `[]byte` rather than the derived type `Binary`.

### Other changes required

Various tests need to be changed to use some variant of `Binary` as they are incorrectly passing.

cho-m:
Fixes #293 

Also adds a few other tests for some code that wasn't covered.


util/reflect.go
```diff
@@ -266,6 +266,14 @@ func InsertIntoStruct(parentStruct interface{}, fieldName string, fieldValue int
 		v = nv
 	}
 
+	// YANG binary fields are represented as a derived []byte value defined in the
+	// generated code. Here we cast the value to the type in the generated code.
+	if ft.Type.Kind() == reflect.Slice && t.Kind() == reflect.Slice && ft.Type.Elem().Kind() == reflect.Uint8 && t.Elem().Kind() == reflect.Uint8 {
```
robshakir: It's worth adding a comment here that this will actually trigger for slices of `uint8` -- i.e., `leaf-list` nodes of type `uint8`. We can safely treat the two as the same though since `[]byte` and `[]uint8` are equivalent in the underlying implementation (see https://play.golang.org/p/hpD9IeYY9hH for a quick check that `SetBytes` won't panic).

util/reflect_test.go
```diff
@@ -568,6 +582,62 @@ func TestIsValueTypeComaptible(t *testing.T) {
...
+	}
+
+	tests := []struct {
+		desc         string
+		parentStruct interface{}
+		fieldName    string
+		fieldValue   interface{}
+		wantVal      interface{}
+		wantErr      string
+	}{
```
robshakir: Can we add a test here that covers the `[]uint8` case that I mentioned above? Thanks for adding this coverage!

ytypes/binary_type.go
```diff
@@ -35,12 +37,13 @@ func validateBinary(schema *yang.Entry, value interface{}) error {
 	}
 
 	// Check that type of value is the type expected from the schema.
-	binaryVal, ok := value.([]byte)
-	if !ok {
+	t := reflect.TypeOf(value)
```
robshakir: We should probably create a helper here -- `isBinaryField(reflect.Type)` and re-write this:

```golang
if t := reflect.TypeOf(value); !isBinaryType(t) { ...
```

Since we use the same logic below.

ytypes/binary_type.go
```diff
@@ -58,21 +61,24 @@ func validateBinarySlice(schema *yang.Entry, value interface{}) error {
...
+	tbl := make(map[string]bool, v.Len())
+	for i := 0; i < v.Len(); i++ {
+		val := v.Index(i)
+		if err := validateBinary(schema, val.Interface()); err != nil {
 			return fmt.Errorf("invalid element at index %d: %v", i, err)
 		}
-		if tbl[string(val)] {
-			return fmt.Errorf("duplicate binary type: %v for schema %s", val, schema.Name)
+		binaryVal := val.Bytes()
+		if tbl[string(binaryVal)] {
```
robshakir:
```golang
if tbl[string(val.Bytes())] { ...
```

util/reflect.go
cho-m: Added comment for `[]uint8` behavior

util/reflect_test.go
cho-m: Added test case for corresponding scenario.

ytypes/binary_type.go
cho-m: Refactored logic to dedicated functions for:
* `isBinaryType` - to do type test previously handled by `.([]byte)`
* `isBinarySliceType` - to do type test previously handled by `.([][]byte)`

ytypes/binary_type.go
cho-m: I had done assignment to `binaryVal` since there is a few more lines of logic that use the value.
Let me know if the preference would be to just use `val.Bytes()` in the corresponding locations.

robshakir: Ah, sorry, missed that you used it again. LGTM.

util/reflect.go
robshakir: Great, thanks.

util/reflect_test.go
robshakir: Perfect, thanks.

PR #299, opened 2019-07-23
Change leafref memo data struct to use a tree parallel to the NodeInfo tree

wenovus:
- Use a tree instead of a hash of a "reflect.Value()" to identify datanodes, which may aid debugging a little by making what has been memoized more easily readable.


wenovus:
Note I don't know why the coverage decreased, when I run the test with the apparently uncovered code deleted, the test fails.

So, the coverage has been verified manually.

PR #306, opened 2019-08-16
Create genutil package; rename and export yangDirectory as Directory.

wenovus:
Rename and export yangDirectory as Directory along with component types. Also add genutil and moved several utility functions from ygen into genutil. A couple new tests were added; a few util functions don't yet have tests and are marked with TODO.

Directory is moved to a new file directory.go; GetOrderedDirectories and GetOrderedFields exported helpers are also added with tests into that file.


robshakir:
Can you remove the BUILD file from this PR please? :-)

ygen/codegen.go
```diff
@@ -472,6 +392,52 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
 	}, nil
 }
 
+func generateEnumCode(goEnums map[string]*yangEnum) ([]string, string, util.Errors) {
```
robshakir: Sorry that I missed this before, but can we ensure that this function is commented?

ygen/codegen.go
```diff
@@ -703,15 +669,15 @@ type mappedYANGDefinitions struct {
...
+// mappedDefinitions finds the set of directory and enumeration entities
 // that are mapped to objects within output code in a language agnostic manner.
 // It takes:
 //	- yangFiles: an input set of YANG schema files and the paths that
 //	- includePaths: the set of paths that are to be searched for included or
 //	  imported YANG modules.
 //	- cfg: the current generator's configuration.
-// It returns a mappedYANGDefinitions struct populated with the directory and enum
-// entries in the input schemas, along with the calculated schema tree.
+// It returns a mappedYANGDefinitions struct populated with the directory & enum
```
robshakir: `s/&/,/`

PR #309, opened 2019-09-05
Set MappedType parameters correctly when a union contains a single type

wenovus:
Currently, `isEnumeratedValue,` `defaultValue,` and sometimes `zeroValue` are ignored when this situation occurs, which may lead to issues later when these values are used.

e.g. previously, a union containing just an enum generates a pointer field due to `isEnumeratedValue` being false. Manual testing shows it's no longer a pointer, consistent with regular enum leaves.


wenovus:
I just realized that the same may need to be done for proto generation as well, need to look into that.

wenovus:
Actually it seems that proto generation is doing this correctly -- I've added tests for this case.

robshakir:
LGTM, thanks for the change Wen!

PR #311, opened 2019-09-06
Export remaining MappedType fields

wenovus:
While MappedType is exported, these unexported fields make this type less usable in other packages. For example, a user package can't easily check the `isEnumeratedValue` field to check whether it's an enum. While it is true that an accessor function can be used instead, you then can't set this field in tests.

My original thought behind only exporting NativeType was that these are "internal states", but since these are just accompanying properties of the type that describe it further, this argument doesn't make much sense, so this is just correcting a problem.

p.s. This change would resolve another issue you raised.

### DIFF
--- a/ygen/goelements.go
+++ b/ygen/goelements.go
@@ -331,7 +331,7 @@ func (s *genState) yangTypeToGoType(args resolveTypeArgs, compressOCPaths bool) 
 //	}
 //
 // Is returned with a goType of Bar_Foo_Union (where Bar_Foo is the schema
-// path to an element). The UnionTypes are specified to be string and int8.
+// path to an element). The unionTypes are specified to be string and int8.
 //
 // The compressOCPaths argument specifies whether OpenConfig path compression
 // is enabled such that the name of enumerated types can be calculated correctly.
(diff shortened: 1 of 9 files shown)


ygen/goelements.go
```diff
@@ -331,22 +331,22 @@ func (s *genState) yangTypeToGoType(args resolveTypeArgs, compressOCPaths bool)
...
+// path to an element). The UnionTypes are specified to be string and int8.
 //
 // The compressOCPaths argument specifies whether OpenConfig path compression
 // is enabled such that the name of enumerated types can be calculated correctly.
 //
 // goUnionType returns an error if mapping is not possible.
 func (s *genState) goUnionType(args resolveTypeArgs, compressOCPaths bool) (*MappedType, error) {
 	var errs []error
-	unionTypes := make(map[string]int)
+	UnionTypes := make(map[string]int)
```
robshakir: It looks like there's a slightly over-aggressive regexp fix applied here -- within this function we shouldn't need to rename this `UnionTypes` -- i.e., the capitalisation here doesn't make a difference to the code.

ygen/goelements.go
```diff
@@ -355,23 +355,23 @@ func (s *genState) goUnionType(args resolveTypeArgs, compressOCPaths bool) (*Map
 
 	// Zero value is set to nil, other than in cases where there is a single type in
 	// the union.
-	zeroValue := "nil"
+	ZeroValue := "nil"
```
robshakir: Same comment here as for `UnionTypes` above.

ygen/gogen.go
```diff
@@ -1280,10 +1280,10 @@ func writeGoStruct(targetStruct *Directory, goStructElements map[string]*Directo
 			scalarField := true
 			fType := mtype.NativeType
 			schemapath := util.SchemaTreePathNoModule(field)
-			zeroValue := mtype.zeroValue
-			defaultValue := goLeafDefault(field, mtype)
+			ZeroValue := mtype.ZeroValue
```
robshakir: Same comment here as in `genstate.go`, we don't need to change the naming here within the function.

ygen/protoelements.go
```diff
@@ -203,22 +203,22 @@ func (s *genState) yangTypeToProtoScalarType(args resolveTypeArgs, pargs resolve
 //	int32 a_int32 = NN;
 // }
 //
-// The MappedType's unionTypes can be output through a template into the oneof.
+// The MappedType's UnionTypes can be output through a template into the oneof.
 func (s *genState) protoUnionType(args resolveTypeArgs, pargs resolveProtoTypeArgs) (*MappedType, error) {
-	unionTypes := make(map[string]*yang.YangType)
-	if errs := s.protoUnionSubTypes(args.yangType, args.contextEntry, unionTypes, pargs); errs != nil {
+	UnionTypes := make(map[string]*yang.YangType)
```
robshakir: Same comment again here as the other files - `s/UnionTypes/unionTypes/`

PR #317, opened 2019-10-08
Fix targets to avoid unstable string comparisons

wenovus:
Syncing changes. Description below:

The Message.String method is intended for debugging and does not guarantee
output stability. Code that relies on the output being stable for all time is
brittle and has unintentionally become a change detector for the proto package.


wenovus:
Note here one file's sync is purposely omitted, this only syncs the changes regarding Message.String().

PR #319, opened 2019-10-17
Add end-to-end integration testing for ypathgen and other polishings.

wenovus:
List of changes:
- Merge CompressOCPaths and ExcludeState ygen configurations into
  CompressBehaviour enum, which also adds ability to prefer the state
  version of child nodes when it exists under both the config and state
  containers as an intended config and operational state respectively.
- Update FindAllChildren() to prefer the state version when determining
  the direct children of a node under PreferOperationalState
  CompressBehaviour.
- Fix small bug in FindAllChildren() whereby *all* duplicates were being
  accepted without error, which masked real duplicates between a
  higher-level leaf and the state container (now the generalized
  deprioritized container).
- Update ypathgen generation to always use PreferOperationalState
  CompressBehaviour.
- Add generator.go, and update.sh for ypathgen path creation tests -- an
  end-to-end integration test for the path API generated by ypathgen.
- Refactor IsScalarField() out of writeGoStruct() in ygen.
- Refactor TestBuildDirectoryDefinitions to use cmp and t.Run.
- Augment GetDirectories to GetDirectoriesAndLeafTypes to return a
  leafTypeMap -- this is meant to provide information to callers of
  ypathgen so that they have more data for augmenting the generated code
  output of ypathgen.


robshakir:
Hi Wen,

Thanks for the changes again, these LGTM. It does look like our unit test coverage is suffering though. Let's figure out whether we can keep it above 90%. We don't need to do this right now, but we should create issues for:

 * Tracking coverage within `genutil`, it seems that [`names.go`](https://coveralls.io/builds/26382504/source?filename=genutil/names.go) and [`common.go`](https://coveralls.io/builds/26382504/source?filename=genutil/common.go) aren't well covered today. 
 * We should exclude `ypathgen/generator` from the coverage metric, since we know that this is just a binary calling a known library. You can add it to the `-ignore` list in [`scripts/coverage.sh`](https://github.com/openconfig/ygot/blob/master/scripts/coverage.sh#L5).
 
We can figure out if there are other places that we can beef up coverage after getting the complete change set submitted.

Cheers,
r.

wenovus:
> Hi Wen,
> 
> Thanks for the changes again, these LGTM. It does look like our unit test coverage is suffering though. Let's figure out whether we can keep it above 90%. We don't need to do this right now, but we should create issues for:
> 
> * Tracking coverage within `genutil`, it seems that [`names.go`](https://coveralls.io/builds/26382504/source?filename=genutil/names.go) and [`common.go`](https://coveralls.io/builds/26382504/source?filename=genutil/common.go) aren't well covered today.
> * We should exclude `ypathgen/generator` from the coverage metric, since we know that this is just a binary calling a known library. You can add it to the `-ignore` list in [`scripts/coverage.sh`](https://github.com/openconfig/ygot/blob/master/scripts/coverage.sh#L5).
> 
> We can figure out if there are other places that we can beef up coverage after getting the complete change set submitted.
> 
> Cheers,
> r.

Most of the newly uncovered lines seem to be errors, or trivial functions whose tests to me look like they would lean towards change detectors. I can open an issue for this, and I'll look for better places to write tests, but do you think there is value in getting those aforementioned places covered? e.g. https://coveralls.io/builds/26388082/source?filename=genutil/common.go#L124 https://coveralls.io/builds/26388082/source?filename=ypathgen/pathgen.go#L125

PR #321, opened 2019-10-23
Fix exampleocpath package name

wenovus:
This is a typo fix.


robshakir:
I think this is because we don't have `go build ./...` in `.travis.yml`, just `go test ./...` -- do we expect that any of these tests depend on exampleocpath today?

wenovus:
ypathgen/path_tests depends on exampleocpath, that's why it was added.

go help test:
> 'Go test' recompiles each package along with any files with names matching
> the file pattern "*_test.go".

Since go test recompiles all the files that's actually not the issue.

Did some playing and I think I know why now -- I didn't know the package clause *can* specify a name different than the containing folder -- when you import a folder named "pkg", but where the file's package clause uses "a", all the qualifiers actually have to be "a", unless a package alias is used.

Anyways, I think we have to make the package name the same to prevent inconveniences.

PR #330, opened 2019-11-15
Remove Support for Go 1.11 and Use io.StringWriter instead of *bytes.Buffer to be more generic.

wenovus:


robshakir:
Hi Wen,

This looks fine to me -- but I'm just wondering what the motivation for the change was (if you can remember)?

Thanks,
r.

wenovus:
It's in order to use `strings.Builder` instead of `bytes.Buffer`.

Since we're only using the `WriteString()` and `String()` functionalities, `strings.Builder` is more efficient and natural. I didn't remove all current usages as speed is not important, but don't want future writers to be stuck with using `bytes.Buffer`.

PR #331, opened 2019-11-15
Support modules with double spaces between key names.

robshakir:
The YANG grammar in rfc6020 specifies:

```
   sep                 = 1*(WSP / line-break)
                         ; unconditional separator
```

where 1* means "at least one". And:

```
   key-stmt            = key-keyword sep key-arg-str stmtend

   key-arg-str         = < a string that matches the rule
                           key-arg >

   key-arg             = node-identifier *(sep node-identifier)
```

Which technically allows `"key1 key2"` and `"key1  key2"` (note the latter is double-spaced) to be semantically equivalent when specifying multiple spaces. This PR changes ygen's logic such that we allow such lists to be generated. Before this change, the latter would cause an error due to not being able to find a key named `""`.


ygen/goelements.go
```diff
@@ -465,6 +465,9 @@ func (s *genState) buildListKey(e *yang.Entry, compressOCPaths bool) (*YangListA
 	var errs []error
 	keys := strings.Split(e.Key, " ")
 	for _, k := range keys {
+		if k == "" {
+			continue
+		}
```
wenovus: I think this doesn't handle tabs or newline:
```
line-break          = CRLF / LF
WSP                 = SP / HTAB
                         ; whitespace
```
I didn't figure out how to deal with CRLF, but maybe [strings.Fields](https://golang.org/pkg/strings/#Fields), which uses [unicode.IsSpace](https://golang.org/pkg/unicode/#IsSpace) as the separator, is sufficient.

robshakir: Good catch -- let me look in goyang -- I've never seen a CRLF in this field - I'm not sure whether goyang does anything to it before it hands it downstream.

wenovus: It looks like goyang removes tabs, but keeps newlines when it parses a key string with these whitespaces. So, I think `strings.Fields` is an appropriate function to use. See https://github.com/openconfig/goyang/pull/107

testdata/modules/openconfig-withlist.yang
```diff
@@ -40,7 +40,9 @@ module openconfig-withlist {
 
       container b {
         list multi-key {
-          key "key1 key2";
+          key "key1
```
aashaikh: could also have a specific test for multiple spaces here; but since the new implementation treats both cases the same, probably not needed.

wenovus: I added spaces as well to the irregular spacing integration test

PR #337, opened 2019-12-14
Remove global var generatedUnionInterfaces and use a local var instead

wenovus:
This is a refactoring effort. genState's numerous global state makes the code somewhat harder to understand -- this helps by removing the one that doesn't need to be a global variable.

### DIFF
--- a/ygen/gogen.go
+++ b/ygen/gogen.go
@@ -1212,6 +1212,9 @@ func writeGoStruct(targetStruct *Directory, goStructElements map[string]*Directo
 	// genUnions stores the set of multi-type YANG unions that must have
 	// code generated for them.
 	genUnions := []goUnionInterface{}
+	// genUnionSet stores a set of union type names such that we can process
+	// each appearance of a union type within the struct once and only once.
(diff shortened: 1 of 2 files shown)


ygen/gogen.go
```diff
@@ -1296,48 +1299,48 @@ func writeGoStruct(targetStruct *Directory, goStructElements map[string]*Directo
...
-						ParentReceiver: targetStruct.Name,
-					}
+			// Only if this union has more than one subtype do we generate the union;
+			// otherwise, we use that subtype directly.
+			// Also, make sure to process a union type once and only once within the struct.
+			// Even if the union has already been processed in another struct, we still need
+			// to generate the union helper with this struct as the receiver and do other
+			// processing as well. On the other hand, if the union type is used by multiple
+			// fields, we assume that it's due to a leafref and use the same union name
+			// without doing further code generation. XXX(wenbli): It's possible that it's a
```
progressnerd: Is this username supposed to be here?

wenovus: Rob put his so I assumed that it's fine here as well.

PR #342, opened 2020-01-16
Add SplitFiles method to GeneratedPathCode

wenovus:
In order to allow consumers to roughly limit the size of each file.

- TestGeneratePathCodeSplitFiles tests using the openconfig-simple schema, which contains **4 directory entries**: 
  - Device
  - Parent
  - Parent_Child
  - RemoteContainer

These 4 are split between the various files.

I have tested on the real OpenConfig schema and splitting by counting structs does have a large variance, but it's a good enough approximation. Counting bytes is more accurate, but I think it's slightly more complex, and also makes the tests a bit harder to write and more of a change detector, so that's why I'm not enthusiastic about doing it that way.


ypathgen/pathgen.go
```diff
@@ -221,6 +221,37 @@ func (genCode GeneratedPathCode) String() string {
...
+
+	files := make([]string, 0, fileN)
+	structsPerFile := structN / fileN
+	var gotCode strings.Builder
+	gotCode.WriteString(genCode.CommonHeader)
+	gotCode.WriteString(genCode.OneOffHeader)
+
+	for i, gotStruct := range genCode.Structs {
+		// The last file contains the remainder of the structs.
+		if i%structsPerFile == 0 && i/structsPerFile > 0 && i/structsPerFile < fileN {
```
progressnerd: I would replace "i/structsPerFile > 0 && i/structsPerFile < fileN" with "i >= structsPerFile && i < fileN*structsPerFile"

PR #344, opened 2020-02-05
Add TolerateJSONInconsistencies option for SetNode

wenovus:
TolerateJSONInconsistencies signals SetNode to tolerate inconsistencies
for val as if it were converted from JSON. As of right now, this is
specifically to deal with uint values being streamed as positive int
values.

The unexported encoding gNMIEncodingWithJSONTolerance is added to allow this exception handling
to occur within ytype's unmarshal codebase.

### DIFF
--- a/ytypes/leaf.go
+++ b/ytypes/leaf.go
@@ -497,9 +497,7 @@ func unmarshalUnion(schema *yang.Entry, parent interface{}, fieldName string, va
 	var valueStr string
 	var ok bool
 	switch enc {
-	case GNMIEncoding:
-		fallthrough
-	case gNMIEncodingWithJSONTolerance:
+	case GNMIEncoding, gNMIEncodingWithJSONTolerance:
 		var sv *gpb.TypedValue_StringVal
 		if sv, ok = value.(*gpb.TypedValue).GetValue().(*gpb.TypedValue_StringVal); ok {
 			valueStr = sv.StringVal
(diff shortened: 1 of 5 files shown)


ytypes/leaf.go
```diff
@@ -665,11 +665,13 @@ func unmarshalScalar(parent interface{}, schema *yang.Entry, fieldName string, v
 	case JSONEncoding:
 		return sanitizeJSON(parent, schema, fieldName, value)
 	case GNMIEncoding:
+		fallthrough
+	case gNMIEncodingWithJSONTolerance:
```
progressnerd: group in prior case with a comma

ytypes/leaf.go
```diff
@@ -831,6 +840,13 @@ func gNMIToYANGTypeMatches(ykind yang.TypeKind, tv *gpb.TypedValue) bool {
 		_, ok = tv.GetValue().(*gpb.TypedValue_IntVal)
 	case yang.Yuint8, yang.Yuint16, yang.Yuint32, yang.Yuint64:
 		_, ok = tv.GetValue().(*gpb.TypedValue_UintVal)
+		if !ok && jsonTolerance {
+			// Allow positive ints to be treated as uints.
+			if v, intOk := tv.GetValue().(*gpb.TypedValue_IntVal); intOk && v.IntVal >= 0 {
+				tv.Value = &gpb.TypedValue_UintVal{UintVal: uint64(v.IntVal)}
```
progressnerd: ok, tv.Value = true, &gpb.TypedValue_UintVal{UintVal: uint64(v.IntVal)}
and then delete the next return statement, so it returns at the bottom like all the other cases

PR #347, opened 2020-02-13
Fix incorrect validation of unions containing a single enumerated value

wenovus:
#309 created this validation corner case -- fixing now.

### DIFF
--- a/ytypes/leaf.go
+++ b/ytypes/leaf.go
@@ -216,18 +216,16 @@ func validateUnion(schema *yang.Entry, value interface{}) util.Errors {
 	}
 
 	util.DbgPrint("validateUnion %s", schema.Name)
-	rkind := reflect.TypeOf(value).Kind()
-	var v reflect.Value
-	switch rkind {
+	v := reflect.ValueOf(value)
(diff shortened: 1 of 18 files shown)


ytypes/leaf.go
```diff
@@ -216,17 +216,27 @@ func validateUnion(schema *yang.Entry, value interface{}) util.Errors {
 	}
 
 	util.DbgPrint("validateUnion %s", schema.Name)
-	// Must be a ptr - either a struct ptr or Go value ptr like *string.
-	// Enum types are also represented as a struct for union where the field
-	// has the enum type.
-	if reflect.TypeOf(value).Kind() != reflect.Ptr {
-		return util.NewErrs(fmt.Errorf("wrong value type for union %s: got: %T, expect ptr", schema.Name, value))
+	rkind := reflect.TypeOf(value).Kind()
+	var v reflect.Value
```
progressnerd: maybe tighten it up with:

v := reflect.ValueOf(value)
switch v.Type().Kind() {
case reflect.Ptr:
  v = v.Elem()
case reflect.Int64:
default:
  return util.NewErrs(...)
}

PR #355, opened 2020-03-07
Add option to use builder format for list path API

wenovus:
A threshold knob is added to the generator script to control from what
length on the builder API format should be used.

The idea is that having a huge combination of list path constructor
functions when the number of keys is large can be confusing to the
reader, and cause too much code to be generated: If there are 3 keys,
then there would be 2^3 = 8 different functions for the list, 20 would
mean 1 million.

Instead of forcing the user to select the right one among the masses,
the user can simply set the value of each key instead.

The downside of the builder API format is that it's not possible to give
back the user the non-wildcarded type even if all keys have been specified.


ypathgen/pathgen.go
```diff
@@ -740,6 +766,74 @@ func generateChildConstructorsForLeafOrContainer(methodBuf *strings.Builder, fie
...
+	if err := goPathChildConstructorTemplate.Execute(methodBuf, fieldData); err != nil {
+		errors = append(errors, err)
+	}
+
+	// The root node doesn't have a wildcard version of itself.
+	if !isUnderFakeRoot {
+		// Generate builder constructor method for wildcard version of parent struct.
+		fieldData.Struct.TypeName += WildcardSuffix
+		if err := goPathChildConstructorTemplate.Execute(methodBuf, fieldData); err != nil {
+			return append(errors, err)
```
progressnerd: deliberately returning immediately rather than continuing to accumulate the errors?

ypathgen/pathgen.go
```diff
@@ -740,6 +766,74 @@ func generateChildConstructorsForLeafOrContainer(methodBuf *strings.Builder, fie
...
+				KeyParamType  string
+				KeyParamName  string
+			}{
+				MethodName:    BuilderKeyPrefix + keyParams[i].varName,
+				TypeName:      fieldData.TypeName,
+				KeySchemaName: keyParams[i].name,
+				KeyParamName:  keyParams[i].varName,
+				KeyParamType:  keyParams[i].typeName,
+			}); err != nil {
+			return append(errors, err)
```
progressnerd: deliberately returning immediately rather than accumulate the errors for all the keys?

ypathgen/pathgen.go
wenovus: Thanks, nice catch!

ypathgen/pathgen.go
wenovus: Thanks, nice catch!

PR #358, opened 2020-03-12
Add customData field to DeviceRoot object

wenovus:
This is to allow root-specific customized fields to be added to the root object via custom setters (as in the `ocpath_augment.go` example). This may be useful when resolving path structs to allow for some custom processing.

I made it private to force custom setters to be written in order to force its usages to be documentable via the setter method, instead of allowing any key to be set.


wenovus:
I experimented with embedding the PathStruct interface and also with a workaround that keeps ygot.NodePath being directly embedded. I feel it's more natural to embed ygot.NodePath directly because ModifyKey() is directly used, so I judged the workaround to be slightly better.

The workaround is just to change ModifyKey to a function instead of a method.

The fundamental issue here is Go only inherits publicly when ideally NodePath is privately inherited.

wenovus:
What's unnatural about embedding PathStruct interface is that a cast would be needed to access ModifyKey.

PR #325, opened 2019-11-07
add go.mod and fixing the ability to build with go 1.13

sysbot:
without go mod, building this will fail with

```
go run generator/generator.go
go: finding github.com/openconfig/ygot v0.6.0
go: finding github.com/openconfig/goyang latest
go: downloading github.com/openconfig/ygot v0.6.0
go: downloading github.com/openconfig/goyang v0.0.0-20190924211109-064f9690516f
go: extracting github.com/openconfig/goyang v0.0.0-20190924211109-064f9690516f
go: extracting github.com/openconfig/ygot v0.6.0
go: downloading github.com/kylelemons/godebug v1.1.0
go: extracting github.com/kylelemons/godebug v1.1.0
go: finding github.com/openconfig/gnmi latest
build command-line-arguments: cannot load github.com/openconfig/ygot/genutil: module github.com/openconfig/ygot@latest (v0.6.0) found, but does not contain package github.com/openconfig/ygot/genutil
```

when these go.mod and go.sum are added

```
go run generator/generator.go
F1106 16:56:13.641810   31670 generator.go:203] Error: no input modules specified
exit status 1
```

PR #359, opened 2020-03-18
Add initial mapping of proto messages to gNMI paths.

robshakir:
This is the start of a set of PRs to put together functionality to map from a ygot-generated protobuf to <path,value> pairs. This is to support efficient encoding on the wire for large datasets, and implementations that use protobuf as transport, whilst using ygot-generated Go internally.

```
 * (M) proto/{yext,ywrapper}
  - Regenerate protos using new versions of protoc-gen-go to take
    advantage of new protobuf/protoreflect APIs.
 * (M) ygot/*
  - Add initial functions to map from input proto to <path, value>
    pairs. Currently handles basic types.
  - Add example proto for tests.
```


ygot/testdata/exschemapath/exschemapath.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+
+package exschemapath;
+
+option go_package = "github.com/openconfig/ygot/ygot/testdata/exschemapath";
+
+import "github.com/openconfig/ygot/proto/ywrapper/ywrapper.proto";
+import "github.com/openconfig/ygot/proto/yext/yext.proto";
+
+message Root {
```
wenovus: Should there be a YANG file for this?

wenovus: That protogen generates from?

robshakir: There could be -- just hand creating this is probably better though since it will give the opportunity to add invalid proto fields (e.g., those that don't have the right annotations) in the future.

ygot/proto_test.go
```diff
@@ -0,0 +1,71 @@
...
+		},
+		wantPaths: map[string]interface{}{
+			"/interfaces/interface/config/description": "hello",
+		},
+	}, {
+		desc: "example message - supported fields",
+		inMsg: &epb.ExampleMessage{
+			Bo: &wpb.BoolValue{Value: true},
+			By: &wpb.BytesValue{Value: []byte{1, 2, 3, 4}},
+			// De is currently unsupported, needs parsing of decimal64 values.
```
wenovus: Why isn't `double` protobuf's parallel to float64? https://developers.google.com/protocol-buffers/docs/proto3#scalar

wenovus: or decimal64 for YANG?

ygot/testdata/exschemapath/exschemapath.proto
```diff
@@ -0,0 +1,39 @@
...
+    System system = 1; 
+    message InterfaceKey {
+        string name = 1 [(yext.schemapath) = "/interfaces/interface/config/name|/interfaces/interface/name"];
```
wenovus: This case will be covered later? Should we have a placeholder test like the other unsupported cases?

ygot/proto.go
```diff
@@ -0,0 +1,51 @@
...
+	"google.golang.org/protobuf/types/descriptorpb"
+
+	yextpb "github.com/openconfig/ygot/proto/yext"
+	wpb "github.com/openconfig/ygot/proto/ywrapper"
+)
+
+// pathsFromProto returns, from a populated proto, a map between the YANG schema
+// path (as specified in the yext.schemapath extension) and the value populated in
+// the message.
+func pathsFromProto(p proto.Message) (map[string]interface{}, error) {
```
wenovus: Is one of the purposes of this change to use the mapping from this function as an intermediate for converting from proto messages to GoStructs? By converting the map to a set of gNMI notifications and then calling `SetNode`?

wenovus: I guess one can also add proto encoding to `SetNode`

ygot/proto_test.go
robshakir: There are a lot of discussions about this one -- YANG decided to use decimal64, and since this format is intended to be able to serialise on the wire exactly, we decided to use something that was fixed precision. For developer ease-of-use in the Go bindings we decided to use a floating point value.

ygot/testdata/exschemapath/exschemapath.proto
robshakir: I've added support for this kind of message -- since it seemed trivial to do so.

ygot/proto.go
robshakir: That's right -- we have some proto input -- which has a schema path specified in it -- the wider change will go from schema tree path to data tree path, and then form the right call to `SetNode` to set the value in the parent.

This [doc](https://docs.google.com/document/d/106Qyw9Zre0kr7Gl3Xfkl4hYQv1Ml1GlTL3XsfcvOBKM/edit) described the high level approach before the new protobuf API, where a plugin to `protoc` was required (equivalent to `protoc-gen-go`), the new API makes this much cleaner.

scripts/install_go_protobuf.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+GIT_TAG=v1.20.0
```
wenovus: Wait actually, why doesn't the latest version of protobuf work? On my local machine with modules it works for me.

wenovus:
resolves #325

scripts/install_go_protobuf.sh
robshakir: Thanks -- it broke something for me in environment -- I think that this was unnecessary once I'd added the go mod support.

PR #366, opened 2020-04-01
Add a merge function that mutates inputs

dkarwowski:
The DeepCopy performed in MergeStructs can be expensive. This provides a
more efficient merging function which allows smaller inputs to be more
efficiently copied into a larger destination struct.

Tests are mostly shared between the two merge functions as they share
the same copying code.

### DIFF
--- a/ygot/struct_validation_map_test.go
+++ b/ygot/struct_validation_map_test.go
@@ -1623,20 +1623,15 @@ func TestMergeStructs(t *testing.T) {
 		inB:     &validatedMergeTestTwo{String: String("north-coast-old-rasputin")},
 		wantErr: "cannot merge structs that are not of matching types, *ygot.validatedMergeTest != *ygot.validatedMergeTestTwo",
 	}, {
-		name:    "error, bad data in A",
-		inA:     &validatedMergeTestTwo{I: "belleville-thames-surfer"},
-		inB:     &validatedMergeTestTwo{String: String("fourpure-beartooth")},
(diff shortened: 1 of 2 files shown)


ygot/struct_validation_map_test.go
```diff
@@ -1680,9 +1675,44 @@ func TestMergeStructs(t *testing.T) {
 		inB: &validatedMergeTestWithSlice{
 			SliceField: []*validatedMergeTestSliceField{{String("chinook-single-hop")}},
 		},
-		wantErr: "error merging b to new struct: source and destination lists must be unique",
+		wantErr: "source and destination lists must be unique",
 	}}
 
+	for _, tt := range tests {
```
robshakir, line 1681: I suggest that we make `tests` be defined outside of this function and then have a `TestMergeStructs` and `TestMergeStructsInto` -- just so we continue to have a 1:1 mapping between function-under-test and testcase.

robshakir:
@dkarwowski Thanks for submitting this here :-)

ygot/struct_validation_map_test.go
wenovus: I agree. I think somehow the same set of tests should be run for both functions, so that any new function added automatically applies to both.

dkarwowski: Sounds good -- split the functions up and pulled out a shared test cases slice for each to use.

PR #367, opened 2020-04-02
Fix defects in merging structs for enums and unions.

robshakir:
```
 * (M) ygot/struct_validation_map(_test)?.go
  - Resolve bug with merging enums where an unset value would overwrite
    a set value.
  - Resolve bug with merging enums where an error would not be thrown
    when the values were explicitly set to different values.
  - Resolve bug with merging unions where two unequal values would not
    result in an error being thrown.
  - Resolve bug with merging unions where an unset value would overwrite
    a set value.
```

### DIFF
--- a/ygot/struct_validation_map.go
+++ b/ygot/struct_validation_map.go
@@ -625,11 +625,11 @@ func copyInterfaceField(dstField, srcField reflect.Value) error {
 	s := srcField.Elem().Elem() // Dereference src to a struct.
 	if !util.IsNilOrInvalidValue(dstField) {
 		dV := dstField.Elem().Elem() // Dereference dst to a struct.
-		if !reflect.DeepEqual(s.Interface(), dV.Interface()){
+		if !reflect.DeepEqual(s.Interface(), dV.Interface()) {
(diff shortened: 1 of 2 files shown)


ygot/struct_validation_map.go
```diff
@@ -600,15 +614,22 @@ func copyPtrField(dstField, srcField reflect.Value) error {
...
 
 	if !util.IsValueInterface(srcField) || !util.IsValueStructPtr(srcField.Elem()) {
 		return fmt.Errorf("invalid interface type received: %T", srcField.Interface())
 	}
 
-	s := srcField.Elem().Elem() // Dereference to a struct.
+	s := srcField.Elem().Elem() // Dereference src to a struct.
+	if !util.IsNilOrInvalidValue(dstField) {
+		dV := dstField.Elem().Elem() // Dereference dst to a struct.
+		if !reflect.DeepEqual(s.Interface(), dV.Interface()) {
```
wenovus, line 628: What's the reason to not use `cmp` here?

robshakir: AFAIK, the only "recommended" move from `reflect.DeepEqual` to `cmp.Equal` is for proto messages, since `reflect.DeepEqual` doesn't work well for protos. In general, I've tried to use `reflect.DeepEqual` where we don't need the flexibility of `cmp`.

PR #368, opened 2020-04-03
Make AppendXXX Safe

robshakir:
When calling `AppendXXX` with a list member struct that didn't contain the keys we would panic because of an NPE.  This change adds clean handling for this error case.


ygen/gogen.go
```diff
@@ -764,20 +764,31 @@ func (t *{{ .Receiver }}) Delete{{ .ListName }}(
...
 		{{- if $key.IsScalarField -}}
-			key := *v.{{ $key.Name }}
+	if v.{{ $key.Name }} == nil {
+		return fmt.Errorf("invalid nil key received for {{ $key.Name }}")
+	}
+
+	key := *v.{{ $key.Name }}
 		{{- else -}}
 			key := v.{{ $key.Name }}
 		{{- end -}}
```
wenovus, line 794: nit: inconsistent spacing

robshakir: Fixed, thanks.

PR #370, opened 2020-04-11
Error out on nil input values when doing gNMI Encoding unmarshalling

wenovus:
Also add a corner case test for ygot.Diff.

NOTE: A nil TypeValue currently doesn't fail, and it looks to be implemented that way. I can change it if this doesn't make sense.

e.g.
```
func unmarshalLeafList(schema *yang.Entry, parent interface{}, value interface{}, enc Encoding) error {
	if util.IsValueNil(value) {
		return nil
	}
```

### DIFF
--- a/ygot/diff_test.go
+++ b/ygot/diff_test.go
@@ -827,6 +827,13 @@ func TestDiff(t *testing.T) {
 			}},
 		},
 	}, {
+		desc:   "extra empty child struct in modified -- no difference",
+		inOrig: &renderExample{},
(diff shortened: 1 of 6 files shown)


wenovus:
Note that I left JSON encoding to pass -- there were several places where a nil input for JSON encoding was expected to pass as no-ops. I don't know whether that should also now return an error, but it does make sense to me that for gNMI encoding, that we don't ever expect an update message with a nil TypedValue, so returning an error seems reasonable to me there.

PR #374, opened 2020-04-13
Make GoEnum a stringer

wenovus:
Note that UNSET will be treated as an invalid, out-of-range value.


ygot/types.go
```diff
@@ -76,6 +79,17 @@ type GoEnum interface {
 	// generated code file. The ygen library generates a static map of
 	// enumeration values that this method returns.
 	ΛMap() map[string]map[int64]EnumDefinition
+	// String provides the string representation of the enum, which will be
+	// the YANG name if it's in its defined range.
+	String() string
+}
+
+func EnumString(e GoEnum, val int64, enumTypeName string) string {
```
robshakir, line 87: Is this covered by any test? It's probably worth adding a simple test somewhere.

robshakir, line 87: I wonder whether we should call this something different? We already have [`ygot.EnumName`](https://godoc.org/github.com/openconfig/ygot/ygot#EnumName) which works slightly differently (it can return an `error` in the case of a number of invalid inputs), so it'd be good to keep to clear as to why users would use this function.

ygot/types.go
```diff
@@ -76,6 +79,17 @@ type GoEnum interface {
 	// generated code file. The ygen library generates a static map of
 	// enumeration values that this method returns.
 	ΛMap() map[string]map[int64]EnumDefinition
+	// String provides the string representation of the enum, which will be
+	// the YANG name if it's in its defined range.
+	String() string
+}
+
```
robshakir, line 86: Missing a docstring for this public API function.

ygen/gogen.go
```diff
@@ -546,6 +546,10 @@ func (E_{{ .EnumerationPrefix }}) IsYANGGoEnum() {}
 // ΛMap returns the value lookup map associated with  {{ .EnumerationPrefix }}.
 func (E_{{ .EnumerationPrefix }}) ΛMap() map[string]map[int64]ygot.EnumDefinition { return ΛEnum; }
 
+func (e E_{{ .EnumerationPrefix }}) String() string {
```
robshakir, line 549: Can we add a comment to this function to ensure that it's understood that this function will return an error string embedded in the output - and therefore the output might not be the enumeration's actual string value? We really need users *not* to call this when doing rendering of this value to a string.

ygot/types.go
wenovus: I renamed it EnumLogString, as the request was that it's hard to read in a log.

ygen/gogen.go
wenovus: Added:
// String returns a logging-friendly string for E_{{ .EnumerationPrefix }}.

PR #375, opened 2020-04-15
Fix enum key string bugs during ytypes.Validate and JSON unmarshalling.

wenovus:
- Changed json.Unmarshal to fmt.Sprint for consistent enum-key
presentation during schema validation (ytypes.Validate).
- Force an error during JSON marshalling if an unset or otherwise
out-of-range enum value is encountered as an enum key of a list.

The first bug was a latent issue exposed by the recent addition of a
String() method for GoEnum.


util/reflect.go
```diff
@@ -1029,6 +1029,10 @@ func getNodesList(schema *yang.Entry, root interface{}, path *gpb.Path) ([]inter
 				if err != nil {
 					return nil, nil, err
 				}
+				// NOTE: Normally we'd like to use ygot.KeyValueAsString for conversion
```
robshakir, line 1032: Is this true now we implemented the log string change? If there are two invalid values - do they end up being the same string representation?

util/reflect.go
```diff
@@ -1053,6 +1057,10 @@ func getNodesList(schema *yang.Entry, root interface{}, path *gpb.Path) ([]inter
 						// If the key is not filled, it is assumed to match.
 						continue
 					}
+					// NOTE: Normally we'd like to use ygot.KeyValueAsString for conversion
```
robshakir, line 1060: I don't think we have to repeat the whole comment here, perhaps just say "as above, we don't require this to be a YANG enum string representation".

util/reflect.go
robshakir: From below, I see that the invalid value is included in the error message, so I think this is fine.

ytypes/leafref.go
```diff
@@ -143,23 +142,27 @@ func leafRefToGNMIPath(root *util.NodeInfo, path string, pathQueryNode *util.Pat
...
 			case 1:
-				actualValue := ns[0]
 				if err != nil {
 					return nil, err
 				}
-				j, err = json.Marshal(actualValue)
-				if err != nil {
-					return nil, err
+				rval := reflect.ValueOf(ns[0])
+				if util.IsValuePtr(rval) {
```
robshakir, line 153: Is this handling the case where there is a pointer key for an element in a list? I can't remember anywhere that ygen would generate code of that nature - aren't all list keys scalars?

wenovus: dataNodesAtPath returns the data node and not the key, so pointers are returned for scalars

robshakir: So if we ask for `/interfaces/interface[name=Ethernet42]` it returns us `config/name` leaf - rather than the map key itself? Makes sense if so.

wenovus: That would return the `oc.Interface` value. In this function, `dataNodesAtPath` is used to get the key value from a path like `current()/../../key`, and since that simply refers to a leaf, it returns the field value, which is a pointer in the case of a scalar. Am I answering the right question?

robshakir: Ah, makes sense.

PR #376, opened 2020-04-17
Add enumgen.go and move around some types, functions, and tests.

wenovus:
There are **no** functional changes. These only group types, functions,
and tests in more appropriate places. A few comments are updated.


ygen/enumgen.go
```diff
@@ -0,0 +1,461 @@
+package ygen
```
robshakir, line 1: Can you add the copyright and licence notice to the top of this file please? (It should be at the top of all code files across the repo.)

PR #377, opened 2020-04-17
Use new protobuf library for testcmp and testutils and update gnmi & goyang libraries

wenovus:


testcmp/cmp.go
```diff
@@ -43,13 +45,18 @@ func UpdateComparer(schemaFunc func() (*ytypes.Schema, error)) (testutil.CustomC
...
-			}
-			return equal
-		}),
+		reflect.TypeOf(&gnmipb.Update{}): protocmp.FilterMessage(&gnmipb.Update{},
+			cmp.Comparer(func(ma, mb protocmp.Message) bool {
+				a := new(gnmipb.Update)
+				proto.Merge(a, ma)
+				b := new(gnmipb.Update)
+				proto.Merge(b, mb)
+				_, equal, err := GNMIUpdateComparer(a, b, schema)
```
robshakir, line 54: Could you explain why there is a need to clone here? Previously, `GNMIUpdateComparer` didn't make this clone AFAICS - and I don't think that there is anything in the comparer that can mutate this, right?

wenovus: Where is the Clone you're referring to?

In your highlighted code, proto.Merge is done to convert the transformed protos back into their original form so that `GNMIUpdateComparer` can continue operating as before.

robshakir: I meant the `new()` and `proto.Merge` = (create new and clone).  If this is needed to get back to the original form, SGTM.

wenovus: Yep, joe handwrote this for me -- `protocmp` transforms the protos into [protocmp.Message](https://godoc.org/google.golang.org/protobuf/testing/protocmp#Message) before comparing, and since `GNMIUpdateComparer` operates on regular protos, `proto.Merge` is a necessary step to convert them back into the correct regular proto.

PR #378, opened 2020-04-28
Add Go GitHub action to run CI.

robshakir:
Switch ygot's CI to run in GitHub actions.


robshakir:
This has been a labour of love. @wenovus PTAL -- I'll switch the required checks to these new ones rather than Travis after this.

.github/workflows/go.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: Get dependencies
+        run: |
+          go get -v -t -d ./...
+
+      - name: Generate dependencies
+        run: |
+          cd go/src/github.com/openconfig/ygot
+          make generate
+          go build -v ./...
```
wenovus, line 54: was this line supposed to remain here or let "Build packages" handle it?

robshakir: It seems like we need to force a build at this point to ensure that there are some dependencies that are retained across steps. I'm not quite sure why this is right now -- I can have a bit of a play around with it and see whether it was something else.

README.md
```diff
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/openconfig/ygot.svg?branch=master)](https://travis-ci.org/openconfig/ygot)
+![Go](https://github.com/openconfig/ygot/workflows/Go/badge.svg)
```
wenovus, line 1: append ?branch=master to link

.github/workflows/go.yml
robshakir: This actually looks to have been a bad path -- fixed! Thanks for pointing it out again.

integration_tests/uncompressed/cp.go
```diff
@@ -16,4 +16,4 @@
 // schemas using real YANG schemas.
 package uncompressed
 
-//go:generate sh -c "go run ../../generator/generator.go -path=yang -output_file=cschema/structs.go -package_name=cschema -generate_fakeroot -fakeroot_name=root -generate_getters -compress_paths yang/uncompressed.yang && go run ../../generator/generator.go -path=yang -output_file=uschema/structs.go -package_name=uschema -generate_fakeroot -fakeroot_name=root -generate_getters yang/uncompressed.yang"
+//go:generate sh -c "go run ${SRCDIR}/generator/generator.go -path=yang -output_file=cschema/structs.go -package_name=cschema -generate_fakeroot -fakeroot_name=root -generate_getters -compress_paths yang/uncompressed.yang && go run ../../generator/generator.go -path=yang -output_file=uschema/structs.go -package_name=uschema -generate_fakeroot -fakeroot_name=root -generate_getters yang/uncompressed.yang"
```
wenovus, line 19: Is this one missing the ./genproto.sh?

demo/protobuf_getting_started/update.sh
```diff
@@ -5,14 +5,18 @@ clean() {
   rm -rf deps
 }
 
+if [ -z ${SRCDIR} ]; then
```
wenovus, line 8: I'm curious if using `dirname "$0"/../..` would be slightly more robust in case there are multiple directories within a GOPATH: https://stackoverflow.com/questions/36017724/can-i-have-multiple-gopath-directories

wenovus: I mean for the default case.

integration_tests/uncompressed/cp.go
robshakir: No, there's no need for the `genproto.sh` trick here, since this is just doing Go generation. This is more like the demo dirs.

demo/protobuf_getting_started/update.sh
robshakir: Good question. I wasn't even aware that you can have >1 path. Let me send a PR to fix this.

PR #381, opened 2020-04-29
Support multiple GOPATHs

robshakir:
Support multiple GOPATHs being set in the calling environment by not using this variable in scripts called by `go generate`.


robshakir:
Thanks for catching that -- I didn't add tests for non-`SRCDIR` generation - because they'll failin different environments (e.g., CI) - but this is now fixed.

PR #383, opened 2020-05-15
Create Annotation JSON Tree using deepcopy instead of Marshal+Unmarshal

wenovus:
Currently, the jsonTree (`map[string]interface{}`) construction is done incorrectly due to a `json.Unmarshal` call on the marshalled data from the custom `Marshaler`, which doesn't make sense if the two calls are not compatible.

I believe the reason Marshal+Unmarshal is done is to copy the data, so instead I've used an MIT licensed library to do a deepcopy so that the final jsonTree contains the `Annotation` struct directly, meaning that a subsequent `json.Marshal` will call the customized `MarshalJSON` function.

Although I don't see how using the `Annotation` struct directly instead of a `map[string]interface{}` would create any new problems, it still doesn't solve the problem of how to unmarshal from bytes to the jsonTree correctly, but I saw this so I'm assuming this is irrelevant for now: `// TODO(robjs): Implement unmarshalling annotations.`

I don't see how to do `bytes -> jsonTree` easily, maybe somehow we'll have to bypass the `Marshaler` interface if we ever do intend to unmarshal `Annotation`s.

### DIFF
--- a/ygot/render.go
+++ b/ygot/render.go
@@ -1355,6 +1355,9 @@ func jsonAnnotationSlice(v reflect.Value) (interface{}, error) {
 
 	vals := []interface{}{}
 	for i := 0; i < v.Len(); i++ {
+		// deepcopy the Annotation to avoid accidentally modifying its contents.
+		// Since Annotations must have an implemented MarshalJSON
+		// function, this should not impact the final JSON.
 		vals = append(vals, deepcopy.Copy(v.Index(i).Interface().(Annotation)))
 	}
 	return vals, nil
--- a/ygot/render.go
+++ b/ygot/render.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"github.com/golang/protobuf/proto"
+	"github.com/mohae/deepcopy"
 	"github.com/openconfig/gnmi/errlist"
 	"github.com/openconfig/gnmi/value"
 	"github.com/openconfig/ygot/util"
@@ -1354,20 +1355,7 @@ func jsonAnnotationSlice(v reflect.Value) (interface{}, error) {
 
 	vals := []interface{}{}
 	for i := 0; i < v.Len(); i++ {
-		fv := v.Index(i).Interface().(Annotation)
-		jv, err := fv.MarshalJSON()
-		if err != nil {
-			return nil, fmt.Errorf("cannot marshal annotation %v type %T to JSON: %v", fv, fv, err)
-		}
-
-		// MarshalJSON returns []byte, but we really want to have this as the unmarshalled
-		// value, since constructJSON returns a series of map[string]interface{} Which
-		// are later marshalled, we therefore unmarshal the []byte into an interface{}
-		var nv interface{}
-		if err := json.Unmarshal(jv, &nv); err != nil {
-			return nil, fmt.Errorf("annotation %v, type %T could not be unmarshalled from JSON: %v", fv, fv, err)
-		}
-		vals = append(vals, nv)
+		vals = append(vals, deepcopy.Copy(v.Index(i).Interface().(Annotation)))
 	}
 	return vals, nil
 }
--- a/ygot/render_test.go
+++ b/ygot/render_test.go
@@ -1579,12 +1579,18 @@ type testAnnotation struct {
 	AnnotationFieldOne string `json:"field"`
 }
 
+// MarshalJSON repeats the string in the JSON representation. This deviation
+// from json.Marshal helps test that the json generation uses the
+// Annotation's custom MarshalJSON/UnmarshalJSON functions.
 func (t *testAnnotation) MarshalJSON() ([]byte, error) {
 	t2 := *t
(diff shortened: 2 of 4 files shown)


ygot/render.go
```diff
@@ -1354,20 +1355,7 @@ func jsonAnnotationSlice(v reflect.Value) (interface{}, error) {
...
-
-		// MarshalJSON returns []byte, but we really want to have this as the unmarshalled
-		// value, since constructJSON returns a series of map[string]interface{} Which
-		// are later marshalled, we therefore unmarshal the []byte into an interface{}
-		var nv interface{}
-		if err := json.Unmarshal(jv, &nv); err != nil {
-			return nil, fmt.Errorf("annotation %v, type %T could not be unmarshalled from JSON: %v", fv, fv, err)
-		}
-		vals = append(vals, nv)
+		vals = append(vals, deepcopy.Copy(v.Index(i).Interface().(Annotation)))
```
robshakir, line 1358: Can you add a comment here? I assume that because we have defined that the `Annotation` must be JSON marshallable itself then we don't need to marshal it in this case?

ygot/render_test.go
```diff
@@ -1580,11 +1580,17 @@ type testAnnotation struct {
 }
 
 func (t *testAnnotation) MarshalJSON() ([]byte, error) {
-	return json.Marshal(*t)
+	t2 := *t
+	t2.AnnotationFieldOne += t2.AnnotationFieldOne
```
robshakir, line 1584: Can you explain why this is required? It seems to be setting `t2` to itself, do you mean to be setting it to a field of `t`?

robshakir: Is there a reason that there are no test changes associated here? Can we add a test case that covers the bug that reported this at least?

ygot/render_test.go
```diff
@@ -1580,11 +1580,17 @@ type testAnnotation struct {
...
+	return json.Marshal(t2)
 }
 
 func (t *testAnnotation) UnmarshalJSON(d []byte) error {
-	return json.Unmarshal(d, t)
+	if err := json.Unmarshal(d, t); err != nil {
+		return err
+	}
+	t.AnnotationFieldOne = t.AnnotationFieldOne[:len(t.AnnotationFieldOne)/2]
```
robshakir, line 1592: Again, can you explain what is happening here please?

ygot/render_test.go
```diff
@@ -2240,6 +2246,9 @@ func TestConstructJSON(t *testing.T) {
 		gotietf, err := ConstructIETFJSON(tt.in, &RFC7951JSONConfig{
 			AppendModuleName: tt.inAppendMod,
 		})
+		if err == nil {
+			_, err = json.Marshal(gotietf)
```
robshakir, line 2250: Can we make this and the one below a separate discussion? i.e., today, if we get the `err != nil` case failing, it's not clear whether it was `ConstructIETFJSON` or `json.Marshal` that threw an error. It'd be good to handle these separately.

ygot/render.go
wenovus: Yes that's right, done.

ygot/render_test.go
wenovus: Comment added -- this change exposes the bug by deviating the unmarshal/marshal behaviour from `json.Marshal/json.Unmarshal`. This deviation causes the previous code to fail due to its unbalanced usage of using the custom `MarshalJSON` followed by the built-in `json.Unmarshal`.

ygot/render_test.go
wenovus: As above.

ygot/render_test.go
wenovus: Done, also restructured the test a bit by using `t.Run()`.

ygot/render_test.go
robshakir: Sorry -- this still doesn't make sense to me. What is the "deviation" that you're referring to here? 

Please could we have a more explicit comment of the form "we change the value of t2 explicitly to ensure that <something doesn't happen>". I think I'm missing context as to why we need the behaviour of this `MarshalJSON` to be different to the behaviour of the builtin `json.Marshal`.

Thanks -- it's very possible I'm just missing something obvious here.

ygot/render_test.go
robshakir: Thanks, this LGTM.

robshakir:
Looks like there's some outstanding comments here @wenovus -- did you see my latest?

ygot/render_test.go
wenovus: Ok updated, the point of this is to prevent the old way of copying -- i.e. `json.Marshal`, then `json.Unmarshal` from succeeding. The previous way was only using the `Annotation` interface for the marshal call, so if the marshal call deviated, `json.Unmarshal` would fail.

robshakir: Thanks -- this is clearer now.

PR #384, opened 2020-05-16
Split each enum class's unique key calculation into separate functions

wenovus:
- Split out `identityBaseKey`, `enumeratedTypedefKey`, and `enumeratedLeafKey` functions from their respective name generation functions.
- Reorganize order of functions within enumgen.go to make the next PR less crazy.

### DIFF
--- a/ygen/enumgen.go
+++ b/ygen/enumgen.go
@@ -74,61 +74,186 @@ func newEnumGenState() *enumGenState {
 	}
 }
 
-// enumeratedUnionEntry takes an input YANG union yang.Entry and returns the set of enumerated
-// values that should be generated for the entry. New yang.Entry instances are synthesised within
-// the yangEnums returned such that enumerations can be generated directly from the output of
-// this function in common with enumerations that are not within a union. The name of the enumerated
-// value is calculated based on the original context, whether path compression is enabled based
-// on the compressPaths boolean, and whether the name should not include underscores, as per the
-// noUnderscores boolean.
-func (s *enumGenState) enumeratedUnionEntry(e *yang.Entry, compressPaths, noUnderscores, skipEnumDedup bool) ([]*yangEnum, error) {
-	var es []*yangEnum
-
-	for _, t := range util.EnumeratedUnionTypes(e.Type.Type) {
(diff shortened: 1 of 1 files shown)


ygen/enumgen.go
```diff
@@ -74,61 +74,186 @@ func newEnumGenState() *enumGenState {
...
+	// a name for each of them, but rather use a common name, we use the non-CamelCase lookup
+	// as this is unique, whereas post-camelisation, we may have name clashes. Since a typedef
+	// does not have a 'path' in Goyang, so we synthesise one using the form
+	// module-name/typedef-name.
+	return fmt.Sprintf("%s/%s", definingModName, typeName), definingModName, typeName, nil
+}
+
+// enumLeafKey calculates a unique string key for the input enum leaf.  If
+// compressPaths is true, it also returns the compress name of the entry for
+// use in name generation, if needed.
```
robshakir, line 166: Shall we highlight here that this covers leaves of type `enumeration` only? I believe that this is the case based on the name, but it might be useful to be explicit.

ygen/enumgen.go
```diff
@@ -74,61 +74,186 @@ func newEnumGenState() *enumGenState {
...
+			// compressed enum names when they describe different
+			// enums.
+			uniqueIdentifier += compressName
+		}
 	}
+	return uniqueIdentifier, compressName
+}
 
-	return es, nil
+// enumIdentifier takes in an enum entry and returns a unique identifier for
```
robshakir, line 211: Do we expect to call this in multiple places going forward? It seems like it's just called in a single place. Should we also be re-using the same function where there is an `enumeration` within a `union`?

wenovus: Indeed it's already being used by `enumeration` within a `union`, just that it's done indirectly via `enumName = s.resolveEnumName(e, compressPaths, noUnderscores, skipEnumDedup)` within `enumeratedUnionEntry()`.

I'll add a comment to `enumIdentifier` saying it can be used for union containing an enum as well.

robshakir: Thanks, SGTM.

PR #386, opened 2020-05-19
Add staticcheck CI Step

wenovus:
Most of the checks have looked useful so far.
I've added two ignore rules at the top of two test files, both due to generated code.


wenovus:
There are a couple types of errors I don't know how to handle. @robshakir could you take a look?

.github/workflows/go.yml
```diff
@@ -92,6 +97,15 @@ jobs:
           done
         working-directory: go/src/github.com/openconfig/ygot
 
+      - name: Staticcheck
```
robshakir, line 100: I wonder whether we should use this action going forward? https://github.com/grandcolline/golang-github-actions

It looks like there are a whole host of things that we'd get there which might give us additional coverage w.r.t static analysis? I'm happy to change to this after we merge this change.

ytypes/list_test.go
```diff
@@ -324,16 +324,10 @@ func TestValidateListStructKey(t *testing.T) {
 		Key2     *int32  `path:"key2"`
 		LeafName *string `path:"leaf-name"`
 	}
-	type BadElemStruct1 struct {
+	type BadElemStruct struct {
 		Key1     *string `path:"key1"`
 		LeafName *string `path:"leaf-name"`
 	}
-	type BadElemStruct2 struct {
```
robshakir, line 331: Can you explain this change?

ytypes/node_test.go
```diff
@@ -808,9 +809,9 @@ func treeNodesEqual(got, want []*TreeNode) error {
 		if !match {
 			paths := []string{}
 			for _, g := range got {
-				paths = append(paths, fmt.Sprintf("< %s | %#v >", proto.CompactTextString(g.Path), g))
```
robshakir, line 811: Did `proto.CompactTextString` get deprecated?

ytypes/list_test.go
wenovus: Unused struct

ytypes/node_test.go
wenovus: I changed to the new proto library the checker recommended. I couldn't find this function in https://godoc.org/google.golang.org/protobuf/encoding/prototext, so I used this.

.github/workflows/go.yml
wenovus: I'm not sure because our actions run on only certain directories, but the implementation there runs on `./...`: https://github.com/grandcolline/golang-github-actions/blob/master/entrypoint.sh#L201

So maybe it's not flexible enough?

ytypes/list_test.go
robshakir: Ah, great.

.github/workflows/go.yml
robshakir: Ah, that makes sense. SGTM.

ytypes/node_test.go
robshakir: Sure - I'm happy with this. Looks like this might be part of the protobuf refactoring that's been going on.

wenovus:
Thanks for the reviews and suggestions!

PR #387, opened 2020-05-20
Change enum name conflict resolution from on-the-fly to all-at-once

wenovus:
#385 converted name generation to all-at-once, but conflict resolution was still on-the-fly.

This change converts conflict resolution also to all-at-once by altering `findEnumSet` to doing two passes over the input enumeration entries (name generation and yangEnum creation), instead of the current single pass that does both at once. The two passes are interceded by name conflict resolution.

It also makes any `identity` or enumerated typedef name conflict an error. Only `enumeration` leaves are allowed to conflict.

NOTE: This is a split change that is never meant to be merged by itself, the next change will introduce the new `enumeration` deduplication rules that completes the overall task of changing the conflict resolution rules of enumerated values.


robshakir:
Can you clarify what's meant by a temporary change please? Do we expect to roll some of this change back, or is this just ensuring that we keep exactly the same behaviour whilst refactoring?

Thanks,
r.

wenovus:
Clarified, it just means it's part of a bigger change.

ygen/enumgen.go
```diff
@@ -363,6 +422,61 @@ func findEnumSet(entries map[string]*yang.Entry, compressPaths, noUnderscores, s
...
+			if err := s.resolveIdentityRefBaseType(e, noUnderscores); err != nil {
+				errs = append(errs, err)
+			}
+		case e.Type.Name == "enumeration":
+			// We simply want to map this enumeration into a new name. Since we do
+			// de-duplication of re-used enumerated leaves at different points in
+			// the schema (e.g., if openconfig-bgp/container/enum-A can be instantiated
+			// in two places, then we do not want to have multiple enumerated types
+			// that represent this leaf), then we do not have errors if duplicates
+			// occur, we simply perform de-duplication at this stage.
```
robshakir, line 459: Is this comment now stale -- since we're explicitly not de-duplicating at this stage right?

ygen/enumgen.go
```diff
@@ -363,6 +422,61 @@ func findEnumSet(entries map[string]*yang.Entry, compressPaths, noUnderscores, s
 	// where there are erroneous config/state differences).
 	sort.Strings(enumPaths)
 
+	s := newEnumGenState()
+
+	// This is the first of two passes over the input enum entries. The
+	// purpose of this pass is to resolve the names of all the enumerated
```
robshakir, line 428: `The purpose of this pass is to establish what the default name of each of the enumerated types that are to be output in the code is. At this stage, we just try and calculate the complete set, rather than resolving any clashes in names that might occur.`

ygen/enumgen.go
```diff
@@ -71,6 +71,67 @@ func newEnumSet() *enumSet {
 	}
 }
```
robshakir, line 73: Throughout this code I wonder whether "duplicate" is the right wording. Particularly because I think there are two cases that I can see:

```yang

grouping A {
  leaf b { type enumeration { enum C; } }
}

container one {
  uses A;
}

container two {
  uses B;
}
```

in this case, leaf `one/b` and `two/b` are output once in our current approach (without `skipDedup` enabled). However, I think your duplicate sets issue is more dealing with the case where we have two values that come up with the same `E_Foo_Bar` type, right?

Can we call these `NameClash` sets or something to be clearer?

ygen/enumgen.go
```diff
@@ -440,103 +562,132 @@ func findEnumSet(entries map[string]*yang.Entry, compressPaths, noUnderscores, s
 type enumGenState struct {
 	// definedEnums keeps track of generated enum names to avoid conflicts.
 	definedEnums map[string]bool
-	enumSet      *enumSet
+	// enumSet contains the final, deduped generated enumerated value names
```
robshakir, line 565: Do you mean uniquely named /and/ deduplicated here?

ygen/enumgen.go
```diff
@@ -440,103 +562,132 @@ func findEnumSet(entries map[string]*yang.Entry, compressPaths, noUnderscores, s
...
+	// to be returned by findEnumSet.
+	enumSet *enumSet
+	// enumeratedTypedefDedupSets stores the set of unique string keys
```
robshakir, line 568: Here's where I think that the "name clash" terminology makes more sense -- since 'deduplication' has typically mean that two leaves with *different* paths share the same underlying type.

ygen/enumgen.go
```diff
@@ -440,103 +562,132 @@ func findEnumSet(entries map[string]*yang.Entry, compressPaths, noUnderscores, s
...
+	// This is an intermediate storage for these names. While the initial
+	// generated name is ideally stored inside enumSet directly,
+	// enumeration names may collide, and thus need de-duplication before
+	// they're stored inside enumSet. This is the reason for being for all
+	// dedup sets within enumGenState.
+	enumeratedTypedefDedupSets map[string]map[string]bool
+	// enumeratedLeafDedupSets stores the unique string keys representing
+	// enumeration leaves that were mapped to the same initial generated
+	// name using the basic name generation rules.
+	enumeratedLeafDedupSets map[string]map[string]bool
```
robshakir, line 580: Can there ever be cases where a key in `enumeratedTypedefDedupSets` is the same as a key in `enumeratedLeafDedupSets`?

ygen/enumgen.go
```diff
@@ -440,103 +562,132 @@ func findEnumSet(entries map[string]*yang.Entry, compressPaths, noUnderscores, s
...
+// de-duplication between different enum keys that mapped to the same initial
+// generated name. Then it stores these final names in their appropriate name
+// maps within enumSet.
+func (s *enumGenState) resolveDedupSets() {
+	s.resolveDedupSet(s.enumeratedTypedefDedupSets, s.enumSet.uniqueEnumeratedTypedefNames)
+	s.resolveDedupSet(s.enumeratedLeafDedupSets, s.enumSet.uniqueEnumeratedLeafNames)
+}
+
+// resolveDedupSet carries out de-duplication on the input dedup set to
+// generate the final names, and stores those names in the given unique map.
```
robshakir, line 604: Here would be a good place to explain what the logic that we use to make these names unique is. I'd suggest a brief synopsis in the doc comment of this function.

ygen/enumgen.go
```diff
@@ -440,103 +562,132 @@ func findEnumSet(entries map[string]*yang.Entry, compressPaths, noUnderscores, s
...
-// enumeratedUnionEntry takes an input YANG union yang.Entry and returns the set of enumerated
-// values that should be generated for the entry. New yang.Entry instances are synthesised within
-// the yangEnums returned such that enumerations can be generated directly from the output of
-// this function in common with enumerations that are not within a union. The name of the enumerated
+// resolveDedupSets scans through all of the dedup sets, and carries out
+// de-duplication between different enum keys that mapped to the same initial
+// generated name. Then it stores these final names in their appropriate name
+// maps within enumSet.
+func (s *enumGenState) resolveDedupSets() {
+	s.resolveDedupSet(s.enumeratedTypedefDedupSets, s.enumSet.uniqueEnumeratedTypedefNames)
```
robshakir, line 599: Do these functions need to in-place modify the argument -- is there a reason not to just assign their output to the right variable in `s`? i.e., `s.enumSet.uniqueEnumeratedTypedefNames = s.resolveDedupSet(s.enumeratedTypedefDedupSets)`

ygen/enumgen_test.go
```diff
@@ -20,9 +20,99 @@ import (
...
+		},
+		inDedupSets: map[string]map[string]bool{
+			"Foo": map[string]bool{
+				"enum-a": true,
+				"enum-b": true,
+			},
+		},
+		wantUniqueNamesMap: map[string]string{
+			"enum-a": "Foo",
+			"enum-b": "Foo_",
```
robshakir, line 63: This `_` is essentially to keep this change a no-op for existing generated code right?

ygen/enumgen_test.go
```diff
@@ -20,9 +20,99 @@ import (
...
+			"Baz": true,
+		},
+		inDedupSets: map[string]map[string]bool{
+			// De-dup is done in lexicographical order.
+			"Bar": map[string]bool{
+				"enum-a": true,
+			},
+			"Foo": map[string]bool{
+				// This entry is ignored due to a name being already assigned.
+				"enum-a": true,
```
robshakir, line 92: Isn't this an error? `enum-a` can't be mapped to two different types, right?

ygen/enumgen_test.go
```diff
@@ -82,6 +172,64 @@ func TestFindEnumSet(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			"/container/config/identityref-leaf2": {
+				Name: "identityref-leaf",
+				Type: &yang.YangType{
+					Name: "identityref",
+					IdentityBase: &yang.Identity{
+						Name: "baseIdentity",
```
robshakir, line 207: Please add a comment here saying that `base-identity` and `baseIdentity` are camel-cased to the same thing -- just so that it's evident to the reader why this is a conflict.

ygen/enumgen_test.go
```diff
@@ -169,6 +317,240 @@ func TestFindEnumSet(t *testing.T) {
...
+				},
+				Parent: &yang.Entry{
+					Name: "state",
+					Parent: &yang.Entry{
+						Name:   "container",
+						Parent: &yang.Entry{Name: "base-module"},
+					},
+				},
+			},
+			"/container/state/enumerationLeaf": {
```
robshakir, line 349: Ditto here re: camelcasing being the same.

ygen/enumgen_test.go
wenovus: Yes, this test's logic is going change.

ygen/enumgen.go
wenovus: No, but the way their name clashes are resolved are different, so the different maps serve to separate their processing.

ygen/enumgen.go
wenovus: I will add it in the next PR when the new uniquification rules are introduced.

ygen/enumgen_test.go
wenovus: Added `uniqueEnumeratedTypedefEntries map[string]bool{}` and `uniqueEnumeratedLeafEntries map[string]bool{}` to ensure each unique enum can only be mapped to a single default name.

ygen/enumgen_test.go
wenovus: I won't do this the next PR will render this test obsolete.

ygen/enumgen_test.go
wenovus: same: I won't do this the next PR will render this test obsolete.

ygen/enumgen_test.go
```diff
@@ -36,7 +105,7 @@ func TestFindEnumSet(t *testing.T) {
 		wantCompressed          map[string]*yangEnum
 		wantUncompressed        map[string]*yangEnum
 		wantSame                bool // Whether to expect same compressed/uncompressed output
-		wantErr                 bool
+		wantErrSubstr           string
```
robshakir, line 108: (no action required) I found [this article](https://blog.golang.org/go1.13-errors) on new error approaches in Go 1.13 pretty interesting - we might want to think about whether we want to experiment with these differences going forward. I don't think we need to right now at all - but I figured it might also be of interest.

wenovus: Interesting, maybe `ytypes.Validate` and `ytypes.Unmarshal`'s errors should be typed to allow for wrapping.

PR #391, opened 2020-05-21
Refactor CI workflows

robshakir:
This is a very messy PR to move checks for coverage into a separate, unrelated, check in GitHub actions to make them more robust to changes needed for code generation.


robshakir:
Yeah -- I really wish that there was some way to be able to locally debug this. Maybe a docker instance that you could spin up locally that used the same environment.

robshakir:
Thanks for the review.

PR #392, opened 2020-05-28
Introduce new clash-resolution algorithm for Enumeration Leaves

wenovus:
`resolveNameClashSet` now has the following mechanism:
- If there is no clash in the enumeration leaf default name, then use the default name.
- Otherwise, use the following algorithm as discussed and pasting here:
```
let a duplicate set be the set of enums that would have the same output name
for each duplicate set:
   let the original name be the non-unique name
  
   prepend the module name
   if prepend makes the set unique:
     use these names and return
   else:
      for each ancestor of each member of the duplicate set:
        prepend the ancestor name to the original name
        if prepend makes the set unique:
         use these names and return
```


ygen/enumgen.go
```diff
@@ -587,49 +590,140 @@ func newEnumGenState() *enumGenState {
...
+					// Try prepending the ancestor name to disambiguate.
+					candidateName = yang.CamelCase(entry.Parent.Name) + delimiter + clashName
+					newNameClashSet[enumKey] = entry.Parent
+				}
+				candidateUniqueNames[candidateName] = enumKey
+			}
+			// Update the clash set entry pointers to their next level of ancestors.
+			nameClashSet = newNameClashSet
+
+			if i == 0 {
```
robshakir, line 709: This should only apply when `compressPaths` is enabled right? i.e., we should not skip the parent directly if we're not doing path compression.

ygen/enumgen_test.go
```diff
@@ -26,69 +26,748 @@ import (
...
+							Name: "gran-gran-b",
+							Parent: &yang.Entry{
+								Name: "base-module",
+							},
+						},
+					},
+				},
+			},
+		},
+		wantUniqueNamesMap: map[string]string{
```
robshakir, line 244: This is where I think we need a difference in implementation. I'd expect that this is `ParentA_Foo` and `ParentB_Foo` when compressing paths is disabled. 

On this note, did we check what the diff created running this for the uncompressed example is? It should really be nothing.

ygen/enumgen.go
wenovus: Thanks for catching this! The uncompressed handling was missing. I've added it now, and it *should not* do any clash resolution, since uncompressed enumeration leaves are named according to their entire path.

ygen/enumgen_test.go
wenovus: I've added test coverage for uncompressed schemas, where if any conflict occurs, generation simply fails due to enumeration leaves being named according to their entire path.

wenovus: I just checked, `uexampleoc` has no changes between master branch and the last PR #393

ygen/enumgen.go
robshakir: Great, thanks -- this handling LGTM.

ygen/enumgen_test.go
robshakir: Great, thanks. It looks like my comment above is outdated because `gran-gran-a` is the node that's used as the prefix here.

PR #399, opened 2020-06-03
Use new goyang dependency that fixes decimal64 typedef range parse bug

### LINKED ISSUE #397: decimal64 typedef with range failing to parse
I upgraded to the most recent master branch and am encountering a new generator error parsing a vendor YANG definition.  This error did not occur in the previous version of ygot I was using.

Here is an example of the error I am seeing:

```
F0603 11:28:56.655930    8941 generator.go:272] ERROR Generating Code: yang/G30_FP4_1_1/amplifier@2019-08-23.yang:146:11: bad range: strconv.ParseUint: parsing "12.0": invalid syntax
```

Here is the YANG snippet for the typedef:

```
  typedef "optical-dB" {
    type "decimal64" {
      fraction-digits "1";
    }

    units "dB";
    description
      "Optical Attribute value with Unit dB";
  }
```

Here is the leaf definition using a range qualification:

```
      leaf "gain-adjustment" {
        type "types:optical-dB" {
          range "-12.0..12.0";
        }

        default "0.0";
        description
          "The value is used for adjustment of gain when the amplifier in automatic control mode,
  the automatically calculated gain will include offset of this attribute.";
      }
```

wenovus:
Fixes #397

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/google/go-cmp v0.4.1
 	github.com/kylelemons/godebug v1.1.0
 	github.com/openconfig/gnmi v0.0.0-20200508230933-d19cebf5e7be
-	github.com/openconfig/goyang v0.0.0-20200506211124-77aa4ea52d03
+	github.com/openconfig/goyang v0.0.0-20200603222342-a0c8aa1ee274
 	github.com/pmezard/go-difflib v1.0.0
 	golang.org/x/net v0.0.0-20200520182314-0ba52f642ac2 // indirect
 	golang.org/x/sys v0.0.0-20200519105757-fe76b779f299 // indirect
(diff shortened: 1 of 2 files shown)


wenovus:
We still won't get automatic updates. AFAICS the benefit of versioning goyang is to create some stability for those who use the library directly rather than through ygot.

wenovus:
Versioning also has a downside that I currently don't know how to work around: when you're debugging goyang with ygot, you have to update goyang first before its effects on ygot can be tested. Turning off modules unfortunately makes ygot not compilable for reasons still unknown to me (has to do with proto version import).

robshakir:
Yes, sorry -- the only intention of versioning was to make the `go.mod` a bit cleaner.

For the latter problem, I guess this indicates that we have some parts of the goyang public API that are not well tested there -- and get their coverage via ygot. We probably shouldn't try and fix this.

Issue #394, opened 2020-05-29
Enumerated Type Unique Identification String Key


wenovus:
`enumgen.go` currently generates the unique ID incorrectly in a few cases:
- enumerated typedefs are identified using `module-name` + `enum/identity name`. This causes same-name enumerated values defined within the same module to be shadowed by each other.
- enumerated typedefs' `module-name` is incorrectly identified, and the user module is currently being used. This causes same-name enumerated values defined in different modules, but used within the same module, to shadow one another.
- enumeration leaves are identified using its path, but *without* the module name. This causes identically-pathed enumerations between different modules to shadow one another in the generated code.

When an instance of "shadowing one another in the generated code" occurs, it means only one of the conflicting enums is generated and used everywhere, and no error is emitted by the generator, and compilable, but incorrect code, is generated.

This issue can be solved by making the mapping of {enum value: unique ID} the one-to-one function that was originally intended.

wenovus:
@robshakir The above-mentioned PRs address all of the issues with leaf, typedef and union combinations for enumerated types, with one exception:

The one exception is the case of a typedef union containing a non-typedef enumeration. In this case, the current naming behaviour is to treat it as if the typedef were inlined into the YANG tree where it's used, meaning it's not de-duped at all. If these enums are also named `DefiningModule_TypedefName`, then another 30+ usages will be deduped into as few as 3 enums, including
```
E_OpenconfigMplsTypes_MplsLabel_Enum
E_OpenconfigPacketMatchTypes_PortNumRange_Enum
E_OpenconfigSegmentRoutingTypes_SrSidType_Enum
```
At least 20 de-duplications are for `mpls-label`, the most popular of these.

Should we just keep inlining them? Or is it also useful to make this change as well, which would be hidden by the same flag that hides #400 ?

If we should change this as well, we should first answer whether #245 is something we're planning to support, because that will affect what name to give to these enums.

robshakir:
Sorry @wenovus -- I'd missed this comment. Thanks for the detailed thought about this issue.

1. I think that we should de-duplicate these if we can, this seems like a good enhancement.
2. Let's make sure that this is covered by the same flag as #400 
3. On #245 -- this is likely something that we should support, especially given that there are now occurrences of this within the schema. Further design discussion below:

In the case of the multiple enumerations in a `union`, I'm not entirely sure how we should name these things. A quick brainstorm of a few options:

```yang

leaf foo {
  type union {
    type enumeration { enum BLUE; }
[...]

Issue #245, opened 2018-10-12
ygot doesn't support a union with multiple enums


upavlo:
When generating a proto message, if more than one enum is defined within a union then ygot will pick whatever the first enum that is declared in the schema, instead of listing them all within oneof message.

Consider example:
```yang
container bar {
    leaf my-mode {
        type union {
            type string;
            type enumeration {
                enum short;
                enum long;
                enum wide;
            }
            type enumeration {
                enum red;
                enum green;
                enum blue;
            }
        }
    }
}
```
When this is given to ygot the oneof message will contain one string and one enum (the first one).

In the real openconfig schema the following leafs have a union type with multiple enums:
- `/openconfig-network-instance/network-instances/network-instance/protocols/protocol/isis/interfaces/interface/levels/level/afi-safi/af/segment-routing/adjacency-sids/adjacency-sid/state/sid-id`
- `/openconfig-network-instance/network-instances/network-instance/protocols/protocol/isis/interfaces/interface/levels/level/afi-safi/af/segment-routing/adjacency-sids/adjacency-sid/config/sid-id`

Interestingly, if union contains multiple identifyrefs then each one will have an entry in the oneof message.

wenovus:
Confirmed that this an issue with the given OC paths. `sid-id`'s DYNAMIC enum value is ungenerated.

wenovus:
Fixing https://github.com/openconfig/goyang/issues/131 is a prerequisite.

PR #400, opened 2020-06-05
enumgen: Change typedef enum/identity prefix to defining module name

wenovus:
Currently, enumerated typedef names are used `UserModule_TypedefName` as opposed to `DefiningModule_TypedefName` as specified in the design doc. This change rectifies all but one exception, so it addresses the basic issues in #394.

The one exception is the case of a typedef union containing a non-typedef enumeration. In this case, the current naming behaviour is to treat it as if the typedef were inlined into the YANG tree, meaning it's not de-duped at all. If these enums are also named `DefiningModule_TypedefName`, then another 30+ usages will be deduped into as few as 3 enums, including
```
E_OpenconfigMplsTypes_MplsLabel_Enum
E_OpenconfigPacketMatchTypes_PortNumRange_Enum
E_OpenconfigSegmentRoutingTypes_SrSidType_Enum
```
At least 20 usages are for `mpls-label`, the most popular of these.

Should we just keep inlining them? Or is it also useful to make this breaking change as well, since we're already making a breaking change?

If we should change this as well, we should first answer whether #245 is something we're planning to support, because that will affect what name to give to these enums.


demo/bgp/main.go
```diff
@@ -70,7 +70,7 @@ func CreateDemoBGPInstance() (*oc.Bgp, error) {
 	// The returned struct represents the new peer that was created.
 	nPeer.PeerAs = ygot.Uint32(29636)
 	nPeer.Description = ygot.String("catalyst2 Services Ltd")
-	nPeer.PeerType = oc.OpenconfigBgp_PeerType_INTERNAL
+	nPeer.PeerType = oc.OpenconfigBgpTypes_PeerType_INTERNAL
```
robshakir, line 73: I think I misunderstood something in our implementation strategy here. 

AIUI, this should only occur when we've changed the flag to change the names? Since this is a backwards incompatible change then I think we want it to be flag-protected -- but it looks like it's being done by default here without a flag.

ygen/enumgen.go
```diff
@@ -224,11 +224,23 @@ func (s *enumSet) identityBaseKey(i *yang.Identity) string {
 	return fmt.Sprintf("%s/%s", definingModYANGName, i.Name)
 }
 
+// nodePath returns the full path of the node from the module name.
+func nodePath(n yang.Node) string {
```
robshakir, line 228: It was interesting to me that we didn't have a `util` function for this. We could consider putting it there if we think that this will be generally useful to others. I'm fine with it staying here though.

ygen/enumgen.go
```diff
@@ -224,11 +224,23 @@ func (s *enumSet) identityBaseKey(i *yang.Identity) string {
...
+	path := n.NName()
+	for n.ParentNode() != nil {
```
robshakir, line 230: I think this can be written in a more compact form, can't it?

```golang
for path := n.NName(); n.ParentNode() != nil; n = n.ParentNode() {
   path = n.NName() + "/" + path
}
return path
```

demo/bgp/main.go
wenovus: I will add the flags to help with the transition, but I'll assume the example files will change since they should showcase best practice.

robshakir: Yes, I think that'll be the case -- but this looks difficult for users to consume otherwise.

wenovus:
@robshakir There is one exception to the `DefiningMod_TypedefName` whose behaviour I want to double check as it's related to this flag. Please see https://github.com/openconfig/ygot/issues/394#issuecomment-642329123.

robshakir:
This LGTM, thanks for the change. Let's figure out whether we're blocked in #394, and then determine how to move forward from there.

PR #401, opened 2020-06-08
Add bool flag shorten_enum_leaf_names (default false).

wenovus:
This flag removes the module prefix from enumeration leaves when set to true.

Changes include
- Unit tests added for when enum names are not shortened. Most tests assume names are shortened. Note that now, `ResolveNameClashSet` is where the names are set, so the lowest level of unit tests go there.
- One integration test added for long names.
- inDefinedEnumsNoUnderscores is removed from TestFindEnumSet to save some code.
- example generated code uses `shorten_enum_leaf_names=true`.
- `skip_enum_dedup` flag also added to ypathgen's generator along with  `shorten_enum_leaf_names` -- this is because if the key is an enum type, the name needs to match with ygen's generated code. Luckily there are no enumeration leaf keys in the example generated code, so no changes occurred after the flag was added.

Note that none of these changes affect proto generation, because enumeration leaf type name generation doesn't use this rule due to nesting.


ygen/enumgen_test.go
```diff
@@ -751,26 +1147,33 @@ func TestResolveNameClashSet(t *testing.T) {
...
-			inDefinedEnums := tt.inDefinedEnums
-			if noUnderscores {
-				inDefinedEnums = tt.inDefinedEnumsNoUnderscores
-			}
 			for compressPaths := range map[bool]struct{}{false: {}, true: {}} {
-				t.Run(tt.name+fmt.Sprintf("@compressPaths:%v,noUnderscores:%v", compressPaths, noUnderscores), func(t *testing.T) {
+				t.Run(tt.name+fmt.Sprintf("@compressPaths:%v,noUnderscores:%v,shortenEnumLeafNames:%v", compressPaths, noUnderscores, tt.inShortenEnumLeafNames), func(t *testing.T) {
 					s := newEnumGenState()
-					for k, v := range inDefinedEnums {
+					for k, v := range tt.inDefinedEnums {
```
robshakir, line 1162: Does this logic always hold for these tests -- it seems like it probably does today, but it also looks slightly fragile insofar that it seems like there might be cases where we could end up with deduplication occurring in the proto case where we have no underscores, and uppercase being used. Am I misparsing here?

wenovus: No your observation is valid, this mapping is definitely not sound currently due to the case you pointed out, and others as well.

I didn't write this to assert this case -- my only purpose was to avoid repeating `tt.inDefinedEnumsNoUnderscores` for every test case, because none of the tests trigger these corner cases.

I see this optimization is too complex because it can cause misunderstanding due to the intrinsic complexity of `definedEnums`, so I've added `tt.inDefinedEnumsNoUnderscores` now.

robshakir: Great, thanks. I get that this is more verbose, but the general rule that I think we try to follow is that we don't introduce logic in a test wherever possible. Of course, there's some cases where that is unavoidable, or the cost of not having the logic is so high that we want to avoid it - but this looks to be a case where the duplication is verbose, but useful.

PR #402, opened 2020-06-12
Add a new mechanism for rendering rfc7951 JSON.

robshakir:
```
 * (M) ygot/{render,render_test}.go
  - Some users wish to be able to render a specific field to
    RFC7951 JSON directly, such that a particular subtree can be
    rendered.
    Today, this is not possible since we require arguments to the
    rendering functions to be specified as GoStructs. This PR adds
    a new rendering method that specifically takes an interface{}
    and returns a string such that these users can call this method
    generically for any valid input type of a ygen generated Go file.
    We lose some type-safety in this mechanism, but this is similar to
    the interface provided by json.Marshal.
    This method is implemented only for RFC7951 JSON, as we aim to
    move away from the 'internal' format (which today is only
    supported for marshalling and not unmarshalling).
```

### DIFF
--- a/ygot/render.go
+++ b/ygot/render.go
@@ -963,13 +963,7 @@ func structJSON(s GoStruct, parentMod string, args jsonOutputConfig) (map[string
 			continue
 		}
 
-		var value interface{}
-
-		/*if util.IsYgotAnnotation(fType) {
-			value, err = jsonAnnotationSlice(field)
-		} else {*/
-		value, err = jsonValue(field, pmod, args)
-
+		value, err := jsonValue(field, pmod, args)
 		if err != nil {
 			errs.Add(err)
 			continue
--- a/ygot/render.go
+++ b/ygot/render.go
@@ -884,6 +884,71 @@ func ConstructInternalJSON(s GoStruct) (map[string]interface{}, error) {
 	})
 }
 
+// EmitRFC7951 renders the supplied interface to RFC7951-compatible JSON. The argument
+// supplied must be a valid type within a generated ygot GoStruct - but can be a member
+// field of a generated struct rather than the entire struct - allowing specific fields
+// to be rendered. The supplied args are used to control RFC7951-specific behaviours
+// of the rendering. The rendered JSON is returned as a string.
+func EmitRFC7951(d interface{}, args *RFC7951JSONConfig) (string, error) {
+	j, err := marshalRFC7951(d, args)
+	if err != nil {
+		return "", err
+	}
+	js, err := json.Marshal(j)
+	if err != nil {
+		return "", fmt.Errorf("could not marshal JSON, %v", err)
+	}
+
+	return string(js), nil
+}
+
+// marshalRFC7951 marshals the supplied interface{} d into an interface{} which
+// can be handed directly to json.Marshal. It has specific handling for struct,
+// map, and slice fields within the supplied interface{} to ensure RFC7951-compatible
+// marshalling.
+func marshalRFC7951(d interface{}, args *RFC7951JSONConfig) (interface{}, error) {
+	v := reflect.ValueOf(d)
+	gsT := reflect.TypeOf((*GoStruct)(nil)).Elem()
+
+	jargs := jsonOutputConfig{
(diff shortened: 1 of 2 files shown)


robshakir:
Big thumbs up for `staticcheck` here @wenovus -- it pointed something out that I'd missed, which meant that I could really simplify this code. Thanks for setting it up!

ygot/render_test.go
```diff
@@ -2903,3 +2903,95 @@ func TestFindUpdatedLeaves(t *testing.T) {
...
+		in:   map[string]*renderExample{},
+		want: `null`,
+	}, {
+		desc: "nil string pointer",
+		in: func() *string {
+			var s *string
+			return s
+		}(),
+		want: `null`,
+	}}
```
wenovus, line 2980: Add test for empty type

```suggestion
        }, {
                desc: "empty type",
                in: &renderExample{
                        Empty: true,
                },
                want: `{"empty":[null]}`,
	}}
```

ygot/render.go
```diff
@@ -884,6 +884,27 @@ func ConstructInternalJSON(s GoStruct) (map[string]interface{}, error) {
 	})
 }
 
+// EmitRFC7951 renders the supplied interface to RFC7951-compatible JSON. The argument
+// supplied must be a valid type within a generated ygot GoStruct - but can be a member
```
wenovus, line 888:
```suggestion
// supplied must be a valid type within a generated ygot GoStruct (i.e. all types except the root GoStruct type) - but can be a member
```

ygot/render_test.go
```diff
@@ -2903,3 +2903,95 @@ func TestFindUpdatedLeaves(t *testing.T) {
...
+	}, {
+		desc: "empty map",
+		in:   map[string]*renderExample{},
+		want: `null`,
+	}, {
+		desc: "nil string pointer",
+		in: func() *string {
+			var s *string
+			return s
+		}(),
```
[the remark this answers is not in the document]
robshakir: Thanks!

ygot/render_test.go
robshakir: Great, thanks! Added.

ygot/render.go
robshakir: Can you explain what you meant here please? I'm not sure I understood what `root` means in this context?

robshakir:
Thanks for the reviews.

@ejbrever - PTAL, I added arguments (trying to keep a consistent behaviour here despite the fact that we don't have the same arguments as `EmitJSON`). It now works similarly to other functions that take an expanded slice of arguments, so you can say:

```golang
ygot.Marshal7951(d, JSONIndent("\t"), &RFC7951Config{AppendModuleNames: true})
```

This also cleans up the base syntax such that you can just do:

```
ygot.Marshal7951(d)
```
 
without needing the extraneous `nil` at the end.

ygot/render.go
wenovus: I just meant that this cannot be used for the root of all the models, because from my reading of the code it seems that should be handled by `EmitJSON`.

ygot/render.go
```diff
@@ -884,6 +888,66 @@ func ConstructInternalJSON(s GoStruct) (map[string]interface{}, error) {
...
+// Marshal7951 renders the supplied interface to RFC7951-compatible JSON. The argument
+// supplied must be a valid type within a generated ygot GoStruct - but can be a member
+// field of a generated struct rather than the entire struct - allowing specific fields
+// to be rendered. The supplied arguments control the JSON marshalling behaviour - both
+// base JSON Marshal (e.g., indentation), as well as RFC7951 specific options such as
+// YANG module names being appended.
+// The rendered JSON is returned as a byte slice - in common with json.Marshal.
+func Marshal7951(d interface{}, args ...Marshal7951Arg) ([]byte, error) {
+	var (
+		rfccfg *RFC7951JSONConfig
```
ejbrever, line 914: big nit...can you make this var rfcCfg just to help readability a little.

robshakir: Of course -- done!

PR #403, opened 2020-06-14
Add bool flag use_defining_module_for_typedef_enum_names (default false)

wenovus:
This flag causes typedef enumeration/identity names to be prefixed with
its defining module name, per [design doc](docs/design.md), instead of the current
behaviour of using the residing module. It also de-dups identical typedefs together when
used between different modules.

Using this flag also fixes a related bug. Namely, if a collision of a
typedef name occurred, behaviour with flag set to false silently ignores
this fact, and uses one of the enums in place of all conflicting enums.
Turning on this flag would now cause a code generation error to be
raised when this conflict occurs.

This flag affects proto generation.

Changes include
- Unit tests. New unit tests for typedef name conflict that exists in new behaviour but not with old behaviour.
- example generated code uses
`use_defining_module_for_typedef_enum_names=true` everywhere.


README.md
```diff
@@ -54,7 +54,7 @@ go run $GOPATH/src/github.com/openconfig/ygot/generator/generator.go -path=yang
 To allow this file to be auto-created, you can place a command which allows this code generation to be done automatically, either by creating a file within the YANG directory, or directly embedding this command within the source file that populates the structures. For an example, see the `demo/getting_started/main.go` file which includes:
 
 ```
-//go:generate go run ../../generator/generator.go -path=yang -output_file=pkg/ocdemo/oc.go -package_name=ocdemo -generate_fakeroot -fakeroot_name=device -compress_paths=true  -exclude_modules=ietf-interfaces yang/openconfig-interfaces.yang
+//go:generate go run ../../generator/generator.go -path=yang -output_file=pkg/ocdemo/oc.go -package_name=ocdemo -generate_fakeroot -fakeroot_name=device -compress_paths=true -shorten_enum_leaf_names -use_defining_module_for_typedef_enum_names -exclude_modules=ietf-interfaces yang/openconfig-interfaces.yang
```
robshakir, line 57: Perhaps we can go with something shorter here? `-typedef_with_defmod` maybe? I think users will need to look up these flags anyway.

PR #405, opened 2020-06-23
Remove Resolve function from generated code

wenovus:
Merge its functionality into `ResolvePath` in `ygot/ygot`. This reduces complexity in the generated code.

The reason why this function was needed in the generated code instead of being able to reside in `ygot/ygot` was it needed to access special fields from the generated root `PathStruct`. Using the pattern analogous to `ygot.NodePath`, a base struct `DeviceRootBase` is defined along with an interface that the generated root object embeds. This allows `ygot` to access those root struct fields.

### DIFF
--- a/ygot/path_types.go
+++ b/ygot/path_types.go
@@ -14,7 +14,11 @@
 
 package ygot
 
-import gpb "github.com/openconfig/gnmi/proto/gnmi"
+import (
+	"fmt"
+
+	gpb "github.com/openconfig/gnmi/proto/gnmi"
+)
 
 const (
 	// PathStructInterfaceName is the name for the interface implemented by all
@@ -23,6 +27,9 @@ const (
 	// PathBaseTypeName is the type name of the common embedded struct
 	// containing the path information for a path struct.
 	PathBaseTypeName = "NodePath"
+	// FakeRootBaseTypeName is the type name of the fake root struct which
+	// should be embedded within the fake root path struct.
+	FakeRootBaseTypeName = "DeviceRootBase"
 )
 
 // PathStruct is an interface that is implemented by any generated path struct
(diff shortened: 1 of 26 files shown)


ygot/path_types.go
```diff
@@ -47,9 +54,38 @@ type NodePath struct {
...
+
+func NewDeviceRootBase(id string) *DeviceRootBase {
+	return &DeviceRootBase{NodePath: &NodePath{}, id: id, CustomData: map[string]interface{}{}}
+}
+
+// DeviceRootBase represents the fakeroot for all YANG schema elements.
+type DeviceRootBase struct {
+	*NodePath
+	id         string
+	CustomData map[string]interface{}
```
progressnerd, line 73: if GetCustomData is exported, does CustomData need to be exported, too?

ygot/path_types.go
```diff
@@ -47,9 +54,38 @@ type NodePath struct {
...
 }
 
-// ResolvePath is a helper which returns the root PathStruct and absolute path
-// of a PathStruct node.
-func ResolvePath(n PathStruct) (PathStruct, []*gpb.PathElem, []error) {
+// FakeRootPathStruct is an interface that is implemented by the fake root path
+// struct type.
+type FakeRootPathStruct interface {
+	PathStruct
+	GetId() string
```
progressnerd, line 61: If you don't need to export CustomData, could change the method names to the cleaner "Id()" and "CustomData()." The "Get" prefixed is sometimes frowned on as a "stop word" in Go.

ygot/path_types.go
```diff
@@ -47,9 +54,38 @@ type NodePath struct {
 	p             PathStruct
 }
 
-// ResolvePath is a helper which returns the root PathStruct and absolute path
-// of a PathStruct node.
-func ResolvePath(n PathStruct) (PathStruct, []*gpb.PathElem, []error) {
+// FakeRootPathStruct is an interface that is implemented by the fake root path
+// struct type.
+type FakeRootPathStruct interface {
```
progressnerd, line 59: does FakeRootPathStruct itself need to be exported?

ygot/path_types_test.go
```diff
@@ -21,14 +21,20 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+type deviceRoot struct {
```
progressnerd, line 24: why is this type necessary? Why can't you just use Device directly?

ygot/path_types.go
wenovus: - The generated package needs to embed the type in order to define new methods for it for path creation.
- The generated root type's `CustomData` needs to allow modification after construction.
- When embedding an out-of-package type, unexported fields are inaccessible to the importer package.

So, the only ways of doing this is to either to define a setter or make it exported.

I guess having a setter makes more sense, so I'll do that, and unexport `CustomData`.

ygot/path_types_test.go
wenovus: I want to showcase that the intended usage is to allow embedding.

ygot/path_types.go
```diff
@@ -47,9 +54,46 @@ type NodePath struct {
...
+	return d.id
+}
+
+// CustomData returns the customData field of the DeviceRootBase struct.
+func (d *DeviceRootBase) CustomData() map[string]interface{} {
+	return d.customData
+}
+
+// SetCustomDataKey modifies an entry in the customData field of the DeviceRootBase struct.
+func (d *DeviceRootBase) SetCustomDataKey(key string, val interface{}) {
```
progressnerd, line 90: Maybe it's not standard in ygot, but I'm used to seeing the verb "put" for maps. And then maybe you no longer need the "key" part, i.e. just "PutCustomData"

progressnerd, line 90: This function isn't strictly necessary because anyone could do CustomData()[key] = val. I could see the case for syntactic sugar, though.

wenovus: Good point, I think it's less obvious that `CustomData()` is the way to modify the map, so I would judge to keep the sugar.

PR #408, opened 2020-06-26
Merge ypathgen's generator into ygen's generator

wenovus:
- Organized generator flags by category.
- Made "-" output to stdout. Previously "" meant nothing was output, so the previous code path that output to stdout was unreachable.


wenovus:
I think Rob is more familiar with this code, but this is related with previous PRs, so I've added @progressnerd here as well.

generator/generator.go
```diff
@@ -50,31 +51,42 @@ const (
...
-	skipEnumDedup          = flag.Bool("skip_enum_deduplication", false, "If set to true, all leaves of type enumeration will have a unique enum output for them, rather than sharing a common type (default behaviour).")
-	shortenEnumLeafNames   = flag.Bool("shorten_enum_leaf_names", false, "If also set to true when compress_paths=true, all leaves of type enumeration will by default not be prefixed with the name of its residing module.")
+
+	// Flags only used for PathStruct generation.
+	schemaStructPath                     = flag.String("schema_struct_path", "", "The import path to use for ygen-generated schema structs. This should only be specified if the generated path structs are to reside in a different package than the schema structs.")
+	preferOperationalStateForPathStructs = flag.Bool("path_struct_prefer_operational_state", true, "If set to true, state (config false) fields in the YANG schema are preferred over intended config leaves in the generated Go code for constructing YANG paths. This flag is only valid for compress_paths=true and exclude_state=false.")
```
robshakir, line 87: It's not clear to me what having a different value specified for `preferOperationalStateForPathStructs` and `preferOperationalState` does -- don't they need to be aligned such that there is definitely a struct generated for each received telemetry path?

generator/generator.go
```diff
@@ -50,31 +51,42 @@ const (
...
-	packageName            = flag.String("package_name", "ocstructs", "The name of the Go package that should be generated.")
-	outputFile             = flag.String("output_file", "", "The file that the generated Go code should be written to.")
-	outputDir              = flag.String("output_dir", "", "The directory that the Go package should be written to.")
-	ignoreCircDeps         = flag.Bool("ignore_circdeps", false, "If set to true, circular dependencies between submodules are ignored.")
+	ocStructsOutputFile     = flag.String("output_file", "", "The file that the generated Go code for manipulating YANG data should be written to. Specify \"-\" for stdout. If both this and output_dir are empty, then schema structs will not be output.")
+	ocStructsOutputDir      = flag.String("output_dir", "", "The directory that the Go package for manipulating YANG data should be written to.")
+	ocPathStructsOutputFile = flag.String("path_structs_output_file", "", "The file that the generated Go code for YANG path construction will be generated. If empty, then path structs are not generated. Specify \"-\" for stdout.")
```
robshakir, line 56: It feels a bit strange to be implicit here about whether we're generating path structs or not. Can we rather have an approach whereby we have a flag that controls whether path structs should be output, and then a set of flags that specify a single output file, or an output directory?

I understand that splitting the output files in the base case sounds reasonable - but if we do want to do this, I think we need to think a little more carefully about making this clear to the user.

generator/generator.go
```diff
@@ -228,8 +248,8 @@ func main() {
 		}
 	}
 
-	if *outputFile != "" && *outputDir != "" {
-		log.Exitf("Error: cannot specify both outputFile (%s) and outputDir (%s)", *outputFile, *outputDir)
+	if *ocStructsOutputFile != "" && *ocStructsOutputDir != "" {
+		log.Exitf("Error: cannot specify both ocStructsOutputFile (%s) and ocStructsOutputDir (%s)", *ocStructsOutputFile, *ocStructsOutputDir)
 	}
```
robshakir, line 253: Can we add error handling for the case that the user gives us a path to the generated structs, but we're also generating them?

generator/generator.go
```diff
@@ -271,27 +291,80 @@ func main() {
...
 		default:
 			// Assign the newly created filehandle to the outfh, and ensure
 			// that it is synced and closed before exit of main.
-			outfh = genutil.OpenFile(*outputFile)
+			outfh = genutil.OpenFile(*ocStructsOutputFile)
 			defer genutil.SyncFile(outfh)
 		}
 
 		writeGoCodeSingleFile(outfh, generatedGoCode)
+	} else if *ocStructsOutputDir != "" {
```
robshakir, line 313: Consider changing this to using a `switch` statement?

generator/generator.go
wenovus: Thanks! Good catch, I was rather careless there.

generator/generator.go
wenovus: I've added both `generate_structs` and `generate_path_structs` flags to also keep them consistent. However their default values are different to keep backwards compatibility.

`output_dir` is shared across both types of generations.
I've also manually tested the 8 combinations of true/false for these flags, and single vs. multi-file output such that all the tests in the repo continue to pass.

generator/generator.go
wenovus: I've also added error handling for the opposite case: requiring users to supply the path to the generated structs when only path structs are being generated.

PR #409, opened 2020-06-30
Add option to split GoStructs by number of files instead of by first letter

wenovus:
Advantages:
- The names of the output struct files are known beforehand, which are useful for build rules that require them (e.g. bazel).
- structs are split more evenly than splitting by the first letter.

Note on `ypathgen` change:
`ypathgen` is using a similar algorithm. The algorithm was improved (hence the change), and here duplicated in `generator`.
See [comment description](https://github.com/openconfig/ygot/compare/split-structs-by-number-of-files?expand=1#diff-9cb29f187178d462bdb058663ac1d886R285) on improvement that was made.


generator/generator.go
```diff
@@ -196,13 +200,84 @@ func makeOutputSpec(goCode *ygen.GeneratedGoCode) map[string]codeOut {
...
+}
+
 // writeGoCodeMultipleFiles writes the input goCode to a set of files as specified
 // by specification returned by output spec.
-func writeGoCodeMultipleFiles(dir string, goCode *ygen.GeneratedGoCode) error {
-	out := makeOutputSpec(goCode)
+func writeGoCodeMultipleFiles(dir string, goCode *ygen.GeneratedGoCode, fileN int) error {
+	var out map[string]codeOut
+	if fileN == 0 {
+		out = splitCodeByLeadingLetter(goCode)
```
robshakir, line 271: Shall we just deprecate this option? I don't know of anyone that is using it - and it seems to be more problematic to make work with build systems that expect a specific number of files to exist.

generator/generator.go
```diff
@@ -146,17 +148,19 @@ func writeIfNotEmpty(b io.StringWriter, s string) {
 type codeOut struct {
 	// contents is the code that is contained in the output file.
 	contents string
+	// writeEvenIfEmpty forces the file to be written even with empty contents.
+	writeEvenIfEmpty bool
```
robshakir, line 152: Consider calling this `writeIfEmpty` to be a bit more pithy?

ypathgen/testdata/structs/openconfig-simple-31.path-txt
```diff
@@ -14,32 +14,180 @@ import (
 	"github.com/openconfig/ygot/ygot"
 )
```
robshakir, line 16: Is the diff here solely because of the change to the algorithm for splitting that you make above?

wenovus: Yes, due to ceiling being used on the structN/fileN to determine structs per file.

robshakir: Great, thanks.

generator/generator.go
wenovus: Removed it. I also removed the `methods.go` file and merged them into the `structs-%d.go` files, because those also scale up pretty quickly. I merged them because I thought it might make sense to put the methods alongside the structs. However, I also see how splitting them might make the structs file easier to browse, so I can put them into a `methods-%d.go` file if the latter makes more sense.

generator/generator.go
wenovus: Removed this struct.

generator/generator.go
robshakir: I think this makes sense - the only real advantage of this was that I thought that folks wouldn't really want to look at the methods, however, I'm not sure it ever really made entire sense. Let's go with this approach and we can always re-do this if we want to in the future.

PR #411, opened 2020-07-07
Add support for lists to protomap.

robshakir:
This CL adds support for lists to protomap, such that protobuf messages that contain `repeated` fields (representing YANG lists) can be mapped from input protobuf messages.

In addition, it refactors the keys used for returned maps to use `gnmi.Path` rather than `string`, since this vastly improves the robustness of path manipulation in this code.


robshakir:
Sorry for the larger than usual diff, it's mostly in the test protobuf, and generated `pb.go` file.

protomap/proto.go
```diff
@@ -18,63 +18,361 @@
...
+// specified. It returns a parsedListField describing the individual field supplied.
+func parseListField(fd protoreflect.FieldDescriptor, v protoreflect.Value, basePath *gpb.Path) (*parsedListField, error) {
+	if fd.IsMap() || fd.IsList() {
+		return nil, fmt.Errorf("invalid field in list, %s", fd.FullName())
+	}
+
+	if fd.Kind() == protoreflect.MessageKind {
+		if t, ok := v.Message().Interface().(proto.Message); ok {
+			// The only case of having proto.Message in a list key is when it is the
+			// value itself, therefore return this value.
```
wenovus, line 286: I feel like "the value itself" requires too much context to make sense of -- I thought it meant the entire list value itself at first. Maybe "the message containing the list's value portion" is better?

protomap/proto.go
```diff
@@ -18,63 +18,361 @@
...
+
+	p := &parsedListField{
+		keys:         map[string]string{keyName: kv},
+		mappedValues: map[*gpb.Path]interface{}{},
+	}
+
+	for _, path := range mappedPaths {
+		childElems := util.TrimGNMIPathElemPrefix(path, removePathKeys(basePath))
+		newPath := &gpb.Path{Elem: append(basePath.Elem, childElems.Elem...)}
+		p.mappedValues[newPath] = v.Interface()
```
wenovus, line 325: Can use the `resolvePath` helper here.

protomap/proto.go
```diff
@@ -18,63 +18,361 @@
...
+			return fmt.Errorf("could not parse a field within the list %s , %v", fd.FullName(), listParseErr)
+		}
+
+		// Handle recursing down the tree for the list that we just extracted.
+		p := proto.Clone(listPath).(*gpb.Path)
+		if basePath != nil {
+			// This is the first time that we have found a path that requires a
+			// data tree path, not a schema tree path.
+			p = resolvedPath(basePath, listPath)
+		}
```
wenovus, line 234: I think having the `if basePath != nil` check within `resolvedPath` makes it slightly more convenient as I see this being written in another place as well.

If clone is also removed, that line and this block can just become `p := resolvedPath(basePath, listPath)`.

protomap/proto.go
```diff
@@ -18,63 +18,361 @@
...
+			}
+			return true
+		})
+
+		if listParseErr != nil {
+			return fmt.Errorf("could not parse a field within the list %s , %v", fd.FullName(), listParseErr)
+		}
+
+		// Handle recursing down the tree for the list that we just extracted.
+		p := proto.Clone(listPath).(*gpb.Path)
```
wenovus, line 229: This clone is not required right?

protomap/proto_test.go
```diff
@@ -73,20 +92,178 @@ func TestPathsFromProto(t *testing.T) {
...
+			}},
+		},
+		wantErrSubstring: "received list key with leaf names that do not match",
 	}}
 
 	for _, tt := range tests {
 		t.Run(tt.desc, func(t *testing.T) {
-			got, err := pathsFromProto(tt.inMsg)
+			got := map[*gpb.Path]interface{}{}
+			err := pathsFromProtoInternal(tt.inMsg, got, tt.inBasePath)
```
wenovus, line 254: `inBasePath` doesn't seem to be tested by any test case. Is there a need to test it though? What about just testing `pathsFromProto()` directly instead of the internal function? That seems right to me.

wenovus:
BTW this protoreflect package is really neat!

PR #413, opened 2020-07-11
Add DefiningType Helper

wenovus:
In #394, it has been agreed to name non-typedef enumerations residing in typedef unions as if they were typedefs themselves.

When figuring out whether an enumeration should be treated as a typedef, we need to know whether it's within a typedef union, and if it is, use the union typedef to name the enumeration.

Since goyang's `YangType` doesn't tells us whether or not it's a union subtype, we have to figure this out somehow.  I considered augmenting `YangType` to tell us in which type it resides, but I'm not sure whether `YangType`s can be reused as union subtypes, in which case there would be multiple such "subsuming" types.

In this PR, I define the "subsuming type" of each `YangType` to mean the closest/innermost type that's not a union subtype:
case 1: for any type not within a union, the subsuming type is itself.
case 2: for any type within a union, the subsuming type is the first containing union type that's a typedef, or highest union type definition, whichever one is closest to the type.

`EnumeratedUnionSubsumingTypes()` extracts these "subsuming" types out of a union.


util/yang.go
```diff
@@ -361,6 +362,37 @@ func EnumeratedUnionTypes(types []*yang.YangType) []*yang.YangType {
 	return eTypes
 }
 
+// EnumeratedUnionSubsumingTypes returns the subsuming types of the enumerated
```
robshakir, line 365: Do you mean the subsumed types, rather than the subsuming types? i.e., we're walking from a parent downwards so everything that is below us is presumably subsumed by the supplied parent -- whereas subsuming types rather sounds like we're walking up from a child to find all of the parent types that include us.

util/yang_test.go
```diff
@@ -1024,6 +1032,67 @@ func TestEnumeratedUnionTypes(t *testing.T) {
...
+				Name: "enumeration",
+				Kind: yang.Yenum,
+			}},
+			wantParentTypes: []*yang.YangType{{
+				Name: "parent-type",
+				Kind: yang.Yunion,
+			}, {
+				Name: "inner-iref",
+				Kind: yang.Yidentityref,
+			}, {
```
robshakir, line 1089: Can you explain why we end up with two different `parent-type` here? I presume it is because `parent-type` includes both an enumeration itself and has a child union that includes an enumeration too?

util/yang.go
```diff
@@ -361,6 +362,37 @@ func EnumeratedUnionTypes(types []*yang.YangType) []*yang.YangType {
...
+// that's not a union subtype:
+// case 1: for any type not within a union, the subsuming type is itself.
+// case 2: for any type within a union, the subsuming type is the first
+// containing union type that's a typedef, or highest union type definition,
+// whichever one is closest to the type.
+func EnumeratedUnionSubsumingTypes(subsumingType *yang.YangType) []*yang.YangType {
+	return enumeratedUnionSubsumingTypes(subsumingType.Type, subsumingType)
+}
+
+func enumeratedUnionSubsumingTypes(types []*yang.YangType, subsumingType *yang.YangType) []*yang.YangType {
```
robshakir, line 379: I wonder if `[]*yang.YangType` is the best return type here -- it seems like we actually just want a tree that gives us the hierarchy of types that exist for a particular leaf, right? Should we define some form of type tree that we return to the user - such that this data structure is a bit clearer?

In the tests, having repeated `parent-type` made it a little difficult to understand what was being represented there -- we've unrolled a tree, but we've denormalised the output based on traversal approach.

It may be that I'm not quite getting why we need to have this structure -- so I'll come back to this comment after reviewing the functionality in the subsequent CL.

util/yang.go
```diff
@@ -348,6 +348,7 @@ func addToEntryMap(to, from map[string]*yang.Entry) map[string]*yang.Entry {
 
 // EnumeratedUnionTypes recursively searches the set of yang.YangTypes supplied to
 // extract the enumerated types that are within a union.
+// It returns the enumerated types in tree order.
```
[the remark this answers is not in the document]
wenovus: Yes, The set of input yang.YangTypes is expected to be the slice of types of the union type.

util/yang.go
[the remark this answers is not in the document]
wenovus: I tried to clear things up, PTAL

util/yang.go
wenovus: I do mean the subsuming types. It's just that the child types don't have a parent pointer, so the implementation currently has to go downwards.

util/yang_test.go
wenovus: I added a docstring comment for the test table's `wantSubsumingTypes` field. Let me know if this is clear enough.

util/yang.go
wenovus: I feel like I made the concept of "subsuming type" too complex in this change. I've tried to make its meaning simpler to understand. The reason it's written this way is to make the code easier to write, as the users of this "subsuming type" finder function are the very same places where `enumeratedUnionType` is used.

I've changed the name of the function, it may be too long now, but I mulled over it and couldn't come up with a shorter one that still encapsulates the meaning.

wenovus:
Now I'm thinking maybe `DefiningTypesOfEnumeratedUnionTypes` may be better than `SubsumingTypes...`.

Yes, let me change that.

util/yang.go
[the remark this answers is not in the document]
wenovus: I changed the name, now hopefully it's much clearer.

util/yang.go
wenovus: I renamed and simplified the semantics of this function. Hopefully it's not confusing anymore.

util/yang.go
wenovus: As above

PR #414, opened 2020-07-11
Uniquely Identify Enumerations within a Typedef Union

### LINKED ISSUE #394: Enumerated Type Unique Identification String Key

wenovus:
* builtin `enumeration` types defined within a non-builtin (i.e. `typedef`)
  union type are named as if the union type represented a non-builtin
  `enumeration` containing the builtin `enumeration`, with `_Enum` appended to
  the name.  So, if the union were named `baz`, and resides in module `bar`, the
  Go name is `Bar_Baz_Enum`.
  * This mapping is handled by `enumgen.go`:`resolveEnumeratedUnionEntry`.
* A new `subsumingType` field has been added to `resolveTypeArgs` to help with name resolution for this case.
* Fix generator/generator.go bug where `useDefiningModuleForTypedefEnumNames` wasn't being passed in to the config.
* Added new test for multiple enumerations in `testdata/modules/enum-multi-module.yang`. Manually checked that ygen/testdata/structs/enum-multi-module.formatted-txt builds. The file contains a TODO for fixing #245, for which https://github.com/openconfig/goyang/issues/131 has been opened to track the fix in goyang.
* #245 is actually partially fixed (notably `sid-id`, the only multiple enumeration instance in Openconfig, is now generated correctly). This is just luck however, as other combinations of multiple enumerations within a union fail, as demonstrated in the above bullet.
* the old behaviour's output having the duplicated enumeration types remain unaffected: enum-module.residing-module-typedef-enum-name.formatted-txt.
* Fixes #394 (last PR for the issue).
* Manually checked that `exampleoc` contains the same output when `useDefiningModuleForTypedefEnumNames` is false.


demo/gnmi_telemetry/gnmi.go
```diff
@@ -105,7 +105,7 @@ func CreateAFTInstance() (*oc.Device, error) {
 	// Each union has a method that is named To_X where the X is the union type, associated
 	// with the struct that the union is within. This attempts to return the right type
 	// based on the input interface.
-	expNull, err := nh.To_NetworkInstance_Afts_NextHop_PushedMplsLabelStack_Union(oc.NextHop_PushedMplsLabelStack_IPV4_EXPLICIT_NULL)
+	expNull, err := nh.To_NetworkInstance_Afts_NextHop_PushedMplsLabelStack_Union(oc.OpenconfigMplsTypes_MplsLabel_Enum_IPV4_EXPLICIT_NULL)
```
robshakir, line 108: Nice, this is much better naming.

docs/design.md
```diff
@@ -143,6 +143,12 @@ struct names.
   and the `typedef` name - i.e., `typedef bar { type enumeration { ... }}` in
   module `baz` is represented by an enumerated type named `Bar_Baz`.
   * This mapping is handled by `enumgen.go`:`resolveTypedefEnumeratedName`.
+* builtin `enumeration` types defined within a non-builtin (i.e. `typedef`)
```
robshakir, line 146: Can I suggest some alternate wording here?

```
* where an `enumeration` is defined within a `typedef` that contains a `union` - 
  the enumerated language type that is generated is named according to the name 
  of the `typedef` with `_Enum` appended to the name. 

  For example:

module bar {
  ...
  typedef baz {
     type union {
        type enumeration { ... }
        type string;
     }
  }
}

  would result in a type named `Bar_Baz_Enum` being generated in the output
  code.
```

testdata/modules/enum-multi-module.yang
```diff
@@ -0,0 +1,45 @@
...
+    leaf inline-multi-value {
+      type union {
+        type uint32;
+        type t:td;
+        type enumeration {
+          enum ONE;
+          enum TWO;
+          enum THREE;
+        }
+        // TODO(wenbli): td2's enum, td-enum, and the second enumeration are not being
```
robshakir, line 33: Can you reference to goyang issue that we opened here?

ygen/codegen_test.go
```diff
@@ -805,6 +817,9 @@ func TestSimpleStructs(t *testing.T) {
 	}}
 
 	for _, tt := range tests {
+		if tt.name != "enumeration behaviour - multiple enumerations within a union" {
```
robshakir, line 820: Stale test skipping code?

BTW -- you can do `go test -run <regexp>` to run an individual test case to avoid this issue when debugging a particular test. If you have spaces in the test name, then they're replaced with `_`.  Subtests can also be referred to by using `TestName/Subtest_Name` as that pattern.

ygen/enumgen.go
```diff
@@ -103,10 +104,14 @@ func (s *enumSet) enumeratedUnionEntry(e *yang.Entry, compressPaths, noUnderscor
 		case t.Enum != nil:
 			var enumName string
 			var err error
-			if util.IsYANGBaseType(t) {
+			enumNameSake := t
+			if useDefiningModuleForTypedefEnumNames {
+				enumNameSake = subsumingTypes[i]
```
robshakir, line 109: It wasn't clear to me why this gives us the right 'subsuming' type -- why is the index here of `EnumeratedUnionTypes` the right index to use in `EnumeratedUnionSubsumingTypes`, since AIUI, if we have a `union` type it'll be returned more than once (see comments on `parent-type` in the other change), so I don't think that I understand why there would be a 1:1 correspondence between the two.

I wonder whether it's best to have something whereby we do:

`typeTree := buildTypeTree(e.Type.Type)`

and then later:
 
`subsumingType := typeTree.nonEnumeratedParentTypeFor(type)`

This would remove there being any dependency on ordering.

ygen/enumgen.go
```diff
@@ -176,7 +181,11 @@ func (s *enumSet) enumeratedTypedefTypeName(args resolveTypeArgs, prefix string,
 	// types which is defined in RFC6020/RFC7950) then we establish what the type
 	// that we must actually perform the mapping for is. By default, start with
 	// the type that is specified in the schema.
-	if !util.IsYANGBaseType(args.yangType) {
+	// If the type is a built-in enumeration residing in a typedef union
+	// (indicated by args.subsumingType), and useDefiningModuleForTypedef
+	// is set, then it's also named as a typedef, with its name based on
+	// the union typedef's name.
+	if !util.IsYANGBaseType(args.yangType) || (useDefiningModuleForTypedefEnumNames && args.yangType.Kind == yang.Yenum && args.subsumingType != nil && !util.IsYANGBaseType(args.subsumingType)) {
```
robshakir, line 188: I'd suggest a helper function (could be in-lined) for this:

```
(useDefiningModuleForTypedefEnumNames && args.yangType.Kind == yang.Yenum && args.subsumingType != nil && !util.IsYANGBaseType(args.subsumingType))
```

Can you explain in what circumstances `util.IsYANGBaseType(args.subsumingType)` can ever return true for a `args.subsumingType`? I assume that it's the case where we have a union that is not a typedef union, right - but I'm not sure we would ever exercise this code path for those.

ygen/enumgen.go
```diff
@@ -229,47 +238,41 @@ func (s *enumSet) identityBaseKey(i *yang.Identity) string {
...
-				typeName = fmt.Sprintf("%s_Enum", enumTypes[0].Name)
-			}
-		case 0:
-			return "", "", nil, fmt.Errorf("enumerated type had an empty union within it, path: %v, type: %v, enumerated: %v", e.Path(), e.Type, enumTypes)
-		default:
-			return "", "", nil, fmt.Errorf("multiple enumerated types within a single enumeration not supported, path: %v, type: %v, enumerated: %v", e.Path(), e.Type, enumTypes)
+	if subsumingType.Kind == yang.Yunion {
+		// We specifically say that this is an enumeration within the leaf.
+		if noUnderscores {
+			typeName = fmt.Sprintf("%sEnum", subsumingType.Name)
```
robshakir, line 256: Suggest to make this a string const somewhere.

ygen/enumgen.go
```diff
@@ -229,47 +238,41 @@ func (s *enumSet) identityBaseKey(i *yang.Identity) string {
...
+			typeName = fmt.Sprintf("%s_Enum", subsumingType.Name)
 		}
 	}
-	if e.Node == nil {
-		return "", "", nil, fmt.Errorf("nil Node in enum type %s", e.Name)
+	if args.contextEntry.Node == nil {
+		return "", "", fmt.Errorf("nil Node in enum type %s", args.contextEntry.Name)
 	}
 
 	// This is the old, buggy behaviour that causes same-named typedefs to shadow one another.
```
robshakir, line 265: Can we add a TODO and link to the issue to remove this code once we've transitioned consumers?

ygen/enumgen.go
```diff
@@ -744,23 +747,36 @@ func (s *enumGenState) resolveNameClashSet(nameClashSets map[string]map[string]*
...
 		switch {
 		case t.IdentityBase != nil:
 			if err := s.resolveIdentityrefBaseTypeFromIdentity(t.IdentityBase, noUnderscores); err != nil {
 				return err
 			}
 		case t.Enum != nil:
-			if util.IsYANGBaseType(t) {
+			enumNameSake := t
+			if useDefiningModuleForTypedefEnumNames {
+				enumNameSake = subsumingTypes[i]
```
robshakir, line 774: To be clear here - `enumNameSake` is a type that is going to be used to give this type its name, because enumerations themselves are anonymous? I think it'd be useful to make a comment here that says something of this nature, just so that we remember this is how it works.

ygen/enumgen.go
```diff
@@ -744,23 +747,36 @@ func (s *enumGenState) resolveNameClashSet(nameClashSets map[string]map[string]*
...
+// entry. The name of each enumerated value is calculated based on the original
+// context: whether path compression is enabled, whether the name should not
+// include underscores, whether the enum name should be de-deduped if the same
+// enumeration is already been seen in a different part of the YANG tree,
+// whether a shortened version of the enumeration names should be given for
+// enumeration leaves, and whether the defining module name (as opposed to the
+// user module's name) should be used to name enumerated typedefs.
 func (s *enumGenState) resolveEnumeratedUnionEntry(e *yang.Entry, compressPaths, noUnderscores, skipEnumDedup, shortenEnumLeafNames, useDefiningModuleForTypedefEnumNames bool) error {
-	for _, t := range util.EnumeratedUnionTypes(e.Type.Type) {
+	enumTypes := util.EnumeratedUnionTypes(e.Type.Type)
```
robshakir, line 760: I think I have the same question here as elsewhere, why is an entry at index `i` of `enumTypes` and `subsumingTypes` equivalent?

ygen/enumgen.go
```diff
@@ -882,9 +898,9 @@ func (s *enumGenState) resolveTypedefEnumeratedName(e *yang.Entry, noUnderscores
 	// as this is unique, whereas post-camelisation, we may have name clashes. Since a typedef
 	// does not have a 'path' in Goyang, we synthesise one using the form
 	// defining-module-name/typedef-name.
-	nodeForModuleName := e.Node
+	nodeForModuleName := args.contextEntry.Node
 	if useDefiningModuleForTypedefEnumNames {
-		nodeForModuleName = enumType.Base
+		nodeForModuleName = args.subsumingType.Base
```
robshakir, line 903: Can `args.subsumingType` ever be `nil` here?

ygen/goelements.go
```diff
@@ -425,17 +436,23 @@ func (s *goGenState) goUnionType(args resolveTypeArgs, compressOCPaths, skipEnum
 // The skipEnumDedup argument specifies whether the current code generation is
 // de-duplicating enumerations where they are used in more than one place in
 // the schema.
-func (s *goGenState) goUnionSubTypes(subtype *yang.YangType, ctx *yang.Entry, currentTypes map[string]int, unionMappedTypes map[int]*MappedType, compressOCPaths, skipEnumDedup, shortenEnumLeafNames, useDefiningModuleForTypedefEnumNames bool) []error {
+func (s *goGenState) goUnionSubTypes(subtype *yang.YangType, ctx *yang.Entry, subsumingType *yang.YangType, currentTypes map[string]int, unionMappedTypes map[int]*MappedType, compressOCPaths, skipEnumDedup, shortenEnumLeafNames, useDefiningModuleForTypedefEnumNames bool) []error {
```
robshakir, line 439: Should we just make this take `resolveTypeArgs` now, since this is all three fields from that struct in use here.

docs/design.md
wenovus: Thanks, this is much cleaner

ygen/codegen_test.go
wenovus: Thanks! That's helpful, very useful for debugging enumgen_test.go tests.

ygen/enumgen.go
wenovus: Now that `DefiningType` is being used, hopefully this looks much cleaner.

ygen/enumgen.go
wenovus: PTAL with `DefiningType`.

ygen/enumgen.go
wenovus: Added `enumeratedUnionSuffix` within enumgen.go

ygen/enumgen.go
wenovus: reverted now that `DefiningType` is being used.

ygen/goelements.go
wenovus: Reverted.

ygen/enumgen.go
wenovus: Added error handling.

ygen/enumgen.go
```diff
@@ -25,6 +25,10 @@ import (
 	"github.com/openconfig/ygot/util"
 )
 
+const (
+	enumeratedUnionSuffix = "Enum"
+)
```
wenovus, line 30: Need to add comment for this.

ygen/enumgen.go
```diff
@@ -229,47 +247,40 @@ func (s *enumSet) identityBaseKey(i *yang.Identity) string {
...
-		case 0:
-			return "", "", nil, fmt.Errorf("enumerated type had an empty union within it, path: %v, type: %v, enumerated: %v", e.Path(), e.Type, enumTypes)
-		default:
-			return "", "", nil, fmt.Errorf("multiple enumerated types within a single enumeration not supported, path: %v, type: %v, enumerated: %v", e.Path(), e.Type, enumTypes)
+	if definingType.Kind == yang.Yunion {
+		// We specifically say that this is an enumeration within the leaf.
+		if noUnderscores {
+			typeName = fmt.Sprintf("%s%s", definingType.Name, enumeratedUnionSuffix)
+		} else {
+			typeName = fmt.Sprintf("%s_%s", definingType.Name, enumeratedUnionSuffix)
```
wenovus, line 265: Does the old behaviour still work the same with `useDefiningModuleForTypedefEnumNames` set to false?

wenovus: Yes. When `useDefiningModuleForTypedefEnumNames` is false, `enumeratedTypedefKey()` is only ever called with the entry being a typedef, meaning the `DefiningType` of the enum entry is always itself, producing the same behaviour.

As for the `len(enumTypes)` part of the old code, this new code makes the decision that *multiple enums within a union is now allowed*. This is a deviation from before, but it's a deviation that's in the right direction and backwards-compatible.

wenovus: There is a minor backwards-compatibility issue with when to append `_Enum` that's now fixed.

PR #419, opened 2020-07-24
ygen refactor design

robshakir:
This PR is specifically for a design document that covers the redesign of ygen to add an intermediate representation for code generation.

```
   Design for ygen intermediate representation.

    * (M) docs/code-generation-design.md
    - This PR adds a document describing a refactoring of the ygen
      library to add an intermediate representation to the generated
      code which can be used across library versions. It addresses
      a number of the shortcomings of the current design and has been
      prototyped in the 'restruct-ygen-interface' branch of the repo.
      This document will be checked-in to make the design POR before
      proceeding with integrating this prototype code into the
      production release.
```


docs/code-generation-design.md
```diff
@@ -0,0 +1,251 @@
...
+	ModelData     []*gpb.ModelData
+}
+```
+
+#### Definitions for Directory / Leaf Nodes
+
+The proposed IR for the directory types in the returned code is as follows. The base philosophy of whether fields are included within this representation is to keep to the set of fields made available to the code output library to their minimum, to ensure that there is clear encapsulation of the complexities of the YANG hierarchy within `ygen`.
+
+```golang
+// ParsedDirectory describes a node that is the root of a subtree
```
wenovus, line 127: nit: "a node that is the root of a subtree" -> "an internal node"

docs/code-generation-design.md
```diff
@@ -0,0 +1,251 @@
...
+
+YANG has a number of different types of enumerated values - particularly, `identity` and `enumeration` statements. Generated code may choose to treat these differently. For example, for an `enumeration` leaf - some languages may choose to use an embedded enumeration (e.g., protobuf can use a scoped `Enum` within a message). For this reason, some of the provenance of a particular value needs to be exposed to the downstream code generation libraries.
+
+Despite this, a significant amount of pre-parsing can be done by the `ygen` library to pre-process enumerated values prior to being output to code. Particularly:
+
+* Handling of an enumerated value's integer ID to an language-specific name can be created (using the relevant features of the `LangMapper` interface).
+* A reverse mapping between the integer ID and defining module and corresponding YANG identifier can be created.
+
+By performing this pre-processing, the functionality of the code generation library is constrained to determine how each type of enumeration should be output, and creating the language-specific constructs required for mapping (e.g., the enumeration map created in `ygen`'s current Go output).
```
wenovus, line 206: "constrained to determine" -> "constrained to determining"

docs/code-generation-design.md
```diff
@@ -0,0 +1,251 @@
...
+The IR format for enumerations is defined to be as follows:
+
+```golang
+type EnumeratedValueType int64
+
+const (
+	_ EnumeratedValueType = iota
+	SimpleEnumerationType
+	DerivedEnumerationType
+	UnionEnumerationType
```
wenovus, line 217: With the new enum changes, is there now a difference between an enum within a typedef union vs. a not within a typedef union that's worth putting here?

Reminder that the new enum changes effectively moved a portion of enums previously considered to be simple enumerations used within a leaf's union type, to typedef enums that belong to the typedef union under which they're defined. The only difference is that since they're not regular typedef enums, they're distinguished with a `_Enum` suffix.

docs/code-generation-design.md
```diff
@@ -0,0 +1,251 @@
...
+
+Particularly, we intend to restructure ygen to enforce a strict multi-stage code generation process. The target end-to-end pipeline is targeted to be:
+
+ * An input set of YANG schemas is to be parsed by `goyang` and resolved through `yang.Node` representations into a set of `yang.Entry` structs. The set of `yang.Entry` structures that exist for a schema will be implemented to be lossless, such that there is no information that is available in the input YANG schema that cannot be extracted from them, and the original format of the structure is maintained.
```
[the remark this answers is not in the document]
robshakir: Yep, makes sense to me. Done.

docs/code-generation-design.md
```diff
@@ -0,0 +1,251 @@
...
+
+This structure will have the following benefits:
+
+* Knowledge of schema transformations will be clearly encapsulated into the `ygen` library -- today, numerous methods that are outputting code must be aware of the different transformations that are being applied to the schema - resulting in complex state tracking being required through the entire generation process. This causes significant additional effort to understand all the hooks required in the code generation libraries to add new output formats.
+* Generation of names for output code entities is moved from being a just-in-time process, which requires careful consideration of order, or can potentially have non-deterministic output - to being an up-front process. Previous changes have shown the benefits of moving enumeration naming to an up-front process where there can be clear understanding of how names are to be resolved, ensuring that there is a strictly defined IR ensures that this pattern can be applied across the code base.
```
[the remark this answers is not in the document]
robshakir: True :-), fixed.

docs/code-generation-design.md
robshakir: Fixed, thanks.

docs/code-generation-design.md
robshakir: Yes, I think that we probably do need to have different definitions here, an enumeration in a typedef union is going to be a global-scoped type in the generated code, whereas an enumeration that is within a non-typedef union can still be locally scoped.

I've added these in.

Issue #418, opened 2020-07-23
MergeStructInto does not support overwrite


soumiksamanta:
If the dstField is set, ideally MergeStructInto should overwrite the value from src

```
if !util.IsNilOrInvalidValue(dstField) {
	s, d := srcField.Elem().Interface(), dstField.Elem().Interface()
	if diff := cmp.Diff(s, d); diff != "" {
		return fmt.Errorf("destination value was set, but was not equal to source value when merging ptr field, (-src, +dst):\n%s", diff)
	}
}
```

There are usecases when one want to merge the dst struct from src struct, adding and updating the fields. 

Is this outside the scope of the `MergeStructInto`?

wenovus:
This doesn't seem to be supported, although it can definitely be implemented. @robshakir are you ok with adding a helper here?

For now, a workaround is perhaps to call [ygot.TogNMINotifications](https://pkg.go.dev/github.com/openconfig/ygot/ygot?tab=doc) on the srcStruct, and then looping over the `Update` field, calling [ytypes.SetNode](https://pkg.go.dev/github.com/openconfig/ygot/ytypes?tab=doc) for each `Path` and `Val` for each `Update` on the destStruct. I admit this may not work and is not ideal, but it may serve what you need.

robshakir:
Hi!

This was the requested behaviour of the consumer of this API that we implemented this call for initially -- in their use case, there was an explicit need for an error when two fields were set to unequal values. 

I suggest that the right way to add this is to add a `MergeOpt` (in a similar manner to the `DiffOpt` that we have today) that gives us a way to control whether overwriting is performed - and then plumb this through the code such that we check it before doing the diff quoted in the OP.

We should ensure to add unit tests for this case too. 

@soumiksamanta -- Wen or I can review a submission on this if you'd like to make a PR. We can also discuss the design further here if you'd like.

Cheers,
r.

soumiksamanta:
Thanks @robshakir @wenovus I will run through the current behavior and try to introduce the `MergeOpt`. I will definitely get back to you if I hit a roadblock.

PR #420, opened 2020-07-28
Mergestructinto mergeopt

soumiksamanta:
@robshakir @wenovus 
This is the initial commit for #418. I have added test cases for the new MergeOpt.
However, I am not sure of any corner cases which should be added for the tests.


ygot/struct_validation_map.go
```diff
@@ -475,14 +475,33 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
 	return o, nil
 }
 
+// MergeOpt is an interface that is implemented by the options to the MergeStructInto
```
robshakir, line 478: Suggest to change this to say "options to the MergeStructs and MergeStructInto functions", since this is now propagated to both implementations.

ygot/struct_validation_map.go
```diff
@@ -475,14 +475,33 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+// function. It allows user specified options to be propagated to the merge
+// method.
+type MergeOpt interface {
+	// IsMergeOpt is a marker method for each MergeOpt.
+	IsMergeOpt()
+}
+
+// DiffPathOpt is a DiffOpt that allows control of the path behaviour of the
```
robshakir, line 486: Update comment from `DiffPathOpt` to cover the fact that this is actually about the `MergeOpt`.

ygot/struct_validation_map.go
```diff
@@ -475,14 +475,33 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+// Diff function.
+type MergeStructOpt struct {
+	// Overwrite specifies whether the dst struct can be overwritten with existing
```
robshakir, line 489: Can we make this a bit clearer? I think we perhaps need to say:

> When overwrite is set to true, fields that are populated in the destination struct will be overwritten
> by values that are populated in the source struct. If the field is unpopulated in the source struct,
> the value in the destination struct will not be modified. When overwrite is set to false, an error
> will be produced if a field is set in both the source and destination structs and their values are
> not equal.

ygot/struct_validation_map.go
```diff
@@ -527,8 +546,26 @@ func DeepCopy(s GoStruct) (GoStruct, error) {
 	return n.Interface().(GoStruct), nil
 }
 
+// hasMergeStructOpt extracts a MergeStructOpt from the opts slice provided. In
```
robshakir, line 549: Since this is only used in a single place - I'd suggest having a single helper function that specifies `fieldOverwriteEnabled` and returns the `bool` value itself (returning false if this function would have returned `nil`).

ygot/struct_validation_map.go
```diff
@@ -475,14 +475,33 @@ func MergeJSON(a, b map[string]interface{}) (map[string]interface{}, error) {
...
+// function. It allows user specified options to be propagated to the merge
+// method.
+type MergeOpt interface {
+	// IsMergeOpt is a marker method for each MergeOpt.
+	IsMergeOpt()
+}
+
+// DiffPathOpt is a DiffOpt that allows control of the path behaviour of the
+// Diff function.
+type MergeStructOpt struct {
```
robshakir, line 488: I'd suggest that we make this a more specific option - since right now I can't see how there would be >1 different option -- it might be better to simply make this be:

```golang
type MergeOverwriteExistingFields struct { }
```

and then simply use the presence of this option as indicating that overwriting should be enabled.

ygot/struct_validation_map_test.go
```diff
@@ -1831,6 +1933,7 @@ func TestMergeStructs(t *testing.T) {
 		name    string
 		inA     ValidatedGoStruct
 		inB     ValidatedGoStruct
+		inOpts  []MergeOpt
 		want    ValidatedGoStruct
 		wantErr string
 	}{
```
robshakir, line 1939: Can we add one test case here that exercises the merge options -- it doesn't need to be exhaustive, but it would be useful to ensure that the implementation passes this through correctly going forward.

ygot/struct_validation_map_test.go
```diff
@@ -1822,6 +1908,22 @@ var mergeStructTests = []struct {
...
+		UnionField: &copyUnionI{42},
+	},
+	inOpts: []MergeOpt{
+		&MergeStructOpt{
+			Overwrite: true,
+		},
+	},
+	want: &validatedMergeTest{
+		UnionField: &copyUnionI{42},
+	},
```
robshakir, line 1926: It looks like we're missing a test case for a struct field that is overwritten - can we add one please?

ygot/struct_validation_map_test.go
soumiksamanta: `TestMergeStructs` already uses the common test cases defined in `mergeStructTests` which has test cases with merge option.

ygot/struct_validation_map.go
```diff
@@ -565,7 +597,11 @@ func copyStruct(dstVal, srcVal reflect.Value) error {
 			vSrc, vDst := srcField.Int(), dstField.Int()
 			switch {
 			case vSrc != 0 && vDst != 0 && vSrc != vDst:
-				return fmt.Errorf("destination and source values were set when merging enum field, dst: %d, src: %d", vSrc, vDst)
+				if !fieldOverwriteEnabled(opts) {
+					return fmt.Errorf("destination and source values were set when merging enum field, dst: %d, src: %d", vSrc, vDst)
+				} else {
```
robshakir, line 602: You don't need the `else` here, since the if the clause above is executed then an error is returned, so we can just have this as the `dstField.Set` outside of the `else`.

ygot/struct_validation_map_test.go
```diff
@@ -1632,6 +1683,14 @@ type validatedMergeTestSliceField struct {
 	String *string
 }
 
+type validatedMergeTestWithStructField struct {
+	StructField reflect.StructField
```
robshakir, line 1687: This is a bit strange as a test case -- `reflect.StructField` is usually formed by doing `reflect.ValueOf` something -- `ygen` can never produce this form of struct. Could you explain which of the features that you were trying to cover here please?

soumiksamanta: In your earlier comment, you had asked me add testcases for StructField.

robshakir: Sorry -- I meant something different - i.e., a field of a struct, not a `reflect.StructField`. Essentially, I was asking whether we have a test whereby we have:

```
src := ExampleStruct{S: "hello"}
dst := ExampleStruct{S: "world"}
```

and we check that the result is that the `S` string field is overwritten.

Apologies for the misunderstanding.

ygot/struct_validation_map.go
```diff
@@ -527,8 +545,22 @@ func DeepCopy(s GoStruct) (GoStruct, error) {
 	return n.Interface().(GoStruct), nil
 }
 
+// fieldOverwriteEnabled looks for the presence of MergeOverwriteExistingFields
+// from the MergeOpt slice provided.
+// In the case that there are multiple MergeOpt structs within the slice,
```
robshakir, line 550: (nit) This sentence seems to be irrelevant now (since we just return a `bool`) can you remove it?

ygot/struct_validation_map_test.go
```diff
@@ -1632,6 +1696,14 @@ type validatedMergeTestSliceField struct {
 	String *string
 }
 
+type validatedMergeTestWithStructField struct {
```
robshakir, line 1699: Can we remove these definitions and the test cases that use this please?

soumiksamanta:
@robshakir Can we create a release? Need to use it in our project.
//Thanks

PR #422, opened 2020-07-31
Allow gNMI paths with union keys embedded to be unmarshalled

wenovus:
Currently, union key values cannot be unmarshalled due to `StringToType`
not having a case statement handling the case when the string is to be
converted to a union type, causing an unmarshal error.

e.g. `a/b[unionKey=1234]/c` would fail as `1234` cannot be converted to unionKey's type. A schema_test has been added to illustrate this.

This change introduces handling for union types, which involves the
creation of new functions `stringToKeyType` and `stringToUnionType`
due to the extra complexity in handling unions -- that being the final concrete
type is not known.

The new handling incorporates plenty of existing
logic, so some refactoring is done to reduce function size and reuse
code segments.


ytypes/util_types.go
```diff
@@ -134,6 +210,134 @@ func StringToType(t reflect.Type, s string) (reflect.Value, error) {
...
 }
 
+// stringToKeyType converts the given string to the type specified by the schema.
+// This is in contrast to StringToType, which requires foreknowledge of the
+// concrete type of the value. This is especially useful for converting the
+// string to a union type, where the final concrete value is not known.
+func stringToKeyType(schema *yang.Entry, parent interface{}, fieldName string, value string) (reflect.Value, error) {
+	ykind := schema.Type.Kind
+	switch ykind {
+	// TODO(wenbli): case yang.Yempty: case yang.Ybits: case yang.Ybool:
```
robshakir, line 220: `empty` cannot be a key type - so this doesn't need to be handled.

Should we add handling for `bool` here? (It's probably rare, but it'd be useful to handle and seems like it should be do-able.)

ytypes/util_types.go
```diff
@@ -134,6 +210,134 @@ func StringToType(t reflect.Type, s string) (reflect.Value, error) {
...
+
+// stringToUnionType converts a string value into a suitable union type
+// determined by where it is located in the YANG tree.
+func stringToUnionType(schema *yang.Entry, parent interface{}, fieldName string, value string) (reflect.Value, error) {
+	util.DbgPrint("stringToUnionType value %v, into parent type %T field name %s, schema name %s", util.ValueStrDebug(value), parent, fieldName, schema.Name)
+	if !util.IsTypeStructPtr(reflect.TypeOf(parent)) {
+		return reflect.ValueOf(nil), fmt.Errorf("stringToKeyType: %T is not a struct ptr", parent)
+	}
+	parentT := reflect.TypeOf(parent)
+	dft, _ := parentT.Elem().FieldByName(fieldName)
```
robshakir, line 279: Do we need to handle the case here that this returns `false`? If I call this function with an invalid name then this seems like it might NPE at the next line?

ytypes/util_types.go
wenovus: I'll add the boolean representation in a follow-up PR. `KeyValueAsString` currently doesn't support it either.

wenovus: follow-up PR has been merged.

PR #424, opened 2020-08-04
Transform leafref paths when prefer_operational_state is set.

wenovus:
Currently, the AST returned by goyang to ygot is never touched, including leafrefs that point to leaves under config.

This change modifies the AST's leafrefs to point to the state versions of the leaves when `prefer_operational_state` is set.

**Reason for Change**:
When `prefer_operational_state` is set, ygot effectively prunes away the config portions of the YANG tree in the IR as well as the generated code representation. This means that whenever the schema is used for any processing in conjunction with either the IR or the generated code, the config path could be used even though the node doesn't exist.

One problem this causes is that the mismatch with IR causes the leafref path to be missing in the path tag in the generated Go code. The result is leafrefs pointing to config leaves in in YANG cannot be unmarshalled.


genutil/common.go
```diff
@@ -454,3 +455,47 @@ func addNewChild(m map[string]*yang.Entry, k string, v *yang.Entry, errs []error
...
+// TransformEntry makes changes to the given AST subtree returned by goyang
+// depending on the compress behaviour.
+// Currently, only PreferOperationalState entails a transformation, where
+// leafrefs pointing to config leaves are changed to point to state leaves.
+func TransformEntry(e *yang.Entry, compressBehaviour CompressBehaviour) util.Errors {
+	if compressBehaviour != PreferOperationalState {
+		return nil
+	}
+
+	pointLeafrefToState := func(e *yang.Entry) error {
```
robshakir, line 468: It might be worth adding a comment here indicating that in OpenConfig this is /always/ safe to do - because there must be a 'state' leaf for each 'config' leaf, and this is guaranteed by the linter. In other schemas this wouldn't necessarily be the case.

genutil/common_test.go
```diff
@@ -651,3 +652,254 @@ func TestFindChildren(t *testing.T) {
...
+				Kind:   yang.LeafEntry,
+				Config: yang.TSTrue,
+				Type: &yang.YangType{
+					Kind: yang.Yleafref,
+					Path: "../config/name", // To be modified within test.
+				},
+			},
+		},
+	}
+	wantEntry = &yang.Entry{
```
robshakir, line 818: It might be best to call this something different than `wantEntry` -- it wasn't immediately clear to me that we were specifying `wantEntry` within this new `wantEntry` maybe the one above can just be called `childEntry`?

robshakir: I'd maybe even just call this `schemaTemplate` - just because we mutate it in the test code.

PR #429, opened 2020-08-12
Add flag ignore_shadowed_schema_paths

wenovus:
This flag keeps the shadowed fields (i.e. state leaves for `PreferIntendedConfig`, and config leaves for `PreferOperationalState`) in a new `shadow-path` tag for Go code. These compressed out config/state leaves can now unmarshal into the generated `GoStruct`s without an error.

Since these paths are compressed out, however, unmarshal doesn't actually apply the data when unmarshal is called. It just silently allows the call without erroring out.

LIMITATION: JSON unmarshalling does not given an error if the supplied shadow path after the shadow element is invalid.

This flag is only relevant when compressPaths=true.


robshakir:
Wen - can we have a quick design discussion here please?

It seems like here you're trying to allow `/foo/bar/config/baz` to unmarshal when we're doing a transformation that removes the `config` leaves from the schema. What is the intention of doing this? Is it to avoid errors being thrown to the user when they try and do the `unmarshal`? It seems like this has to be the intention, since we do not have a means for a user to access this after the fact.

There are different design questions that I think that we need to consider.

1. Why is the user using a compressed schema when data that is not compatible with compression is being accessed in their system?
2. What is the required access to the data that is in the 'compressed out' part of the schema?
3. Why do we need to carry additional information in the IR which breaks the abstraction that we've established (all schema transformations having been completed in the ygen phase) to meet these requirements? It seems like `ShadowLeaves` now becomes something that gets us back to the complexity of the previous representation.

I'm happy to discuss these in-realtime, but I'm not particularly comfortable with this change before we discuss this further.

wenovus:
Good questions! I did not check the future IR -- using `MapPaths` is much better than adding some shadow yang.Entries.

That being said, I think these shadow entries' paths might need to be stored separatedly in, say `ShadowMapPaths`, in order to allow the user to choose whether to allow the "compressed out" parts of the schema to unmarshal or marshal.

My response to the design questions:
1. I'm guessing a user may still receive these paths, but still chooses compressed schemas for convenience, and when both are perceived to be the same?
2. The required access would be to marshal and unmarshal.
3. As above, I agree using `MapPaths` is cleaner and abides by the IR.

robshakir:
1. I can think that in the case that we have `state` only generated code, then the use-case I can see is that the user is still getting `config` leaves back. This seems worth solving somehow -- i.e., this should not be an error _if_ they are valid schema paths.
2. I'm not sure that marshalling makes sense here -- since it means that we'd have to have data to populate these. We'd need to decide whether a particular "shadow" path should be marshalled to in all cases (should state == config in all cases?) -- this decision seems like it might be implementation specific.

   The question I'd have for unmarshalling is whether the user really just wants to receive these leaves without errors, or whether they want to consider `config` and `state` equivalent, such that getting `state/x` and `config/x` are considered absolutely equivalent? If not, can we just add something that says "if we receive a leaf that is valid in the schema - i.e., the path exists - but was compressed out of the created structs, don't error?". Does that meet the use case that we have here?

wenovus:
I like this proposal -- i.e. add a code-generation flag that toggles whether to ignore config leaves, as well as leafref updates that point to config leaves, when unmarshalling, without erroring out.

If marshalling them to the same values doesn't make sense, then instead of doing `path:"state/x|config/x|x"`, we might do `path:"state/x"`, shadow-paths:"config/x|x"` which allows for different handling by the marshal and unmarshal functionalities.

generator/generator.go
```diff
@@ -61,13 +61,15 @@ var (
...
+	excludeModules         = flag.String("exclude_modules", "", "Comma separated set of module names that should be excluded from code generation this can be used to ensure overlapping namespaces can be ignored.")
+	packageName            = flag.String("package_name", "ocstructs", "The name of the Go package that should be generated.")
+	ignoreCircDeps         = flag.Bool("ignore_circdeps", false, "If set to true, circular dependencies between submodules are ignored.")
+	fakeRootName           = flag.String("fakeroot_name", "", "The name of the fake root entity.")
+	skipEnumDedup          = flag.Bool("skip_enum_deduplication", false, "If set to true, all leaves of type enumeration will have a unique enum output for them, rather than sharing a common type (default behaviour).")
+	preferOperationalState = flag.Bool("prefer_operational_state", false, "If set to true, state (config false) fields in the YANG schema are preferred over intended config leaves in the generated Go code with compressed schema paths. This flag is only valid for compress_paths=true and exclude_state=false.")
+	// TODO(wenbli): Invalid shadow paths currently do not cause an error during JSON unmarshalling.
```
robshakir, line 71: This might now be solved by #468 -- but we can validate this later. Can you file a bug here? (The same caveat was true for non-shadowed paths.)

generator/generator.go
```diff
@@ -61,13 +61,15 @@ var (
...
+	ignoreShadowSchemaPaths              = flag.Bool("ignore_shadow_schema_paths", false, "If set to true when compress_paths=true, the shadowed schema path will be ignored while unmarshalling instead of causing an error. NOTE: Invalid shadow paths currently do not cause an error during JSON unmarshalling.")
```
robshakir, line 72: Should we explain what a 'shadow' path is here (or link to somewhere that does)?

robshakir (approved):

generator/generator.go
wenovus: Done.

generator/generator.go
wenovus: Added explanation to flag description.

PR #434, opened 2020-08-17
Add oc-ext:posix-pattern Support

wenovus:
- Use `yang.YangType.POSIXPattern` instead of `yang.YangType.Pattern` whenever the former exists.
- `CompilePOSIX` is used for `oc-ext:posix-pattern`, and `Compile` is used for `pattern` for backwards compatibility. (e.g. inet-ipv4address fails to compile with `CompilePOSIX` due to `\p` -- it seems there is some of the same character class support in the regular `regexp.Compile` function).
- fix corner case bug in `fixYangRegexp` and move to util package.


ytypes/string_type.go
```diff
@@ -105,69 +110,18 @@ func validateStringSchema(schema *yang.Entry) error {
...
-
-		buf.WriteRune(ch)
-
-		if i == len(pattern)-1 {
-			if addParens {
-				buf.WriteRune(')')
-			}
-			if ch != '$' {
-				buf.WriteRune('$')
-			}
```
wenovus, line 166: @robshakir The edge case being fixed comes from here:
`abc$` -> `^(abc$)` due to `buf.WriteRune(ch)` being before `buf.WriteRune(')')`

The fix is `^(abc)$`. While `^(abc\$)$` is also possible, the `case '$'` block seems to suggest that the former is the intended behaviour.

wenovus:
The OC models have already moved over actually. I'll add some tests.

wenovus:
Oh I see, we haven't yet updated the public models repository.

How should the integration test be done when both regexes exist?

The regex integration test fix I made is on a test which was passing only because a different error (i.e. key mismatch) was occurring, and not because of a regex error.

wenovus:
@robshakir You can disregard my previous comments.
Is this good to submit given that now,
- There is a working integration test on a regex here: ytypes/schema_tests/validate_test.go
- I have added a new unit test for both POSIX and non-POSIX that fails due to string going past anchors.
- The edge case has been explained.

wenovus:
Since I believe the comments have been addressed and approval has already been given, I will submit this.

Issue #172, opened 2018-03-13
Default value of a enum is not used as the UNSET value in Go


robshakir:
In the protobuf generation code, the default value of an enumeration is read and this is used as the `UNSET` value - such that the default value of the enum is the default. This same logic is not implemented in the Go code - but should be.

wenovus:
#247 is an interesting case that this should handle. Maybe just generating an initialization routine that sets `default` values on enum fields within `GetOrCreate*` and `New*` would be sufficient, instead of having to know whether all enum usages use the same default value and then removing the UNSET value only for those enums.

If we do this, however, what would be the justification for not extending the initialization routine to cover all the other fields?

If we don't do initialization for other types according to their `default` values, then why do it for enums only?

wenovus:
Extending it to all other values is talked about in #101

wenovus:
So I think that #101 is the right discussion for this question and should be resolved by its resolution. I think we shouldn't look at only providing default support for enums, because it would be incongruent support between all the types involved, and I don't think enums are that special.

Issue #101, opened 2017-10-19
Mechanism to access default values


zhiweicen:
It is common for a leaf node to have a default value ([here](https://github.com/openconfig/public/blob/master/release/models/wifi/phy/openconfig-wifi-phy.yang#L135) is an example). If those nodes are unset, the current behavior is that the node is marked as unset (nil). This is not ideal as the Go code lost the information of the default value. I can think of two possible solutions to this.

1) Provide an additional API for each node to access its default value (nil if there is no default value). Or
2) Set the field to have the value of the default even if the field is unset.

Option 2) has the drawback of losing the fact when a field is purposefully unset but either implementation would be a good step forward compared with the current behavior.

robshakir:
For option 2, do you mean that this is set when accessing the field, as well as when marshalling? IMHO, this gives us difficult user semantics. I would prefer some additional API. This could be a defaults method with each type as a receiver, it'd return a copy of the receiver type with the defaults populated if they existed.

zhiweicen:
For option 2 I was only thinking about accessing the field, but not putting it in when marshalling may create other problems (e.g. when the JSON file is to be consumed by other parsers). I agree that Option 1) is a better choice.

PS the github renderer showed my comment about option 2) as another option (number three). I just updated the original post to avoid this.

robshakir:
OK - so I propose we implement this per the following:

 * Implement a `ΛDefaults` method on each struct (continuing the naming based on characters that are not legal in YANG). This would return the type with the defaults populated if they were set. 
 * There is a little work to do here since goyang's `Entry` struct populates this as a `string` (see [here](https://github.com/openconfig/goyang/blob/master/pkg/yang/entry.go#L73)). We need to convert the string to the relevant schema type to return it.
 * `EmitJSON`/`ConstructJSON` would have a new flag to specify whether defaults should be included, in this case, it'd call the `ΛDefaults` method and (when the value is `nil`) use the default.

ostromart:
How about Unmarshal? Seems like we'd want defaults there too.

robshakir:
I don't think that Unmarshal has to handle this -- because if we're unmarshalling then we might want to know that explicitly the input data did not set this field - this would mean that we leave the field as `nil` in the case that the leaf is not present; explicitly set it to the default if it is set (as per today's implementation); and leave the handler of the struct to retrieve the default in the unset case.

Does that make sense?

KBA07:
[...]

PR #436, opened 2020-08-26
Make UNSET enums' String() return "UNSET"

wenovus:
Returning this message for UNSET seems rather excessive, as reflected by a user: `fmt.Sprintf("out-of-range %s enum value: %v", enumTypeName, val)`

Since all UNSET values are currently 0, we can just use this fact to return "UNSET" if they're not found within the enum map.


wenovus:
Actually, I think UNSET might be better, since it might be used for logging. Seeing `UNSET` is probably better than nothing.

robshakir:
There was the intention to make the default value of an enumeration be the 0 value in the case that a default was specified for the leaf -- is that currently not-operational? If so, it would mean that `UNSET` as a string literal might not be the right thing to return.

wenovus:
Since `TogNMINotifications` doesn't use the schema to determine the value to marshal, doesn't it mean that this is not operational?
https://github.com/openconfig/ygot/blob/c6ce7e55aa79fe2cdd2f5c655b882b9c98f71fd3/ygot/render.go#L287

If we have a typedef enum used in multiple places, then in order to stream the correct default value, `TogNMINotifications` necessarily has to get this from the schema metadata.

wenovus:
#172 confirms that this is not in operation in Go. It is also of dubious operation in proto due to the possibility of re-uses of enum typedefs. I think the question of how we're going to support default values in Go needs to be answered first before the concern raised in this PR can be answered.

wenovus:
Just clarifying here that this PR is dependent on how #101 and specifically #172 are resolved.

robshakir:
🧹: Closing stale PRs. Please re-open if you're interested in contributing this code upstream in the future.

PR #439, opened 2020-09-02
Remove template maps; Use template vars directly

wenovus:
The style now matches that of `ypathgen`.

The advantage of this style is that it makes finding the usages templates slightly easier, and better avoids typo errors.


wenovus:
Removed `protoListKeyTemplate`. It looks pretty complicated -- is this a TODO for something that we should keep in a comment section?

robshakir:
No, I think it was actually a stale template from some design changes in the code -- today, the key message for a list is generated as a child message of the message itself, and there are test cases that cover this. Given that CI passes, and this was unreferenced, I think we're safe to blow it away.

Thanks!
r.

robshakir (approved):

PR #440, opened 2020-09-04
Custom validations

soumiksamanta:
Adding custom validation to container. Though it's not what we want in the longer run. We need to handle the custom validation in a more generic way.


robshakir:
I think this version is a bit more confusing than what we have today - I suspect what we need is to have:

```
type CustomValidator struct {
  T reflect.Type
   Fn func(interface{}) error
}
```

such that we can load in each case - then each `Validate` function should check the type against a `map[reflect.Type]func(interface{}) error` and execute that function if we've been asked to.

This approach seems to make the function half-way generic -- it's generic, but just for containers.

PR #442, opened 2020-09-11
Add support for new Go unions for ytypes' Validate functions

wenovus:
Added testutil/types.go for unit-testing new Go union concrete types. These are the same as what would be generated by ygen in a new flag for generating these simpler unions.

Validate functionality is modified to be able to validate either a code on the old union API or the new union API. I use the adjective "simple" to describe the new unions in tests, and added `(wrapper union)` to the descriptors of the tests for the old union types. This helps remove this functionality when we elevate the flag to be the normal behaviour.

This is the first of a series of PRs that add support and unit testing to the ytypes/ygot packages. The new code generation will be added last, after the support code (ytypes/ygot) is ready.

### DIFF
--- a/testutil/types.go
+++ b/testutil/types.go
@@ -73,11 +73,11 @@ type TestUnion interface {
 	IsTestUnion()
 }
 
-func (String) IsTestUnion()  {}
-func (Int16) IsTestUnion()   {}
-func (Int64) IsTestUnion()   {}
-func (*Binary) IsTestUnion() {}
+func (String) IsTestUnion() {}
+func (Int16) IsTestUnion()  {}
+func (Int64) IsTestUnion()  {}
+func (Binary) IsTestUnion() {}
 
-func (String) IsUnion()  {}
-func (Int64) IsUnion()   {}
-func (*Binary) IsUnion() {}
+func (String) IsUnion() {}
+func (Int64) IsUnion()  {}
+func (Binary) IsUnion() {}
--- a/testutil/types.go
+++ b/testutil/types.go
@@ -64,6 +64,11 @@ type Unsupported struct {
 	Value interface{}
 }
 
+// Union1 is an interface defined within *this* (testutil) package that is
+// satisfied by a subset of the above union types to aid testing within other
+// packages.
+// Enumerations defined within other test packages can still satisfy this
+// interface by defining an IsUnion1() method.
 type Union1 interface {
 	IsUnion1()
 }
--- a/testutil/types.go
+++ b/testutil/types.go
@@ -1,0 +1,78 @@
+// Copyright 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testutil
+
+// Binary is a type that is used for fields that have a YANG type of
+// binary. It is used such that binary fields can be distinguished from
+// leaf-lists of uint8s (which are mapped to []uint8, equivalent to
+// []byte in reflection).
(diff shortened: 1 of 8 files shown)


testutil/types.go
```diff
@@ -0,0 +1,78 @@
...
+// Bool is a bool type assignable to unions of which it is a subtype.
+type Bool bool
+
+// Unsupported is an interface{} wrapper type for unsupported types. It is
+// assignable to unions of which it is a subtype.
+type Unsupported struct {
+	Value interface{}
+}
+
+type Union1 interface {
```
progressnerd, line 67: Add godoc comment above exported type?

testutil/types.go
```diff
@@ -0,0 +1,78 @@
...
+	IsUnion1()
+}
+
+func (String) IsUnion1()  {}
+func (Int16) IsUnion1()   {}
+func (Int64) IsUnion1()   {}
+func (*Binary) IsUnion1() {}
+
+func (String) IsUnion()  {}
+func (Int64) IsUnion()   {}
```
progressnerd, line 77: what is there no IsUnion() method on Int16?

ytypes/leaf.go
```diff
@@ -71,13 +72,17 @@ func validateLeaf(inSchema *yang.Entry, value interface{}) util.Errors {
 			return util.NewErrs(fmt.Errorf("bad leaf type: expect Bool for empty type for schema %s, have type %v", schema.Name, ykind))
 		}
 		rv = value
+	case reflect.Int8, reflect.Int16, reflect.Int32, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Float64, reflect.String:
+		if ykind != yang.Yunion {
+			return util.NewErrs(fmt.Errorf("bad leaf type: expect %v for union type for schema %s, have type %v", rkind, schema.Name, ykind))
+		}
```
progressnerd, line 78: why do the integer cases (both Int64 above and these cases) not assign "rv = value"?

progressnerd (changes requested):

testutil/types.go
wenovus: added

testutil/types.go
wenovus: I only added methods for those union types that are used in the `ytypes` test. Essentially Int16 is a `Union1`, but not a `Union`.

ytypes/leaf.go
wenovus: Nice find, doing this actually breaks, and the reason is because of some unnecessary complexity in the existing code. This is fixed now.

progressnerd (approved):

robshakir (changes requested):
Couple of questions here -- I think this generally LGTM though.

It'd be good to have a short document that describes the different types of unions that we have in ygot-generated code now -- this can be some cut of the design document that we wrote (just with the selected option).  We can add this to this [doc](https://github.com/openconfig/ygot/blob/master/docs/design.md#yang-union-leaves) in a future change.

ytypes/leaf.go
```diff
@@ -217,22 +220,13 @@ func validateUnion(schema *yang.Entry, value interface{}) util.Errors {
...
-		// A union containing a single enumerated type would simply resolve to the
-		// enum type, which represented directly by a derived Int64 type.
-	default:
-		return util.NewErrs(fmt.Errorf("wrong value type for union %s: got: %T, expect ptr or Int64 (enumerated type)", schema.Name, value))
 	}
 
-	// Unions of enum types are passed as ptr to interface to struct ptr.
-	// Normalize to a union struct.
-	// TODO(wenbli): Remove this if statement in a future PR.
+	// All wrapper unions and unsupported types for simplified unions are passed as ptr to interface to struct ptr.
```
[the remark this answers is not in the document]
wenovus: Yes that's right.

ytypes/leaf_test.go
```diff
@@ -308,14 +309,13 @@ func TestValidateLeaf(t *testing.T) {
 // UnionContainer and types below are defined outside function scope because
 // type methods cannot be defined in function scope.
 type UnionContainer struct {
-	UnionField Union1 `path:"union1"`
+	UnionField testutil.Union1 `path:"union1"`
```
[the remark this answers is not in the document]
wenovus: Renamed to `TestUnion`.

robshakir (approved):
Thanks for the changes here!

PR #443, opened 2020-09-11
Add tests for new Go unions for Unmarshalling JSON

wenovus:


ytypes/leaf_test.go
```diff
@@ -938,27 +938,29 @@ func TestValidateLeafRef(t *testing.T) {
...
+	BinaryLeaf           Binary                `path:"binary-leaf"`
+	BoolLeaf             *bool                 `path:"bool-leaf"`
+	DecimalLeaf          *float64              `path:"decimal-leaf"`
+	EnumLeaf             EnumType              `path:"enum-leaf"`
+	UnionEnumLeaf        EnumType              `path:"union-enum-leaf"`
+	UnionLeaf            UnionLeafType         `path:"union-leaf"`
+	UnionLeaf2           *string               `path:"union-leaf2"`
+	EmptyLeaf            YANGEmpty             `path:"empty-leaf"`
+	UnionLeafSlice       []UnionLeafType       `path:"union-leaflist"`
+	UnionLeafSingleType  []string              `path:"union-stleaflist"`
```
robshakir, line 960: How does this type get generated now? In the previous approach we would generate this simply as `*string`, since there was no need to have a complex type here. Is there a change that has been made to the code that does that - such that we now create some basic type (or is this actually a `leaf-list` of string `union`)?

robshakir (changes requested):

ytypes/leaf_test.go
wenovus: The union name generation code is unchanged (see [last PR](https://github.com/openconfig/ygot/pull/451/files?file-filters%5B%5D=.go)). The union generation code is only changed for when `len(types) > 1`. So, the simple type situation is unchanged from before.

robshakir: Great, thanks.

robshakir (approved):
Thanks for the updates here!

PR #448, opened 2020-09-11
Add support and tests for new Go unions for ygot's marshal functions

wenovus:


ygot/render.go
```diff
@@ -651,8 +668,8 @@ func EncodeTypedValue(val interface{}, enc gnmipb.Encoding) (*gnmipb.TypedValue,
 	switch {
 	case util.IsValueNil(vv) || !vv.IsValid():
 		return nil, nil
-	case vv.Type().Kind() == reflect.Int64:
-		// Invalid int64 that is not an enum.
+	case vv.Type().Kind() == reflect.Int64 && unionTypedefUnderlyingTypes[vv.Type().Name()] == nil:
+		// Invalid int64 that is not an enum or a union typedef type.
```
robshakir, line 672: `or a simple union typedef type (i.e., a union subtype that does not have a struct wrapper)`?

ygot/render.go
```diff
@@ -681,6 +698,10 @@ func EncodeTypedValue(val interface{}, enc gnmipb.Encoding) (*gnmipb.TypedValue,
 		if util.IsNilOrInvalidValue(vv) {
 			return nil, nil
 		}
+	default:
+		if underlyingType, ok := unionTypedefUnderlyingTypes[vv.Type().Name()]; ok && vv.Type().ConvertibleTo(underlyingType) {
+			vv = vv.Convert(underlyingType)
+		}
```
robshakir, line 704: What happens here if the type is not `ConvertibleTo` the type that we specify in the underlying types -- it looks like we'll fall through to returning the `Interface()`, but if this was an underlying type, isn't that an error case?

(In practice, I assume that this actually isn't possible, because we know that our derived 'string in a union' type is actually just an alias to string?)

ygot/render_test.go
```diff
@@ -2344,6 +2596,12 @@ func (*uFieldMulti) IsU() {}
 
 func TestUnionInterfaceValue(t *testing.T) {
 
+	// This is the only test that is used by the simple union API -- it is
```
robshakir, line 2599: Does this comment mean that this test case is only used by the simple union API, or the simple union API only tests this case?

robshakir (changes requested):
A couple of comments here to understand some of the behaviour, PTAL.

ygot/render_test.go
wenovus: This is meant to say that the function under test, `unionPtrValue`, is mostly used for the wrapper union API -- the only simple union API that uses this function is the unsupported type.

I'll update the test function name and this comment to make this clearer.

ygot/render.go
wenovus: Good point, I'll add the error return.

ygot/render.go
```diff
@@ -427,7 +429,7 @@ func findUpdatedLeaves(leaves map[*path]interface{}, s GoStruct, parent *gnmiPat
 			continue
 		case reflect.Interface:
 			// This is a union value.
-			/*val, err := unionInterfaceValue(fval, false)
+			/*val, err := unwrapUnionInterfaceValue(fval, false)
```
robshakir, line 432: Debugging comments?

ygot/render.go
```diff
@@ -1370,11 +1372,11 @@ func jsonValue(field reflect.Value, parentMod string, args jsonOutputConfig) (in
 		var err error
 		switch {
 		case util.IsValueInterfaceToStructPtr(field):
-			if value, err = unionInterfaceValue(field, appmod); err != nil {
+			if value, err = unwrapUnionInterfaceValue(field, appmod); err != nil {
 				return nil, err
 			}
-		case field.Elem().Kind() == reflect.Ptr && field.Elem().Elem().Type().Name() == BinaryTypeName && field.Elem().Elem().Kind() == reflect.Slice:
-			if value, err = jsonSlice(field.Elem().Elem(), parentMod, args); err != nil {
+		case field.Elem().Kind() == reflect.Slice && field.Elem().Type().Name() == BinaryTypeName:
```
robshakir, line 1378: Can you explain this change please?

Previously, we had an interface was dereferenced by `Elem` to get to a pointer which was dereferenced by `Elem` to get to the slice, which we then checked whether it was a `Binary` or not. In this changed code I don't see handling for that case.

In the changed version it looks like you just handle the new `union` types, such that we can't now handle a `union` that contains a `binary` of the old format.

Looking at  https://github.com/openconfig/ygot/blob/master/ygot/render_test.go, we're missing coverage for the old format of `union` that contains `binary`, so it might be worth adding here.

robshakir (changes requested):

ygot/render.go
wenovus: I'm just changing existing debug comments here to match the new function name.

robshakir: Ah, can we remove them completely? I don't think they need to be there.

robshakir (changes requested):

ygot/render.go
wenovus: The change you pointed out is after `Binary` is changed to use value receiver to satisfy the new union API instead of using pointer receiver. Now `*Binary` is no longer accepted.

Binary tests and fixes added for wrapper unions.

robshakir: Oh, I'm sorry - this diff is not from HEAD, it's a diff within this version. This makes more sense.

My understanding was that we had not added this `field.Elem().Kind()` check in this branch, but had rather added it previously, and we were now changing it. This makes more sense to me.

Thanks for adding the test coverage!

robshakir (approved):

PR #449, opened 2020-09-11
Add Support for new Go unions for ygot's GoStruct helper functions

wenovus:


ygot/struct_validation_map.go
```diff
@@ -666,23 +666,53 @@ func copyInterfaceField(dstField, srcField reflect.Value, opts ...MergeOpt) erro
...
+		}
+
+		srcVal := srcField.Elem()
+		ns := reflect.Zero(srcVal.Type())
+		for i := 0; i < srcVal.Len(); i++ {
+			ns = reflect.Append(ns, srcVal.Index(i))
+		}
+		dstField.Set(ns)
+		return nil
+	case util.IsValueScalar(srcField.Elem()) && (isGoEnum || unionTypedefUnderlyingTypes[srcField.Elem().Type().Name()] != nil):
```
robshakir, line 705: Should this be declared in `ygen` since it's really ygen that defines what can be output here, rather than ygot?

robshakir (approved):

ygot/struct_validation_map.go
wenovus: That's not possible because `ygen` depends on `ygot`. There is necessarily some type name information that `ygot` needs to know, e.g. `BinaryTypeName`, so I think it's reasonable to leave it in `ygot.

PR #451, opened 2020-09-11
Add New Go Union Generation under -generate_simple_unions

### LINKED ISSUE #441: Unmarshal doesn't work for Binary, YANGEmpty, and Might be Broken by Unsupported types
Here is a sample of what a current union's `To_XXX_Union` helper with an enumeration, an unsupported type (e.g. instance-identifier), uint32, and binary looks like:
```
    switch v := i.(type) {
    case Binary:
        return &Platform_Component_Power_Union_Binary{v}, nil
    case E_Component_Power:
        return &Platform_Component_Power_Union_E_Component_Power{v}, nil
    case interface{}:
        return &Platform_Component_Power_Union_Interface{v}, nil
    case uint32:
        return &Platform_Component_Power_Union_Uint32{v}, nil
    default:
        return nil, fmt.Errorf("cannot convert ...
    }
```
Here, `[]byte` will fail to match the `Binary` case since `gnmi.TypedValue`'s `GetBytes()` returns `[]byte`, and for the same reason `YANGEmpty` also doesn't work.

A second issue is these cases are listed by key order due to template ordering:
https://github.com/openconfig/ygot/blob/c6ce7e55aa79fe2cdd2f5c655b882b9c98f71fd3/ygen/gogen.go#L999-L1002

A third issue is placing the `interface{}` case before other non-default cases [pre-empts](https://play.golang.org/p/LGaERaij0k0) the rest of the cases. It might make more sense for unsupported types to be put last.

It might make sense, if inconvenient, to simply ignore these issues in the current union API, and fix them in the upcoming simplified API.

wenovus:
This PR contains the code generation (`ygen`) changes for generating the new unions.

The preceding PRs contain modifications for support code to support these new Go unions.

`exampleoc`, `uexampleoc`, `exampleoc/opstateoc` are all converted to use the new Go unions.
The former `exampleoc` is moved to `exampleoc/wrapperunionoc` to continue doing schema tests for the old unions.

NOTE: Click "File filter..." and filter out the .formatted-txt files when reviewing.

Resolves #441


ygen/goelements.go
```diff
@@ -31,6 +31,20 @@ const (
 	goEnumPrefix string = "E_"
 )
 
+// unionConversionSpec stores snippets that converts primitive Go types to
```
progressnerd, line 34: converts->convert

ygen/gogen.go
```diff
@@ -211,11 +211,14 @@ type goStructField struct {
...
 // be generated for a multi-type union in YANG.
 type goUnionInterface struct {
-	Name           string            // Name is the name of the interface
-	Types          map[string]string // Types is a map keyed by the camelcase type name, with values of the Go types in the union.
-	LeafPath       string            // LeafPath stores the path for the leaf for which the multi-type union is being generated.
-	ParentReceiver string            // ParentReceiver is the name of the struct that is a parent of this union field. It is used to allow methods to be created which simplify handling the union in the calling code.
-	TypeNames      []string          // TypeNames is an list of Go type names within the union.
+	Name                 string                 // Name is the name of the interface
+	Types                map[string]string      // Types is a map keyed by the camelcase type name, with values of the Go types in the union.
+	LeafPath             string                 // LeafPath stores the path for the leaf for which the multi-type union is being generated.
```
progressnerd, line 216: path *of* the leaf

ygen/gogen.go
```diff
@@ -211,11 +211,14 @@ type goStructField struct {
...
+	ParentReceiver       string                 // ParentReceiver is the name of the struct that is a parent of this union field. It is used to allow methods to be created which simplify handling the union in the calling code.
+	TypeNames            []string               // TypeNames is an list of Go type names within the union.
```
progressnerd, line 218: *a* list

progressnerd (approved):

wenovus:
This resolves #441 for the new unions. The behaviour of the old unions are not affected.

ytypes/schema_tests/validate_test.go
```diff
@@ -131,7 +132,76 @@ func TestValidateInterface(t *testing.T) {
...
+	if err := vlan0.Validate(); err != nil {
+		t.Errorf("vlan0 success: got %s, want nil", err)
+	}
+
+	// Set vlan-id to be out of range (1-4094)
+	vlan0.Vlan = &oc.Interface_Subinterface_Vlan{
+		VlanId: oc.Uint16(4095),
+	}
+	// Validate the vlan.
+	if err := vlan0.Validate(); err == nil {
```
robshakir, line 148: I think `errdiff` will do this for you - so you can just do:

```golang
err := vlan0.Validate()
if diff := errdiff.Substring(err, ...); diff != ""  {
 // error mesage
}

when `err == nil` then the diff will not be empty.

File: ytypes/schema_tests/validate_test.go [hunk 5]
```diff
@@ -131,7 +132,76 @@ func TestValidateInterface(t *testing.T) {
...
+		}
+		t.Errorf("did not get expected errors when validating device, got:\n %s (len: %d), want 5 errors", b.String(), len(errs.(util.Errors)))
+	}
+}
+
+func TestValidateInterfaceWrapperUnion(t *testing.T) {
+	dev := &woc.Device{}
+	eth0, err := dev.NewInterface("eth0")
+	if err != nil {
+		t.Errorf("eth0.NewInterface(): got %v, want nil", err)
```
robshakir, line 171: `dev.NewInterface` rather than `eth0.NewInterface` (nit)

ytypes/schema_tests/validate_test.go
```diff
@@ -131,7 +132,76 @@ func TestValidateInterface(t *testing.T) {
...
+		t.Errorf("root success: got %s, want nil", err)
+	}
+	// Validate an element in the device subtree.
+	if err := eth0.Validate(); err != nil {
+		t.Errorf("eth0 success: got %s, want nil", err)
+	}
+
+	// Key in map != key field value in element. Key should be "eth0" here.
+	dev.Interface["bad_key"] = eth0
+	if err := dev.Validate(); err == nil {
```
robshakir, line 188: Same thing here as above (I think this is a artifact of my original code :-))

robshakir (changes requested):

demo/bgp/main.go
```diff
@@ -99,15 +99,9 @@ func CreateDemoBGPInstance() (*oc.Bgp, error) {
...
-	// helper function.
-	cid, err := (*oc.Bgp_Neighbor_RouteReflector)(nil).To_Bgp_Neighbor_RouteReflector_RouteReflectorClusterId_Union("10.0.1.2")
-	if err != nil {
-		return nil, err
-	}
-
+	// Set the peer as a route reflector client using the String union helper typedef.
 	bgp.Neighbor["192.0.2.1"].RouteReflector = &oc.Bgp_Neighbor_RouteReflector{
-		RouteReflectorClusterId: cid,
+		RouteReflectorClusterId: oc.String("10.0.1.2"),
```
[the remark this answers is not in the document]
wenovus: I like `oc.UnionString` in order to tell them apart. `Ustring` cannot be used because uint32 would be `Uuint32`.

ygen/gogen.go
```diff
@@ -1010,6 +1056,60 @@ func (t *{{ .ParentReceiver }}) To_{{ .Name }}(i interface{}) ({{ .Name }}, erro
...
+
+	// unionTypeSimpleTemplate outputs the type that corresponds to a multi-type union
+	// in the YANG schema. It does so by enhancing generated typedefs.
+	unionTypeSimpleTemplate = mustMakeTemplate("unionTypeSimple", `
+// {{ .Name }} is an interface that is implemented by valid types for the union
+// for the leaf {{ .LeafPath }} within the YANG schema.
+// Union type can be one of [{{ .SubtypeDocumentation }}].
+type {{ .Name }} interface {
+	// Union type can be one of [{{ .SubtypeDocumentation }}]
+	Documentation_for_{{ .Name }}()
```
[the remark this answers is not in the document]
wenovus: `godoc` doesn't seem to treat this comment any differently -- on the webpage the comment appears as-is, even when the name doesn't start with the name of the function.

The purpose of this different name is to assist with users who might be using autocomplete to figure out a particular union's subtypes. e.g. if the user doesn't know the union subtypes of A_B_Union, they can type out in their IDE A_B_Union, click on this documentation string, and see what the subtypes are.

ytypes/schema_tests/validate_test.go
wenovus: Thanks for catching it, yes I simply copied it -- no time to look for style.

ygen/gogen.go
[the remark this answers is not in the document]
robshakir: Ah, I understand. I was wondering whether there was something specific in terms of calling this `Documentation...` maybe just make it `{{ .Name }}_Doc` since the most important thing seems to be the existence of the method rather than the name?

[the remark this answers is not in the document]
wenovus: For a user who is confused, the `Documentation_...` is an obvious thing to read. Putting `_Doc` last might get hidden behind a really long union name, which contains the entire path of the node. Also, `Doc_{{ .Name }}` is not as clear as `Documentation_for_{{ .Name }}` as to the purpose of the method.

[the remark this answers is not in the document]
robshakir: Sure, that SGTM.

robshakir (approved):

PR #453, opened 2020-09-16
Add Functionality to Trim Organization Name from Module Name

wenovus:
This allows the [organization name](https://tools.ietf.org/html/rfc8407#section-4.1) to be trimmed from the module name.

If this trimming is done in the generated code, it helps to improve autocomplete experience by avoiding having to type it before autocomplete becomes useful.


progressnerd (approved):

robshakir (approved):
This LGTM -- one consequence this might have is that there may now be more name clashes. 

Assuming that there are uncompressed schemas, then we don't trim the module name from the generated code types, so there will non-deterministic code generation if this happens - because we don't do the structured name resolution that we use for enumerations.

Can we add a caveat for this option that the user needs to be aware of this? In practice, it might actually be something that isn't recommended for users that need to support e.g., `ietf-interfaces` and `openconfig-interfaces`.

wenovus:
> Assuming that there are uncompressed schemas, then we don't trim the module name from the generated code types, so there will non-deterministic code generation if this happens - because we don't do the structured name resolution that we use for enumerations.

@robshakir I didn't quite understand this, what do you mean by "structured name resolution"? For uncompressed generation, the enumeration names is simply the entire path -- there is no enumeration name clash resolution for uncompressed schemas. So I'm not seeing what would be causing the non-deterministic code generation.

If you see the next PR: #454, I decided for implementation convenience to allow trimming for uncompressed schemas -- I didn't see the harm because users can simply not use the flag.

robshakir:
`ParentModulePrettyName` doesn't (today) require that it is used only for enumerations AFAICS - so what I was wondering is whether we are now creating a situation where a `struct` might become named something different when this flag is specified. Looking further it doesn't seem to be a problem (`ParentModulePrettyName` is only called in `enumgen.go`), but given that this is in `genutil` it seems like there is this possibility.

My concern about a `struct` being named differently is that we do not apply the same name clash resolution to `struct` and `enum` - so there will be `_` appended to structs that have clashing names. This is non-deterministic because of the random order of processing nodes in the tree (until we change to up-front code generation for Go -- the refactoring changes that we proposed before). 

Why is this a concern in this context? We know we have `ietf-interfaces` and `openconfig-interfaces`, so as soon as you strip IETF or OpenConfig from these names then there will be clashes.

The change that I'd propose is to move this functionality into `ygen` itself, and scope it to enumerated types, rather than providing a util which may be problematic.

I'm also not sure that we should allow this in uncompressed schemas unless there's a real user demand for it.

wenovus:
OK I will change the next PR to ignore this flag for uncompressed schemas.

As for the current trimming logic being in `ParentModulePrettyName` undermining struct name generation though, I'm not seeing this problem because there is a variable number of arguments, so the calling function can simply not supply any prefixes to trim. For example, even if `ygen` today did call this function other than `enumgen.go`, it just passes the `enumOrgPrefixesToTrim` flag to `enumgen.go` only. The naming of the flag serves to prevent the implementor from trimming non-enum names.

Essentially, the util function `ParentModulePrettyName` is getting a backward-compatible extension of a functionality to trim prefixes or not, which helps `enumgen.go` do the trimming. I don't see how extending the function alone creates a danger for misuse.

robshakir:
I don't think there's a problem with the way that this function can be called -- just the fact that today this was intended to be a function that is general purpose for any type. 

I'm of the opinion that trimming an organisation name is not a good idea _in the general case_ -- i.e., there are a lot of cases where this could become quite risky with overlapping trees, such that we should probably think about it very carefully before doing it for anything other than a) OpenConfig only consumers, and b) when we're already applying schema transformations already. Out of that abundance of caution I'm suggesting that we really just make this implemented where we've widely considered the safety of it (i.e., for OC, and for enumerations), such that we have specific functionality that is called here - rather than extend the general use function for a use case we've only considered in one deployment case.

That being said - per the approval, this isn't a blocking comment.

wenovus:
Ok sounds good. I will propose in the next PR to make the generation flag only apply to OpenConfig, instead of allowing it for other organizations.

PR #454, opened 2020-09-16
Add Flag -trim_enum_openconfig_prefix to Trim Enumerated Names

wenovus:
This is a very straightforward change that simply trims the "openconfig-" organization prefix whenever constructing the module part of the various enumerated names (`enumeration` and `identity`).

This flag only affects enums within compressed code generation.


progressnerd (approved):

robshakir (changes requested):
Per my comments in the other PR - I'm not sure that this should apply to uncompressed schemas without a more robust way to deal with name clash sets, given that it will likely cause more. Let's discuss this.

ygen/codegen_test.go
```diff
@@ -1088,6 +1133,11 @@ func TestSimpleStructs(t *testing.T) {
 			wantCode := string(wantCodeBytes)
 
 			if gotCode != wantCode {
+				// FIXME(wenbli): debug
+				if err := ioutil.WriteFile(fmt.Sprintf("/usr/local/google/home/wenbli/tmp/%s", path.Base(tt.wantStructsCodeFile)), []byte(gotCode), 0644); err != nil {
+					panic(err)
+				}
+
```
wenovus, line 1140: Need to remove this.

robshakir (approved):
Thanks for the changes here.

PR #459, opened 2020-09-25
Error out for Go code generation if list key contains binary

### LINKED ISSUE #458: Code Generation Should Fail when Binary is a list key type
This applies to both simple binary types and binary types within unions.

wenovus:
Since binary is defined as `type Binary []byte`, an unhashable type, we
currently do not support in Go such YANG models where a binary could
occupy a list key.

Fixes #458


wenovus:
Closes #458

ygen/codegen.go
```diff
@@ -401,6 +422,10 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
 	enumTypeMap := map[string][]string{}
 	var structSnippets []GoStructCodeSnippet
 	for _, directoryName := range orderedDirNames {
+		if errs := checkForBinaryKeys(dirNameMap[directoryName]); len(errs) != 0 {
```
robshakir, line 425: Do we want to do this ahead of time, or should we just do it when we get to the code generation stage? It seems like we could avoid walking all directories here ahead of doing that generation.

The pro of having this implemented this way seems to be that we wouldn't need this in both ypathgen and ygen/Go -- but it's only called here today. Is my understanding of this choice right?

wenovus: The advantage is that only Go generation disallows this scenario. In particular, `ypathgen` is fine with a binary as a list key, so it shouldn't see an error just because the generated Go code doesn't like it. Thus, `buildDirectoryDefinition()` should not error out due to this condition.

robshakir: Sure, that makes sense to me. I guess `ypathgen` isn't OK with this if we're allowing retrieval of non-leaf nodes, since the structs that are also generated for ypathgen won't compile?

wenovus: So `ypathgen` doesn't have that functionality, it can only generate paths for now. But yes, if `ypathgen` were to be extended to associate with each path with a `Get()` or other method that returns the corresponding `GoStruct`, then `ypathgen` should check this and error out.

robshakir: Ah, yes, apologies -- I forgot about the split of functionality here.

robshakir (approved):

PR #460, opened 2020-09-29
Parse compressed-out paths into new ShadowFields field for ygen.Directory

wenovus:
In the future, the implementation should be changed to adding an extra field in the IR `Directory` type
called `ShadowedFields` to fit with the refactoring effort.


genutil/common.go
```diff
@@ -169,7 +186,10 @@ func TranslateToCompressBehaviour(compressPaths, excludeState, preferOperational
 // FindAllChildren finds the data tree elements that are children of a YANG entry e, which
 // should have code generated for them. In general, this means data tree elements that are
 // directly connected to a particular data tree element; however, when compression of the
-// schema is enabled then recursion is required.
+// schema is enabled then recursion is required. The second return value is
+// only non-empty during compression, and it contains the fields that have been
```
robshakir, line 190: `only populated when compression is enabled`?

genutil/common.go
```diff
@@ -169,7 +186,10 @@ func TranslateToCompressBehaviour(compressPaths, excludeState, preferOperational
...
+// duplicated via compression, and are thus "shadow" fields of their
```
robshakir, line 191:
```
that have been removed, due to having a duplicate (e.g., `config/foo` is a duplicate of `state/foo` when `PreferOperationalState` is used)
```

genutil/common.go
```diff
@@ -310,19 +330,23 @@ func FindAllChildren(e *yang.Entry, compBehaviour CompressBehaviour) (map[string
...
-	// that are added as children. This is a whitelist for any duplicate
+	// that are added as children. This is a whitelist for any shadowed
 	// names in the deprioritized data container.
 	prioNames := map[string]bool{}
 	// directChildren is used to store the nodes that will be mapped to be direct
 	// children of the struct that represents the entry e being processed. It is
 	// keyed by the name of the child YANG node ((yang.Entry).Name).
 	directChildren := make(map[string]*yang.Entry)
+	// shadowChildren is are direct children that are duplicates of direct
+	// childrens that have the same name and de-prioritized due to the
```
robshakir, line 341: `that are duplicates of direct children`

genutil/common.go
```diff
@@ -336,17 +360,21 @@ func FindAllChildren(e *yang.Entry, compBehaviour CompressBehaviour) (map[string
...
-					// that is returned, as we allow those duplicates.
+					ch := map[string]*yang.Entry{"": configStateChild}
 					if util.IsChoiceOrCase(configStateChild) {
-						// Duplicates could occur in a choice/case as well.
-						errs = addNonChoiceChildrenDuplist(directChildren, configStateChild, prioNames, errs)
-					} else if !prioNames[configStateChild.Name] {
-						errs = addNewChild(directChildren, configStateChild.Name, configStateChild, errs)
+						// Compress out (do not map) choice/case nodes that are in the
+						// config or state container. This is again specifically for the
+						// OpenConfig routing policy model.
```
robshakir, line 367: Suggest to remove this comment re: the OC routing policy model, since this has been removed.

ygen/directory.go
```diff
@@ -135,9 +136,23 @@ func GetOrderedDirectories(directory map[string]*Directory) ([]string, map[strin
 // of a Directory. The Field is specified as a name in order to guarantee its
 // existence before processing.
 func FindSchemaPath(parent *Directory, fieldName string, absolutePaths bool) ([]string, error) {
+	return findSchemaPath(parent, fieldName, false, absolutePaths)
+}
+
+// findSchemaPath finds the relative or absolute schema path of a given field
+// of a Directory, or the shadowed field path (field duplicated and
```
robshakir, line 143: Do we need `shadowSchemaPaths` as an argument here. The `Directory` has been built with some particular behaviour in mind, so I wonder whether the right implementation here is:

- check whether this field was included in the `Directory`
- if it wasn't, then look for it in the shadowed fields.
- do the schema path resolution

I guess this is a bit harder to implement, since we need the paths for each `Entry` compared to this approach - but it might be neater as a way to be able to make `FindSchemaPath` have a simpler behaviour and not need to be given anything about the way that the IR was created in the future.

robshakir (changes requested):

genutil/common.go
wenovus: Reworded.

ygen/directory.go
wenovus: I'm not quite seeing how this flag can be removed, because one of the new purposes of `FindSchemaPath` is to populate the new `shadow-path` struct tag, and without indicating to the function whether it wants the shadow or non-shadow path, it wouldn't know to do this.

For context see snippet from next PR (#429):
https://github.com/openconfig/ygot/blob/56a5401e16957ffda2481fcab16d15dbb91ef937/ygen/gogen.go#L1538-L1562

robshakir: Sorry for the delay replying here -- I'm a bit confused. The snippet you linked to doesn't seem to call `FindSchemaPath` right?

My first confusion is that this function is meant to be able to return either the shadow paths for a field that is within the view of the schema that we have - or its shadowed versions. For example, for `/foo/bar/config/baz` with `PreferIntendedConfig` on, then this should return `/foo/bar/config/baz` when `shadowSchemaPaths` is set to `true` and `/foo/bar/state/baz` when `shadowSchemaPaths` is set to `false`, right? In this model, we wouldn't have a way to be able to determine that within a particular directory an operational-state only field that has been removed is valid, I think -- which seems fine. This was my initial understanding of what we were trying to do (say whether any path was valid within the schema despite compression).

As for the exact implementation of the logic discussed above. AIUI, we're adding a new shadow path to the struct tags, which we might need to populate if this field happened to have both an intended and applied path. My assumption was that we would generate directories that either had `Directory{}.ShadowedFields` populated or not, and we wouldn't necessarily need to have this logic be in a common `findSchemaPath` implementation -- rather we would do something like:

* `findRealSchemaPaths` -> form `path` struct tag.
* `findShadowedSchemaPath` -> form `shadowPath` struct tag if the list is not-nil.

If the `ShadowedFields` map is empty then we would always get `nil` back from the 2nd function. I guess you're intending to have my two "real" and  "shadow" paths be one function with a flag? I think it'd be clearer as two - but I'm happy with improving the commenting such that we can say that `shadowSchemaPaths` actually changes the source of truth for this this function from the 'real' fields to those that were removed.

wenovus: Sorry for being not as clear as I could be. Please feel free to just have me explain again.

The snippet I linked calls `findMapPaths`, which in turn calls `findSchemaPath`:
https://github.com/openconfig/ygot/blob/56a5401e16957ffda2481fcab16d15dbb91ef937/ygen/gogen.go#L2083-L2088

`findMapPaths` is called twice during Go generation, once for retrieving the unshadowed leaf path, and once for retrieving the shadowed leaf path, which are then used to populate the struct tags. The boolean flag simply tells `findMapPaths` and `findSchemaPath` which tag is being populated so it can retrieve the correct path; therefore, if the shadow path didn't exist (say it's a state leaf), the function *should* return nothing. You can see how this is done in this test here (file from this PR): https://github.com/openconfig/ygot/blob/428adc619285f99ed99bf43561318be3ef37f249/ygen/genstate_test.go#L104-L177

robshakir: Makes sense - apologies for the delay in me understanding this!

robshakir (approved):

PR #462, opened 2020-10-04
Update to goyang 0.2.0 and change type of Min/Max Elements

hslatman:
Recently the type of MinElements and MaxElements in the `goyang` repository was changed to uint64 (see this [PR](https://github.com/openconfig/goyang/pull/125)). It was released in v0.2.0.

I've updated the version used in `ygot` to use the v0.2.0 release and changed the usage of MinElements and MaxElements accordingly. I've commented out three tests that I think are no longer in scope, but maybe I should just change them a bit.

This PR is a draft, because I don't know what the policy is for upgrading dependencies and backwards compatibility.

### DIFF
--- a/experimental/ygotutils/getnode_test.go
+++ b/experimental/ygotutils/getnode_test.go
@@ -89,7 +89,7 @@ func TestGetNodeSimpleKeyedList(t *testing.T) {
 					"simple-key-list": &yang.Entry{
 						Name:     "simple-key-list",
 						Kind:     yang.DirectoryEntry,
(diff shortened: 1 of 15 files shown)


hslatman:
Many tests fail due to MaxElements being set to 0. How can I best solve this?

wenovus:
I would recommend using [go modules](https://blog.golang.org/using-go-modules) to avoid these kinds of dependency issues, as they're likely to come up in the future as well. We can't guarantee close coordination between goyang and ygot.

wenovus:
Dup of #463.

PR #463, opened 2020-10-05
Update goyang dependency to v0.2.1

wenovus:
This update makes ygot compatible with the new backwards-incompatible min/max-elements changes to goyang.

The schemas need to be updated for them to be populated with the new `MinElement`/`MaxElement` values.

If https://github.com/openconfig/goyang/pull/147 is also merged, can depend on that function instead of defining a new one in util.

@hslatman discussed about this issue in #462


hslatman:
Looks nice! 😄

ytypes/util_schema.go
```diff
@@ -88,13 +88,11 @@ func validateListAttr(schema *yang.Entry, value interface{}) util.Errors {
 		return util.NewErrs(fmt.Errorf("schema %s ListAttr is nil", schema.Name))
 	}
 
-	var size int
-	if value == nil {
-		size = 0
-	} else {
+	var size uint64 = 0
```
robshakir, line 91: This can just be `var size uint64` - it'll be initialised with the zero-value anyway.

ytypes/util_schema.go
```diff
@@ -103,29 +101,12 @@ func validateListAttr(schema *yang.Entry, value interface{}) util.Errors {
...
-			errors = util.AppendErr(errors, err)
-		} else if max, err := maxN.Int(); err != nil {
-			errors = util.AppendErr(errors, err)
-		} else if max < 0 {
-			errors = util.AppendErr(errors, fmt.Errorf("list %s has negative max required elements", schema.Name))
-		} else if int64(size) > max {
-			errors = util.AppendErr(errors, fmt.Errorf("list %s contains more than max allowed elements: %d > %d", schema.Name, size, max))
-		}
+	if size > schema.ListAttr.MaxElements {
+		errors = util.AppendErr(errors, fmt.Errorf("list %s contains more than max allowed elements: %d > %d", schema.Name, size, schema.ListAttr.MaxElements))
```
robshakir, line 108: This is a lot neater, thanks for the change here.

ytypes/util_test.go
```diff
@@ -304,7 +305,7 @@ func TestStringToKeyType(t *testing.T) {
 	listSchema := &yang.Entry{
 		Name:     "struct-key-list",
 		Kind:     yang.DirectoryEntry,
-		ListAttr: &yang.ListAttr{MinElements: &yang.Value{Name: "0"}},
+		ListAttr: util.NewListAttr(),
```
robshakir, line 308: I'd suggest putting this in `testutil` rather than `util`, since it's not quite general-purpose enough for a utility is it (we might expect that to have a way to specify min/max values). If we decide not to put it in `testutil` then at least it should be named something like `NewDefaultListAttr`.

robshakir (changes requested):

ytypes/util_test.go
wenovus: I'm waiting for https://github.com/openconfig/goyang/pull/147 to be merged, and to use the function from goyang instead.

robshakir (approved):

hslatman:
Thank you! 😄

PR #464, opened 2020-10-05
Add `util.PathElemSlicesEqual` to compare slices of *gpb.PathElem.

wenovus:
Now we have
`util.PathElemsEqual` (same as before)
`util.PathElemSlicesEqual`


robshakir:
Do we have to make this a backwards incompatible change? I guess the reason that you chose to is the naming of `Elems` vs `Elem` in the function name? It's reasonable to estimate the impact to this code, but seems like it requires other callers to change for relatively little benefit?

robshakir (approved):
LGTM, thanks for the changes here.

PR #466, opened 2020-10-20
Add docstring on valid types for list keys for list constructors

wenovus:
Whenever a list path constructor has a parameter, it has a docstring on the valid types for the parameter.

This is mostly useful for knowing which types are valid for unions when constructing list paths using `ypathgen`.
e.g.
```
// Neighbor: string
// SidId: [UnionUint32, E_MplsTypes_MplsLabel_Enum, UnionString, E_AdjacencySid_SidId]
```

Since the old unions are no longer recommended, only new union names are displayed.

### DIFF
--- a/ygen/genstate.go
+++ b/ygen/genstate.go
@@ -16,6 +16,7 @@ package ygen
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/openconfig/goyang/pkg/yang"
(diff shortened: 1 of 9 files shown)


exampleoc/ocpath.go
```diff
@@ -97,6 +99,7 @@ func (n *AclPathAny) AclSetAnyType(Name string) *Acl_AclSetPathAny {
 }
 
 // AclSetAnyName returns from AclPath the path struct for its child "acl-set".
+// Type: E_Acl_ACL_TYPE
```
progressnerd, line 102: I see examples in other files, where the type name is preceded by the "[modulename.] in front. Is that what. we can expect to happen in practice?

progressnerd (approved):
One comment below

exampleoc/ocpath.go
wenovus: In this case, as well as likely in practice, the path library is **the same** as the goStructs library, so there is no "modulename." prefix.

progressnerd: Should it have the name of it's own self in front (the likely name by which it's imported), or just assume the user knows that?

wenovus: It's not possible to know what the user's import prefix is -- the user might give an alias to the package path, and then it's no longer the package name.

PR #467, opened 2020-10-22
Add PathElemsMatchQuery helper for checking path membership to a telemetry path

wenovus:
This helper checks whether a particular concrete path matches a query
path per [gNMI path specifications](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md)

NOTE: the new pathstrings.go is **COPIED** from the ygot package, as
otherwise we have a circular dependency. In the future, the copied
helpers should move to the util package.

### DIFF
--- a/util/gnmi.go
+++ b/util/gnmi.go
@@ -134,6 +134,41 @@ func FindPathElemPrefix(paths []*gpb.Path) *gpb.Path {
 	}
 }
 
+// pathElemMatchesQuery returns true if the given concrete (no wildcard)
+// PathElem is a possible gNMI telemetry name match for the reference PathElem.
+// In particular, it matches wildcards for names and list keys, but *not* "...".
+// See https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md
+func pathElemMatchesQuery(elem, refElem *gpb.PathElem) bool {
+	if elem == nil || refElem == nil {
+		return elem == nil && refElem == nil
+	}
+
+	if refElem.Name != "*" && elem.Name != refElem.Name {
+		return false
+	}
+
+	for k, ref := range refElem.Key {
+		if v, ok := elem.Key[k]; !ok || (ref != "*" && v != ref) {
+			return false
+		}
+	}
+	return true
+}
+
+// PathElemsMatchQuery returns true if the given concrete (no wildcard)
+// PathElem path slice is a possible gNMI telemetry path match for the
+// reference PathElem path slice.
+// In particular, it matches wildcards for names and list keys, but *not* "...".
+// See https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md
+func PathElemsMatchQuery(elems, refElems []*gpb.PathElem) bool {
+	for i := 0; i != len(elems) && i != len(refElems); i++ {
+		if !pathElemMatchesQuery(elems[i], refElems[i]) {
+			return false
+		}
+	}
+	return len(refElems) <= len(elems)
+}
+
 // PopGNMIPath returns the supplied GNMI path with the first path element
 // removed. If the path is empty, it returns an empty path.
 func PopGNMIPath(path *gpb.Path) *gpb.Path {
--- a/util/pathstrings.go
+++ b/util/pathstrings.go
@@ -1,0 +1,131 @@
+// Copyright 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"strings"
+
+	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
+)
+
+// stringToStructuredPath takes a string representing a path, and converts it to
+// a gnmi.Path, using the PathElem element message that is defined in gNMI 0.4.0.
+// XXX: This is copied code from ygot package. ygot's code should probably
+// live in this package instead.
+func stringToStructuredPath(path string) (*gnmipb.Path, error) {
+	parts := PathStringToElements(path)
+
+	gpath := &gnmipb.Path{}
+	for _, p := range parts {
+		name, kv, err := extractKV(p)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing path %s: %v", path, err)
+		}
+		gpath.Elem = append(gpath.Elem, &gnmipb.PathElem{
+			Name: name,
+			Key:  kv,
+		})
+	}
+	return gpath, nil
+}
+
+// extractKV extracts key value predicates from the input string in. It returns
+// the name of the element, a map keyed by key name with values of the predicates
+// specified. It removes escape characters from keys and values where they are
+// specified.
+// XXX: This is copied code from ygot package. ygot's code should probably
+// live in this package instead.
+func extractKV(in string) (string, map[string]string, error) {
+	var inEscape, inKey, inValue bool
+	var name, currentKey string
+	var buf bytes.Buffer
+	keys := map[string]string{}
+
+	for _, ch := range in {
+		switch {
+		case ch == '[' && !inEscape && !inValue && inKey:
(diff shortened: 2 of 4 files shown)


util/pathstrings_test.go
```diff
@@ -0,0 +1,290 @@
...
+	"strings"
+	"testing"
+
+	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/proto"
+
+	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
+)
+
+// XXX: This is copied code from ygot package. ygot's code should probably
```
robshakir, line 27: Is there a reason to take on this debt in the codebase rather than fix it now? It seems like there'd be relatively minimal refactoring here -- i.e., we move the actual implementation to util, and keep the public APIs in ygot just calling these util functions.

robshakir: We maybe also should consider whether `util` itself is going to grow into a similar situation, and create a smaller package just for path handling functions.

util/gnmi.go
```diff
@@ -134,6 +134,41 @@ func FindPathElemPrefix(paths []*gpb.Path) *gpb.Path {
 	}
 }
 
+// pathElemMatchesQuery returns true if the given concrete (no wildcard)
+// PathElem is a possible gNMI telemetry name match for the reference PathElem.
+// In particular, it matches wildcards for names and list keys, but *not* "...".
+// See https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md
+func pathElemMatchesQuery(elem, refElem *gpb.PathElem) bool {
```
robshakir, line 141: I'd suggest to call the first argument here `queryElem` and the second `pathElem` so that it's clear which of the two we're trying to match against each other.

util/gnmi.go
```diff
@@ -134,6 +134,41 @@ func FindPathElemPrefix(paths []*gpb.Path) *gpb.Path {
 	}
 }
 
+// pathElemMatchesQuery returns true if the given concrete (no wildcard)
```
robshakir, line 137: Can you clarify `concrete` here? I think that you're trying to say if the given query element matches the reference element -- but the `no wildcard` disclaimer is throwing me off, because it seems like wildcards are handled?

util/gnmi.go
```diff
@@ -134,6 +134,41 @@ func FindPathElemPrefix(paths []*gpb.Path) *gpb.Path {
...
+	if elem == nil || refElem == nil {
+		return elem == nil && refElem == nil
+	}
+
+	if refElem.Name != "*" && elem.Name != refElem.Name {
+		return false
+	}
+
+	for k, ref := range refElem.Key {
+		if v, ok := elem.Key[k]; !ok || (ref != "*" && v != ref) {
```
robshakir, line 151: If `ref == "*"` then you don't need to check that `v != ref` right? Because even if `v == ref` then if `ref == *` we'll return true.

util/gnmi.go
```diff
@@ -134,6 +134,41 @@ func FindPathElemPrefix(paths []*gpb.Path) *gpb.Path {
...
+func pathElemMatchesQuery(elem, refElem *gpb.PathElem) bool {
+	if elem == nil || refElem == nil {
+		return elem == nil && refElem == nil
+	}
+
+	if refElem.Name != "*" && elem.Name != refElem.Name {
+		return false
+	}
+
+	for k, ref := range refElem.Key {
```
robshakir, line 150: There's one confusing thing here -- you say that the reference path must not include wildcards, but then we check whether `ref == "*"`, which can presumably never happen.

We should probably note that if the key is not specified in the reference path, then we never check the value in the queried path, so therefore we assume that the element matches. This seems fine, but we should say that this function isn't schema aware -- it just takes in good faith that both paths are valid schema paths.

util/gnmi.go
```diff
@@ -134,6 +134,41 @@ func FindPathElemPrefix(paths []*gpb.Path) *gpb.Path {
...
+	}
+	return true
+}
+
+// PathElemsMatchQuery returns true if the given concrete (no wildcard)
+// PathElem path slice is a possible gNMI telemetry path match for the
+// reference PathElem path slice.
+// In particular, it matches wildcards for names and list keys, but *not* "...".
+// See https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md
+func PathElemsMatchQuery(elems, refElems []*gpb.PathElem) bool {
```
robshakir, line 163: Similar comment here as above about making the concrete wording clearer, and the argument names.

util/pathstrings.go
```diff
@@ -0,0 +1,131 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
+)
+
+// stringToStructuredPath takes a string representing a path, and converts it to
+// a gnmi.Path, using the PathElem element message that is defined in gNMI 0.4.0.
+// XXX: This is copied code from ygot package. ygot's code should probably
```
robshakir, line 28: Same comment here about making the change to avoid code duplication.

util/pathstrings.go
```diff
@@ -0,0 +1,131 @@
...
+		})
+	}
+	return gpath, nil
+}
+
+// extractKV extracts key value predicates from the input string in. It returns
+// the name of the element, a map keyed by key name with values of the predicates
+// specified. It removes escape characters from keys and values where they are
+// specified.
+// XXX: This is copied code from ygot package. ygot's code should probably
```
robshakir, line 51: Same as above - please don't duplicate code.

robshakir (changes requested):

util/pathstrings_test.go
wenovus: From a quick scan of the internal dependencies it looks like it's possible to split out all **gnmi**-path related functions into a new package. This is all of `gnmi.go` and the `GetNode` and relatives in `reflect.go`.

If we move **all** path-related functions into a new package, we'll essentially be moving `util` to a new package.

So, I'm not sure if we can quite create a package called `pathutil`, but `gnmiutil` is still possible.

wenovus: After looking at the internal dependencies further, I'm still not seeing why we should split the path functions into its own package.

The `util` package has roughly the following high-level components in topological/dependency order:
1. Debug utilities
2. Basic error type
3. ygot-specific goyang schema helpers
4. string path helpers operating on the goyang schema, which uses 3
5. gNMI path helpers, which uses 4
 - 4 can be separated into helpers that don't use the goyang schema (4a) vs. do (4b), and 5 only depends on those that don't (4b); however, 4a&4b are so intertwined that 4b also depends on 4a's internal methods, so it doesn't make sense functionally to split 4 into two.
6. GoStruct helpers
7. GoStruct helpers that operating on goyang's schema, which uses 4
8. GoStruct traversal functions based on goyang's schema, which uses 4, 6-7
9. GoStruct traversal functions based on gNMI paths and goyang's schema, that uses 3-7.

While 5&9 can be separated into another package called `gnmiutil`, which would contain pathstrings.go as well, it would be splitting based on that they both use `gpb.Path` -- the meaning of `gnmiutil` would still be unclear to the reader, which I think is the true problem: that when a reader reads `util`, they don't know what it does.

I think it makes sense, rather, to split `util` into the following packages in topological order:
- `util`: 1-3
- `pathutil`: 4-5 + ygot/pathstrings.go
- `gostructutil` 6-9
What do you think @robshakir

robshakir: I think this set sounds a good set to have. Apologies for the delay in responding here. I'm happy with this split, which AIUI should mean that we don't need to duplicate code here.

wenovus: Since generic package names are not recommended, perhaps a better set of names would be
`internal`: 1 https://golang.org/doc/go1.4#internalpackages
`ygoterrors`: 2
`yangschema`: 3 (we have the option of simplifying some of the function names, e.g. `util.SchemaPath` -> `yangschema.Path`)
`yangpath`: 4-5 + ygot/pathstrings.go
`gostruct`: 6-9

We might also consider renaming `genutil` to `ygotgen`.

wenovus: Part of util refactoring effort: https://github.com/openconfig/ygot/issues/556

robshakir:
Wen, this was on the review stack but looks to have outstanding comments. Is there anything you need from me here?

wenovus:
It's actually not this PR, but rather this comment: https://github.com/openconfig/ygot/pull/467#discussion_r707869638
I'm wondering if you think it's worth splitting the util package up into these packages in order to make the functionalities more modular. If you think this is a good idea we might consider doing this prior to v1 release.

util/pathstrings_test.go
robshakir: Ah, sorry -- I thought that this was what you had decided upon. I think that this sounds a reasonable split -- I'm all OK with dividing out to very specific packages, I think in general it means that we end up with a cleaner import set from other modules.

PR #474, opened 2020-11-24
Add protogenerator flag -consistent_union_enum_names

### LINKED ISSUE #471: Inconsistent Generated proto enum Names
https://developers.google.com/protocol-buffers/docs/style#enums
```
enum FooBar {
  FOO_BAR_UNSPECIFIED = 0;
  FOO_BAR_FIRST_VALUE = 1;
  FOO_BAR_SECOND_VALUE = 2;
}
```

The underscores are currently omitted in the generated protos.

wenovus:
Add protogenerator flag `-consistent_union_enum_names`.

This flag, when turned on along with `-typedef_enum_with_defmod`, does the
following:
- Uses standard enumeration naming for enumeration within unions (like Go).
- Removes non-typedef enumerations within non-typedef unions from the global enum definition file.
- Removes non-typedef enumerations within **typedef** unions from the messages definition.

When this flag is flipped, it automatically adds the "Enum" suffix for
simple union enumerations.

Resolves #471.

For union enums, the current way to name them in the `enums.proto` generated file is to use the "valuePrefix" annotation instead of their assigned name from enumgen.go. This is inconsistent with the Go generated code, and I don't see the value of doing it -- there is no added readability benefit, and there is chance for a name conflict either.

The effect of this change is that the generated union enums in `enums.proto` now follow the generated name given by enumgen.go instead of the path of the union.
- For typedef enums, this means the name is now deterministic, and not dependent on which was the first path where the typedef was encountered during traversal:
```enums.proto
 // OpenconfigRibBgpTestUnionEnumEnum represents an enumerated type generated for the YANG enumerated type union.
 enum OpenconfigRibBgpTestUnionEnumEnum {
-  OPENCONFIGRIBBGP_BGPRIB_AFISAFIS_AFISAFI_STATE_FOO_UNSET = 0;
-  OPENCONFIGRIBBGP_BGPRIB_AFISAFIS_AFISAFI_STATE_FOO_ALPHA = 1 [(yext.yang_name) = "ALPHA"];
-  OPENCONFIGRIBBGP_BGPRIB_AFISAFIS_AFISAFI_STATE_FOO_BRAVO = 2 [(yext.yang_name) = "BRAVO"];
+  OPENCONFIGRIBBGPTESTUNIONENUMENUM_UNSET = 0;
+  OPENCONFIGRIBBGPTESTUNIONENUMENUM_ALPHA = 1 [(yext.yang_name) = "ALPHA"];
+  OPENCONFIGRIBBGPTESTUNIONENUMENUM_BRAVO = 2 [(yext.yang_name) = "BRAVO"];
 }
```
- Non-typedef enums' names are only affected via suffixing "Enum" in their local proto message definitions.
```enums.proto
message Outer {
  message Inner {
    enum Leaf1Enum {
      LEAF1ENUM_UNSET = 0;
      LEAF1ENUM_ONE = 1 [(yext.yang_name) = "ONE"];
      LEAF1ENUM_TWO = 2 [(yext.yang_name) = "TWO"];
      LEAF1ENUM_THREE = 3 [(yext.yang_name) = "THREE"];
    }
    oneof leaf1 {
      Leaf1Enum leaf1_leaf1enum = 407657582;
      uint64 leaf1_uint64 = 388491926;
    }
  }
}
```
---
Old aside comments, keeping for history. These have been incorporated in this PR.

Aside1: If we're to fix the TODO referenced above, the proposal is to remove the unused local enum generated for typedef enums, and remove the unused global (i.e. enums.proto) enum generated for inlined enums.

Aside2: You might have noticed the "EnumEnum" stuttering above. The reason is because an "Enum" suffix is currently given to typedef union enums by enumgen.go. This is inconsistent with inlined union enums where a suffix is not applied. I would be happy to make this consistency change as well, but it would break current users who are using the "-typedef_enum_with_defmod" generation flag.


robshakir:
I wonder whether for the suffix case we should actually apply the suffix even for inlined enumerations -- there was this bug: https://github.com/openconfig/gribi/issues/3 which raised the issue that enumerations had a name clash if they happened to be within a union (since in this case we generate `oneof Foo` and `enum Foo`).

ygen/protogen.go
```diff
@@ -819,19 +819,8 @@ func writeProtoEnums(enums map[string]*yangEnum, annotateEnumNames bool) ([]stri
 			}
 			p.Values = ge.Values
 
-			// If the supplied enum entry has the valuePrefix annotation then use it to
-			// calculate the enum value names.
+			// Capitalize name per proto style.
 			p.ValuePrefix = strings.ToUpper(enum.name)
-			if e, ok := enum.entry.Annotation["valuePrefix"]; ok {
```
robshakir, line 825: Can we make this be behind a flag? (It looks like today this will be independent of whether `-typedef_enum_with_defmod` is set or not) 

Given that we want users to be able to be backwards compatible here if they need to regenerate their protos and don't want this change - then I suspect we maybe have to have another flag -- if so, we should also fix the suffixing issue before exposing the change to the protogenerator binary.

wenovus: A flag is added, where the "Enum" suffix is now added for simple union enums, and the redundant union enums have also been removed from the generated code (please see first section of the newly updated PR description)

robshakir: Great, thanks!

robshakir (approved):

proto_generator/protogenerator.go
```diff
@@ -52,6 +52,7 @@ var (
 	preferOperationalState               = flag.Bool("prefer_operational_state", false, "If set to true, state (config false) fields in the YANG schema are preferred over intended config leaves in the generated messages with compressed schema paths. This flag is only valid for compress_paths=true and exclude_state=false.")
 	skipEnumDedup                        = flag.Bool("skip_enum_deduplication", false, "If set to true, all leaves of type enumeration will have a unique enum output for them, rather than sharing a common type (default behaviour).")
 	useDefiningModuleForTypedefEnumNames = flag.Bool("typedef_enum_with_defmod", false, "If set to true, all typedefs of type enumeration or identity will be prefixed with the name of its module of definition instead of its residing module.")
+	useConsistentNamesForProtoUnionEnums = flag.Bool("consistent_union_enum_names", false, "If set to true, uses more consistent names for enumerations defined within a union. This flag is only active when typedef_enum_with_defmod is also set.")
```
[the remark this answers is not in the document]
wenovus: I've left it out for users for that reason. I still want to keep "proto" because within enumgen.go there is no longer that context (as these all become free-floating flags), and using a different name once the flag gets there makes the code harder to read IMO, which defeats the point of keeping the name tidy.

PR #479, opened 2020-12-02
Remove TODOs on shadow path limitations during JSON unmarshalling

wenovus:
#468 fixes the vast majority of this error detection issue, so the TODOs are removed. It's not perfect, so added comment on its limitations.


robshakir:
As we discussed, I think that this is not necessarily an issue.

* A 'shadow' leaf can only occur where a leaf exists in both `config` and `state`.
* We do not allow containers to be children of `config`, only `state` - therefore all shadow leaves must be direct parents of `config` and `state`.

So we should not get cases where there are shadow leaves that cannot be verified.

This is already covered by the linter today, I added a test case in the [`oc-pyang`](https://github.com/openconfig/oc-pyang/pull/35) repo to ensure that this is covered.

robshakir:
Wen, should we abandon this PR at this point?

wenovus:
Not yet. I see two things to do here:
1. Remove the TODOs.
2. (optional) We preferably should error out for non-leaves that we receive for shadow JSON nodes.

I think 2 is very low priority so I'm ok with just dropping it, but for now I see a bug actually in the current code, so let me fix that first.

robshakir (approved):
LGTM, thanks for the changes.

PR #480, opened 2020-12-02
Treat "max-elements" as "unbounded" during validation

### LINKED ISSUE #475: list node contains more than max allowed elements
I have a very simple model like this 

```
grouping e2node-info {
    description
      "The top level container";
    leaf ip {
      type inet:ip-address;
      description
        "The IP address of the node";
    }
    leaf port {
      type inet:port-number;
      description
        "The port of the node";
    }
    leaf plmn-id {
      type string;
      description
        "PLMN id";
    }
    leaf id {
      type string;
      description
        "eNB id";
    }
  }

  container nodes {
    list node {
      key "id";
      uses e2node-info;
      description
        "e2 node";
    }
    description
      "The list of e2 nodes";
  }
```

but when I try to set a value using gnmi_cli like this 

```
gnmi_cli -address localhost:5150 -set \
    -insecure \
    -proto "update: <path: <elem: <name: 'nodes'> elem: <name: 'node' key:<key:'id' value:'test2'>> elem: <name: 'plmn-id'>> val: <string_val: '123'>>" \
    -timeout 5s -alsologtostderr;;
```

I am getting the following error:

```
E1124 12:32:04.065760   17020 gnmi_cli.go:180] target returned RPC error for Set("update:{path:{elem:{name:\"nodes\"} elem:{name:\"node\" key:{key:\"id\" value:\"test2\"}} elem:{name:\"plmn-id\"}} val:{string_val:\"123\"}}") : rpc error: code = Internal desc = error in creating config struct from IETF JSON data: /device/nodes: /device/nodes/node: list node contains more than max allowed elements: 1 > 0
```

Any ideas why this should happen?

wenovus:
This makes `yang.Entry` more useable by making the default construction of `ListAttr{}` valid, obviating having to call `yang.NewDefaultListAttr` in order to give `MaxElements` the value of `MaxUint64`.

Resolves #475.


ytypes/list_test.go
```diff
@@ -219,7 +219,7 @@ func TestValidateListSimpleKey(t *testing.T) {
 	listSchema := &yang.Entry{
 		Name:     "list-schema",
 		Kind:     yang.DirectoryEntry,
-		ListAttr: yang.NewDefaultListAttr(),
```
robshakir, line 222: Why did this change happen? Shouldn't `NewDefaultListAttr()` give us a range of 0->max per [the current implementation](https://github.com/openconfig/goyang/blob/fd7d014c61f43a05d53146582e61503f374787df/pkg/yang/entry.go)?

robshakir: Is the concern here where there is an older implementation of goyang being used? Shouldn't Go modules help us with this?

ytypes/util_schema.go
```diff
@@ -105,7 +105,10 @@ func validateListAttr(schema *yang.Entry, value interface{}) util.Errors {
 	if size < schema.ListAttr.MinElements {
 		errors = util.AppendErr(errors, fmt.Errorf("list %s contains fewer than min required elements: %d < %d", schema.Name, size, schema.ListAttr.MinElements))
 	}
-	if size > schema.ListAttr.MaxElements {
+	// 0 is an invalid value for MaxElements
+	// (https://tools.ietf.org/html/rfc7950#section-7.7.6).
+	// For useability it best represents the value "unbounded".
+	if schema.ListAttr.MaxElements != 0 && size > schema.ListAttr.MaxElements {
```
robshakir, line 111: Is there a reason that we shouldn't just error here?

robshakir (changes requested):

ytypes/list_test.go
wenovus: The concern is for any derived package which tries to instantiate a `yang.Entry` where `ListAttr` is needed, then the user may run into the issue where the zero value is not useful, and they have to go and change every place where they instantiate a `yang.Entry` type.

See https://golang.org/doc/effective_go.html#allocation_new
"""
Since the memory returned by new is zeroed, it's helpful to arrange when designing your data structures that the zero value of each type can be used without further initialization. This means a user of the data structure can create one with new and get right to work.
"""

ytypes/util_schema.go
wenovus: See above, I realized that this might create a useability issue.

robshakir (approved):
Ah, makes sense! Thanks for explaining.

PR #481, opened 2020-12-03
Improve union enum testing

wenovus:
- Reorder enumeration tests on union enums into a consecutive test block for easier reading.
- Add new integration test for Go and proto to verify how union enums are named.


robshakir (approved):
Thanks for adding this testing. LGTM.

PR #482, opened 2020-12-03
Update tests in demo package due to recent model changes in AFT

wenovus:
Also improve test output -- using the diff was confusing because the updates are not ordered.


robshakir (approved):
Thanks for the change!

PR #483, opened 2020-12-03
Add flag for consistent "Enum" suffix for all inlined enumerations within unions

wenovus:
`-enum_suffix_for_simple_union_enums`:
If set to true when typedef_enum_with_defmod is also true, all inlined enumerations within unions will be suffixed with "Enum", instead of adding the suffix only for inlined enumerations within typedef unions.


generator/generator.go
```diff
@@ -72,6 +72,7 @@ var (
...
 	shortenEnumLeafNames                 = flag.Bool("shorten_enum_leaf_names", false, "If also set to true when compress_paths=true, all leaves of type enumeration will by default not be prefixed with the name of its residing module.")
 	useDefiningModuleForTypedefEnumNames = flag.Bool("typedef_enum_with_defmod", false, "If set to true, all typedefs of type enumeration or identity will be prefixed with the name of its module of definition instead of its residing module.")
+	appendEnumSuffixForSimpleUnionEnums  = flag.Bool("enum_suffix_for_simple_union_enums", false, "If set to true when typedef_enum_with_defmod is also true, all inlined enumerations within unions will be suffixed with \"Enum\", instead of adding the suffix only for inlined enumerations within typedef unions.")
```
robshakir, line 75: At this point, I'm wondering whether we just deprecate all these in v1.0.0 and cut this a bit further down the line (Q2 '21 maybe?). It seems like we can get rid of all three of these flags then.

robshakir: (no action required)

robshakir (approved):
LGTM, thanks for weeding out these inconsistencies.

generator/generator.go
wenovus: Yep that's what I'm thinking too -- this is not only complexity for the users but also a maintenance burden.

PR #485, opened 2020-12-08
Correctly reject leaves on a non-leaf node during JSON unmarshalling

wenovus:
The trie-building and checkTree had a bug that cleverly escaped testing.
Adding a test for a path that's a single element shorter instead of two shorter exposed the problem.


robshakir (approved):
Well caught! Thanks for catching this!

PR #490, opened 2020-12-15
Fix AppendErr statements

wenovus:
Fix misusages of `AppendErr` function.


robshakir (approved):
Thanks for this, I think we made this act like `append` and then ignored the fact that it would! Oops!

Issue #492, opened 2021-01-11
Dealing with Overlapping Slices when calling `MergeStruct`


wenovus:
@robshakir pointed out that an issue here is in RFC7950, non-config leaf-lists now allow duplicates, making the behaviour more complicated for `MergeStruct`.

I see 3 possible behaviours for a particular leaf-list node:

### Case 1: RFC6020 leaf-list or RFC7950 config leaf-list (duplicates disallowed)
- If leaf-lists are identical, then use the value. This is what we do for other field types: scalar, pointer, binary, or union. Currently we throw an error, which is problematic.
- If leaf-lists are overlapping, then erroring out is correct.

### Case 2: RFC7950 non-config leaf-list (duplicates allowed)
Simply concatenate the leaf-lists.

### Case 3: RFC7950 non-config leaf-list (duplicates allowed, but not desired by user)
Concatenate and uniquify the merged leaf-list.

-----

Since having or not having a path "config/*" within the path tag can be used to determine whether a field is config or state (correct me if I'm wrong), I think each of the three cases can be readily implemented.

The problem is I'm not sure how to actually decide what combinations of these behaviours to offer to the user. It seems a config node can follow either 1 or 2, and non-config nodes can behave according to any of the 3 states, giving 5 unique combinations.

I feel we should just implement one of these, and if users would like other behaviours, they can add them one-by-one under a `MergeOpt`.

robshakir:
The `config` element in the path isn't something that can be relied on -- since the merge needs to be able to deal with non-OpenConfig elements.

I think that you're right that we should implement one of these options as you say -- and then we can figure out what the right approach is in the PR.

PR #493, opened 2021-01-11
Skip merging identical slices instead of erroring

wenovus:
Currently, merging identical slices when calling `MergeStruct*` causes an error, complaining that the slices aren't unique per RFC6020.

This behaviour is inconsistent with what we do for unions, scalars, pointers, and binary, where we skip merging, and simply keep the identical value.

This PR implements this more consistent behaviour, which is the (1, 1) behaviour combination described in #492.


wenovus:
@robshakir note I didn't add a `MergeOpt` here, because I currently think skipping the merge makes more sense as the default behaviour than erroring out, although I do see the argument that people might assume a behaviour that's not what they intended, where having the `MergeOpt` would make sense.

As the PR currently stands, I think we have a good support for the base case of RFC6020. Special de-duping behaviour, or RFC7950 support can be added with `MergeOpt` in the future.

ygot/struct_validation_map.go
```diff
@@ -820,6 +820,10 @@ func copySliceField(dstField, srcField reflect.Value, opts ...MergeOpt) error {
 	}
 
 	if _, ok := srcField.Interface().([]Annotation); !ok {
+		if diff := cmp.Diff(srcField.Interface(), dstField.Interface()); diff == "" {
```
robshakir, line 823: Does `cmp.Diff` do something different to `cmp.Equal` here? I can't remember why we used `cmp.Diff` originally.

robshakir (approved):
LGTM, thanks for this change -- I think you're right that things brings things back to what we'd expect for equal fields. (I think the original reason that we had an error here was that we expected that values *would not* be populated in both cases, and rather we'd error if the field was populated in both -- but slices weren't covered when the fix to allow for equal values was allowed).

robshakir:
Actually, one other thing -- can you update the docstring [here](https://godoc.corp.google.com/pkg/google3/third_party/golang/ygot/ygot/ygot#MergeStructs) since I think we need to say that if the slices are equal, then they are not merged.

ygot/struct_validation_map.go
wenovus: https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff
```
It returns an empty string if and only if Equal returns true for the same input values and options.
```

robshakir: Given that you don't use the `diff` here, I suggest just changing this to be `if !cmp.Equal()`.

PR #498, opened 2021-02-05
Fix Unsupported in Go Unions, Fix singleton union default value generation, add tests for Go integration tests

wenovus:


.github/workflows/go.yml
```diff
@@ -67,6 +66,42 @@ jobs:
...
+            name=$(basename $1)
+            cp "$1" "$tmpdir"/"${name}.go"
+            cd $tmpdir
+            goimports -w *.go
+            go build
+            cd $cwd
+            rm -r $tmpdir
+          }
+
+          expectFail=(
```
robshakir, line 84: Why do we expect that these fail? If we're outputting anything shouldn't it compile?

robshakir (approved):
One question on this, otherwise it LGTM.

.github/workflows/go.yml
wenovus: Thanks for mentioning this, I've added comments detailing the reasons in the CI script.

robshakir: In both the cases below, should we just fix these? It seems like:

* the unsupported symbol shouldn't be output in the code
* the default unions case should return an error (or produce code that compiles - with a caveat that says the default isn't supported)

robshakir (changes requested):
Reverting LGTM whilst we agree on whether code should compile.

wenovus:
Recent changes fixes the bug introduced in #496 where default value for singleton unions were not generated correctly.

.github/workflows/go.yml
wenovus: Fixed.

wenovus:
@robshakir PTAL, all substantial recent changes are in this commit: https://github.com/openconfig/ygot/pull/498/commits/7bbefe14a7cf2e8d8af0379c386e6cdd094bc239
- Address corner case of default value generation for singleton unions.
- Fail if user provides YANG file with a default value for a non-singleton union (https://github.com/openconfig/ygot/pull/498/commits/7bbefe14a7cf2e8d8af0379c386e6cdd094bc239#diff-b75c80aad7fc147614bd565f03ab0194af2b76410e8862d976615f92355845df)
- Use a unicode letter to make an integration test's output compilable.

robshakir (approved):

PR #507, opened 2021-03-12
[ygot merge] copyMapField and copySliceField updates dst in place instead of creating a new one. This change also fixes a bug that copyMapField is not overwriting dst.

chentianheng604:


chentianheng604:
@googlebot I signed it!

PR #509, opened 2021-03-12
[ygot merge] copyMapField and copySliceField updates dst in place instead of creating a new one. This change also fixes a bug that copyMapField is not overwriting dst.

chentianheng604:


ygot/struct_validation_map.go
```diff
@@ -741,39 +741,26 @@ func copyMapField(dstField, srcField reflect.Value, opts ...MergeOpt) error {
...
-			// into.
-			v := m.field.MapIndex(k)
-			var d reflect.Value
-			var ok bool
-			if d, ok = existingKeys[k.Interface()]; !ok {
-				d = reflect.New(v.Elem().Type())
-				existingKeys[k.Interface()] = v
-			}
+	if dstField.Len() == 0 {
+		dstField.Set(reflect.MakeMapWithSize(reflect.MapOf(m.key, m.value), 0))
```
wenovus, line 745: nit: srcField.Len() instead of 0.

wenovus (approved):
This looks fine to me, just one nit.

@robshakir, Tianheng measured this and found that this can dramatically improve performance. Since `MergeStructs` first does a `DeepCopy` of the destination struct prior to merge, I don't see any reasons to not do this. To double check, I looked at the code history, and I'm reasoning that the current behaviour could be an artifact of the original code rejecting a merge when the dst slice or map was set at all, and so created the empty map/slice (#64: https://github.com/openconfig/ygot/pull/64/files#diff-53ea16782fc8ade8cfb72e932418546cafff36de50091bd2ceced0d8594cd085R486, https://github.com/openconfig/ygot/pull/64/files#diff-53ea16782fc8ade8cfb72e932418546cafff36de50091bd2ceced0d8594cd085R455), with the subsequent changes retaining the behaviour (#177, #81) instead of optimizing them.

ygot/struct_validation_map.go
chentianheng604: Done.

robshakir (approved):
This LGTM.

@wenovus - absolutely agree that this was the reason for this code originally, we expected that any field that was specified in the target to be merged into would be skipped, so we had to create the map, and I think we overlooked the cost of creating and merging here.

Thanks for the contribution, review and for fixing the bug on the merge!

PR #510, opened 2021-03-15
CI and Makefile improvements

dmmarkov:
Hello!

I made some improvements in Makefile and CI which helps me test my changes in local env and inside of GitHub in the same way.

I hope it will be useful.


dmmarkov:
@googlebot I signed it!

demo/protobuf_getting_started/update.sh
```diff
@@ -1,21 +1,20 @@
-#!/bin/bash
-
-clean() {
```
robshakir, line 3: What's the logic here not to clean up the deps and public? It generally seems useful to do so.

robshakir (commented):
@dmmarkov -- thanks for cleaning this up :-), it looks good to me.

@wenovus - PTAL, I don't see a problem here, but would appreciate your review too. Will leave final LGTM to you.

.github/workflows/go.yml
```diff
@@ -11,11 +11,11 @@ on:
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
dmmarkov: latest is flow label. Today it maybe one image, but tomorrow it maybe another. Using the most precise label is the best practice.

demo/protobuf_getting_started/update.sh
dmmarkov: I deleted it because I don't see any `public` or `deps` folder inside of `demo/protobuf_getting_started`.

Makefile
```diff
@@ -2,11 +2,17 @@
 #
 # This makefile is used by Travis CI to run tests against the ygot library.
```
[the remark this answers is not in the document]
dmmarkov: Yes. I changed it.

go.mod
```diff
@@ -5,15 +5,15 @@ go 1.14
 require (
 	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
 	github.com/golang/protobuf v1.4.2
-	github.com/google/go-cmp v0.4.1
+	github.com/google/go-cmp v0.5.0
 	github.com/kylelemons/godebug v1.1.0
 	github.com/openconfig/gnmi v0.0.0-20200508230933-d19cebf5e7be
 	github.com/openconfig/goyang v0.2.2
```
wenovus, line 11: Since you're already upgrading some dependencies, could you run `go get -u ./...` in ygot to update dependencies such as goyang as well?

demo/protobuf_getting_started/update.sh
```diff
@@ -1,21 +1,20 @@
...
+
+if [ -z ${GO_PATH} ]; then
+  echo "no GOPATH defined!!!"
 
-clean
+  exit 1
+fi
+
+GO111MODULE=off go get -u github.com/openconfig/ygot || :
+GO111MODULE=off go get -u github.com/google/protobuf || :
```
wenovus, line 17: "Go 1.17 will ignore GO111MODULE": (https://blog.golang.org/go116-module-changes).

What's the reason why we should have this environment variable set?

wenovus (commented):
LGTM, couple small suggestions.

PR #511, opened 2021-03-19
Consistently use goEnumPrefix and fix bug generating enum default value in wrapper union node.

wenovus:
I introduced a bug in #496 (https://github.com/openconfig/ygot/pull/496/files#diff-19ecb3e61250b63aaf93a11d7017203a90637b52bee811a021ab253b01c42748L275-L277) where the default value for wrapper unions was no longer being generated correctly for enumerated values when the simple union flag is unset.

To fix this, `yangTypeToGoType` now consistently **does not** call `enumDefaultValue`, and `goLeafDefault` consistently **does** call `enumDefaultValue` when generating the default value for typedef enums.

Also `goEnumPrefix` is now always used when generating enum names instead of prepending `"E_"`.


wenovus:
@robshakir, providing more details on this change since my initial comment wasn't quite clear (the only code changes for this PR are in [goelements.go](https://github.com/openconfig/ygot/pull/511/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yang#diff-19ecb3e61250b63aaf93a11d7017203a90637b52bee811a021ab253b01c42748) and [gogen.go](https://github.com/openconfig/ygot/pull/511/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yang#diff-b75c80aad7fc147614bd565f03ab0194af2b76410e8862d976615f92355845df)):

### Background
The recent union default value generation changes [removed a call to `enumDefaultValue`](https://github.com/openconfig/ygot/pull/496/files#diff-19ecb3e61250b63aaf93a11d7017203a90637b52bee811a021ab253b01c42748L275-L277) in order to use the *raw* default value string provided in the YANG file to look up the existence of the enum value in goyang's AST before accepting the default value as an enum type. The `yangDefaultValueToGo` function only processes these raw values for simplicity. However, this breaks generating typedef enums in wrapper union mode because a call to `enumDefaultValue` is required to make the output correct.

### This change
What this change does is that during the processing of both inline and typedef enumerated values by `yangTypeToGoType`, `enumDefaultValue` is **not called** when populating `MappedType`'s `DefaultValue` field. This means that **all** enumerated values that subsequently get processed by `yangDefaultValueToGo`, which come after the call to `yangTypeToGoType`, will see the raw default value. Most importantly, `goLeafDefault`, a call that happens in wrapper-union mode, can now transform the raw default values for enums into their "cooked" versions without discriminating between inline and typedef enumerated values. This is what fixes this problem.

robshakir (approved):
Thanks for the clear explanation Wen, it made it much easier to review this code!

LGTM - thanks for fixing this.

PR #515, opened 2021-04-07
Cache regexp compilations during string validation.

wenovus:
If there are a lot of routing prefixes to be validated, regexp compilation can become a performance bottleneck.

Aside: This could be because Go's regexp implementation might be relatively slow compared to other languages: https://github.com/golang/go/issues/11646

### DIFF
--- a/ytypes/schema_tests/posix_regex_benchmark_test.go
+++ b/ytypes/schema_tests/posix_regex_benchmark_test.go
@@ -24,10 +24,10 @@ import (
 func BenchmarkPOSIXPattern(b *testing.B) {
 	d := &oc.Device{}
 	prefixSet := d.GetOrCreateRoutingPolicy().GetOrCreateDefinedSets().GetOrCreatePrefixSet("foo")
-	for i := 0; i != 100; i++ {
+	for i := 0; i != 256; i++ {
 		prefixSet.NewPrefix(fmt.Sprintf("%d.%d.%d.0/24", i, i, i), "exact")
 	}
-	for i := 0; i != 100; i++ {
+	for i := 0; i != 256; i++ {
 		prefixSet.NewPrefix(fmt.Sprintf("FFFF:%d:EEEE:AAAA::/64", i), "60..64")
 	}
 
--- a/ytypes/schema_tests/posix_regex_benchmark_test.go
+++ b/ytypes/schema_tests/posix_regex_benchmark_test.go
@@ -1,0 +1,40 @@
+// Copyright 2021 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package validate
+
+import (
+	"fmt"
+	"testing"
+
+	oc "github.com/openconfig/ygot/exampleoc"
+)
+
+func BenchmarkPOSIXPattern(b *testing.B) {
+	d := &oc.Device{}
+	prefixSet := d.GetOrCreateRoutingPolicy().GetOrCreateDefinedSets().GetOrCreatePrefixSet("foo")
+	for i := 0; i != 100; i++ {
+		prefixSet.NewPrefix(fmt.Sprintf("%d.%d.%d.0/24", i, i, i), "exact")
+	}
+	for i := 0; i != 100; i++ {
+		prefixSet.NewPrefix(fmt.Sprintf("FFFF:%d:EEEE:AAAA::/64", i), "60..64")
+	}
+
+	b.ResetTimer()
+	for i := 0; i != b.N; i++ {
+		if err := d.Validate(); err != nil {
+			b.Fatalf("d.Validate() failed: %v", err)
+		}
+	}
+}
--- a/ytypes/string_type.go
+++ b/ytypes/string_type.go
@@ -27,55 +27,73 @@ import (
 
 // Refer to: https://tools.ietf.org/html/rfc6020#section-9.4.
 
-var (
-	// posixRegexCache stores previously-compiled Regexp objects.
-	// This helps the performance of validation of, say, a large prefix
-	// list that have the same pattern specification.
-	//
-	// Concurrency Requirements
-	//
-	// Only the regexp cache has to be protected by mutexes, since
-	// a Regexp is safe for concurrent use by multiple goroutines:
-	// https://golang.org/src/regexp/regexp.go
-	posixRegexCache = map[string]*regexp.Regexp{}
-	posixRegexMutex = &sync.RWMutex{}
-	re2RegexCache   = map[string]*regexp.Regexp{}
-	re2RegexMutex   = &sync.RWMutex{}
-)
+// reCache is the global regexp cache used for speeding up the validation of
+// pattern-restricted strings.
+var reCache = newRegexpCache()
+
+// regexpCache stores previously-compiled Regexp objects.
+// This helps the performance of validation of, say, a large prefix
+// list that have the same pattern specification.
+//
+// Concurrency Requirements
+//
+// Only the regexp cache map has to be protected by mutexes, since
+// a Regexp is safe for concurrent use by multiple goroutines:
+// https://golang.org/src/regexp/regexp.go
+type regexpCache struct {
+	posixMu sync.RWMutex
+	posix   map[string]*regexp.Regexp
+
+	re2Mu sync.RWMutex
+	re2   map[string]*regexp.Regexp
+}
+
+// newRegexpCache returns a regexpCache with all fields in a useable, empty
+// state.
+func newRegexpCache() *regexpCache {
+	return &regexpCache{
+		posix: map[string]*regexp.Regexp{},
+		re2:   map[string]*regexp.Regexp{},
+	}
+}
 
 // compilePattern returns the compiled regex for the given regex
 // pattern. It caches previous look-ups for faster performance.
 // Go's regexp implementation might be relatively slow compared to other
 // languages: https://github.com/golang/go/issues/11646
-func compilePattern(pattern string, isPOSIX bool) (*regexp.Regexp, error) {
-	regexCache := re2RegexCache
-	regexMutex := re2RegexMutex
+func (c *regexpCache) compilePattern(pattern string, isPOSIX bool) (*regexp.Regexp, error) {
+	regexCache := c.re2
+	regexMutex := &c.re2Mu
 	regexCompile := regexp.Compile
 	if isPOSIX {
-		regexCache = posixRegexCache
+		regexCache = c.posix
+		regexMutex = &c.posixMu
 		regexCompile = regexp.CompilePOSIX
-		regexMutex = posixRegexMutex
 	}
 
-	regexMutex.RLock()
-	r, ok := regexCache[pattern]
-	regexMutex.RUnlock()
-	if !ok {
-		var err error
-		r, err = regexCompile(pattern)
-		if err != nil {
-			return nil, err
-		}
-		// It's true that there may be multiple writers into the map at
-		// the same time. This, however, doesn't impact correctness,
-		// since any compiled Regexp objects are acceptable, and
-		// displaced objects are simply dropped at the end of this
-		// function.
-		regexMutex.Lock()
-		regexCache[pattern] = r
-		regexMutex.Unlock()
+	// Attempt to read a previously cached regexp.Regexp.
+	if re := func() *regexp.Regexp {
+		regexMutex.RLock()
+		defer regexMutex.RUnlock()
+		return regexCache[pattern]
+	}(); re != nil {
+		return re, nil
 	}
-	return r, nil
+
+	// Read unsuccessful (cache-miss). Compile and populate the cache.
+	re, err := regexCompile(pattern)
+	if err != nil {
+		return nil, err
+	}
+	// Multiple unsuccessful readers might try to populate their own
+	// compiled regexp.Regexp objects into the same cache entry.
+	// This is ok, as since any regexp.Regexp value for the same cache
+	// entry is equivalent, it does not matter which compiled instance is
+	// introduced into the map, or returned to the caller.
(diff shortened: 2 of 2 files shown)


ytypes/string_type.go
```diff
@@ -26,6 +26,35 @@ import (
 
 // Refer to: https://tools.ietf.org/html/rfc6020#section-9.4.
 
+var (
+	POSIXregexCache = map[string]*regexp.Regexp{}
```
robshakir, line 30: Is it right to have this as a global -- I think that this might bring us some challenges if someone were to do something like:

```golang
d := &oc.Device{}
e := &oc.Device{}

go func() {
   _ = d.Validate()
}()

go func() {
    _ = e.Validate()
}()
```

In this case, you have two goroutines that are potentially changing the global map at the same time.

Should we rather have this such that there is some validation context that is passed between different validations within the same call? Or, alternatively, have it such that we create the cache as part of the generated package, and protect it with a mutex?

ytypes/string_type.go
```diff
@@ -26,6 +26,35 @@ import (
...
+	RE2regexCache   = map[string]*regexp.Regexp{}
+)
+
+// compilePattern returns the compiled regex for the given regex
+// pattern. It caches previous look-ups for faster performance.
+// Go's regexp implementation might be relatively slow compared to other
+// languages: https://github.com/golang/go/issues/11646
+func compilePattern(pattern string, isPOSIX bool) (*regexp.Regexp, error) {
```
robshakir, line 38: For discussion -- since I thought about it when reviewing. Is there ever a case where we have the same <pattern, isPOSIX> tuple, but don't have the same regexp result -- such that we need to cache based on the path instead? I don't think so - since the result of a check against that regexp (`Match`) would always be the same but I figured I'd just raise it to figure out whether there's something else we need to consider here.

robshakir, line 38: Is this function adequately tested from the unit tests that we have today? I presume so, but just wanted you to take a look and confirm.

ytypes/string_type.go
robshakir: Thinking about this - it's a bit harder to take it from the generated code whilst keeping the same function signatures that we have today - but we could add a `ValidationOpt` that covers this. So perhaps I'd suggest that this could be something that is created in the first container validation and then handed off -- this would mean that we perhaps don't need the same concurrency protection because there'll be a single context within which we're doing a validation - given that we do not parallelise the validation today.

robshakir (changes requested):
Thanks for the change Wen. I'm not quite sure about the design here, so PTAL.

I'd also like to see whether we have some benchmarking that captures what the performance improvement is so that we can keep an eye on regressions.

ytypes/string_type.go
robshakir: On a second glance, this is why the race tests are failing AFAICS.

ytypes/string_type.go
wenovus: My thought is that this shouldn't a be problem even if it were the case. The behaviour should be the same even if the output is nondeterministic, right?

robshakir: I think that's right, yes.

ytypes/string_type.go
wenovus: So I see that you're concerned about memory usage by the cache, and that we don't want to have the program store compiled version of the regexes for another part of the schema when we might now be working with a different part?

My initial thought was that since the generated code already unzips the schema, that a proportional increase in background memory use would not likely cause a problem.

So, could we just keep the global caches in `ytypes` and just introduce a mutex to solve the race condition?

----

I understand you to mean that we introduce a new type `StringValidationOpt` that would store the caches, and it's only an argument to `ValidateStringRestrictions`. The other validation functions always invokes this option when calling `ValidateStringRestrictions`? I'm fine with this, although I would prefer to not introduce this flag. I'm not quite convinced having a global cache would be so bad.

If instead, we introduce a `ValidationOpt` behaviour that's not default for all other validation functions, then it's conceivable the user to experience the following:
1. Don't use it initially since it takes some reading to become aware of it, and then some thought to figure out whether it applies to their situation since most want to minimize their code complexity.
2. Notice a performance problem (at which time it's already too late).
3. Debug their performance problem.
4. Look at the docs to find that this applies to their situation, and then uses it.

----

To make this easier for consuming packages like `ygot`, we can optionally move the regex compilation and caching down one level to `goyang` by introducing a new exported slice variable for each pattern type:
```
type YangType struct {
...
	Pattern          []string
	POSIXPattern     []string
        CompiledPatterns []*regexp.Regexp
        CompiledPOSIXPatterns []*regexp.Regexp
}
```
+ backwards compatible change without needing a new `ValidationOpt` of any kind.
+ Caching is preserved across `Validate` calls.
- Regexes always get compiled, so will incur a performance cost when constructing the AST. The user may then not ever use them. <- my guess is this is not a practical problem since the user almost always wants the patterns to be compiled as is.

However, I wonder, even though Go is backwards compatible, if the `*regexp.Regexp` were serialized and then deserialized by a different Go version, could it break?

I don't find this solution to be practically much different than using a global cache, however.

ytypes/string_type.go
wenovus: Yes I believe so. The current tests for `ValidateStringRestrictions` verifies many regexes, and some of the tests use the same regex, which tests the caching code. Although there isn't a mock to verify the caching behaviour.

ytypes/string_type.go
robshakir: I'm not concerned about memory usage, I'm concerned about whether the map is safe across different goroutines, because we currently don't have anything that is outside of the context of a single validation call today.  We could introduce mutex-protected global caches in the `ytypes` package for sure - but what I was thinking is whether we want to do this in the context of a particular validation such that we don't end up with locking in between different validation calls -- how we'd do this would be something that we can figure out if we agree what the scope should be. (I understand the downsides with a `ValidationOpt` - I was just discussing the different ways that we might change away from the global map.)

On the approach of moving the compilation into goyang -- I'm not clear that this is the right thing to do. It introduces a new dependency for all consumers of this library - whereas only a few might want to do this. However, one thing we could do is move to something like this when parsing the regexps.

wenovus: Ok I see. Currently, there isn't any practical difference between using a global cache vs. keeping the context within a single `Validate` call. I missed this and so assumed you meant space, apologies.

Locking and unlocking might introduce performance issues, since reading also requires locking, so using a global cache might itself cause a performance problem. I'll work on a benchmark test to help resolve this question.

robshakir: I think that using a global cache has some upsides though, right? Since you don't need to call `Compile`/`CompilePOSIX` more than exactly once per regexp in the model set that is compressed in the generated package. So there's some savings in CPU time during building the map (which I think we're saying is the slow thing today).

Building the cache once per `Validate` is 'safer' in terms of not having potential lock contention if this is done in parallel, but it then means we spend CPU time compiling regexps multiple times - the number of times that `Validate` is called.

If we consider a use case of:

 - Generate 1,000 `oc.Interface{}`, and validate each of them in turn.
 - Merge these interfaces into an `oc.Device{}` and then validate this.

which is something that we know we have - then it seems like we're better to have a global cache, since this will mean just one compile for each regexp vs. 1001 that would be needed in the following case.

Further to this, I was comparing this to what we did for the leafref case - and concluded that they are different (I'd originally considered them quite equivalent). In the leafref case we are caching results based on a lookup _in a particular data tree_, but in this case, we're caching results based on _a particular schema tree_, so it makes sense that it's generally OK to store this as a global for the `ytypes` package.

chentianheng604: FYI. The ZTC use case is:

1. Config gen of multiple entities are run in parallel.
2. For each config gen, we only validate the final OC device.

Therefore, from ZTC perspective, I would prefer per-Validate cache. If we implement global cache with mutex, we essentially serialize the validation calls, which could affect the performance of metroshard config gen (as metroshard is expanded into EK_NETWORK + mutiple EK_DEVICE in ZTC). On the other hand, since ZTC has stopped validating each OC piece (another performance improvement one year ago), the downside of per-Validate cache does not matter that much.

wenovus:
790c26b's benchmark test was CPU-profiled using "runtime/pprof", and I see these results for the relevant functions:

The reason why I kept n small for the benchmark tests is that I found other functions (i.e. reflect's MapKeys()) was scaling even worse than the extra compile operations. In order to see a sizeable difference in the benchmark test, I chose a smaller number (too small also doesn't work).

### without optimization
```
BenchmarkPOSIXPattern-12              13          89135241 ns/op
PASS
ok      github.com/openconfig/ygot/ytypes/schema_tests  4.972s
```
#### Running with n=5000 for ipv6-prefix to get a good pprof profile:
```
Duration: 3.64s, Total samples = 5.92s (162.48%)
flat  flat%   sum%        cum   cum%
         0     0% 90.20%      1.08s 18.24%  regexp.CompilePOSIX (inline)
         0     0% 90.20%      1.08s 18.24%  regexp.compile
         0     0% 90.20%      0.62s 10.47%  github.com/openconfig/ygot/ytypes.ValidateStringRestrictions
         0     0% 90.20%      1.12s 18.92%  github.com/openconfig/ygot/ytypes.validateString
         0     0% 90.20%      0.49s  8.28%  github.com/openconfig/ygot/ytypes.validateStringSchema
(*Regexp).MatchString` didn't appear as a profile entry.
```

### with optimization
```
BenchmarkPOSIXPattern-12              19          57036685 ns/op
PASS
ok      github.com/openconfig/ygot/ytypes/schema_tests  4.865s
```
#### Running with n=5000 for ipv6-prefix to get a good pprof profile:
```
Duration: 2.33s, Total samples = 3.44s (147.72%)
         0     0% 92.44%      0.02s  0.58%  runtime.lock (inline)
         0     0% 92.44%      0.02s  0.58%  runtime.lockWithRank (inline)
         0     0% 92.44%      0.05s  1.45%  github.com/openconfig/ygot/ytypes.ValidateStringRestrictions
         0     0% 92.44%      0.05s  1.45%  github.com/openconfig/ygot/ytypes.validateString
`validateStringSchema` didn't appear as a profile entry.
         0     0% 92.44%      0.03s  0.87%  regexp.(*Regexp).MatchString (inline)
```

From the pprof results, it appears the locking operations (in the absence of contention), does not add any significant overhead as compared with the validation operations themselves, which take much less time than the unoptimized version.

So, I think the global cache makes sense as an implementation as-is.

ytypes/string_type.go
wenovus: PTAL, I've added the mutex and a benchmark test. Also see comment below.

ytypes/schema_tests/posix_regex_benchmark_test.go
```diff
@@ -0,0 +1,40 @@
+// Copyright 2021 Google Inc.
```
robshakir, line 1: (nit) Is this too specific of a test file name? Should we just put this in the existing test files?

ytypes/string_type.go
```diff
@@ -26,6 +27,57 @@ import (
...
+	// posixRegexCache stores previously-compiled Regexp objects.
+	// This helps the performance of validation of, say, a large prefix
+	// list that have the same pattern specification.
+	//
+	// Concurrency Requirements
+	//
+	// Only the regexp cache has to be protected by mutexes, since
+	// a Regexp is safe for concurrent use by multiple goroutines:
+	// https://golang.org/src/regexp/regexp.go
+	posixRegexCache = map[string]*regexp.Regexp{}
```
robshakir, line 40: I'd make these a type:

```golang
type regexpCache struct {
   posixMu sync.RWMutex 
   posix map[string]*regexp.Regexp
   
   re2Mu sync.RWMutex
   re2 map[string]*regexp.Regexp
}
```

This seems cleaner to me in terms of it being clearer what the mutexes are designed to guard (similar to the comment made w.r.t mutexes in [Effective Go](https://golang.org/doc/effective_go#commentary)).

ytypes/string_type.go
```diff
@@ -26,6 +27,57 @@ import (
...
+	regexCompile := regexp.Compile
+	if isPOSIX {
+		regexCache = posixRegexCache
+		regexCompile = regexp.CompilePOSIX
+		regexMutex = posixRegexMutex
+	}
+
+	regexMutex.RLock()
+	r, ok := regexCache[pattern]
+	regexMutex.RUnlock()
```
robshakir, line 62: Prefer to use `defer` here so that we don't end up changing this code in some way that gets us in lock contention. Here I think the usual style is to do this through an inline-function:

```golang
r := func() *regexp.Regexp {
   regexpMutex.RLock()
   defer regexpMutex.RUnlock()
   return regexpCache[pattern]
}()

if r != nil {
   return r, nil
}

re, err := regexpCompile(pattern)
if err != nil {
 return nil, err
}
regexpMutex.Lock()
defer regexpMutex.Unlock()
regexpCache[pattern] = re

return re, nil 
```

ytypes/string_type.go
```diff
@@ -26,6 +27,57 @@ import (
...
+	if !ok {
+		var err error
+		r, err = regexCompile(pattern)
+		if err != nil {
+			return nil, err
+		}
+		// It's true that there may be multiple writers into the map at
+		// the same time. This, however, doesn't impact correctness,
```
robshakir, line 70: There aren't multiple writers at the same time given the mutex right? I think we mean here "It does not matter which instance of pattern is introduced into the map, since the compiled regexp.Regexp are equal. If we reach this codepath and another writer has already introduced the pattern to the map whilst we waited to acquire the lock, it will be overwritten with identical contents."

robshakir (changes requested):

ytypes/schema_tests/posix_regex_benchmark_test.go
wenovus: My thought is it makes it clear where the benchmark tests are, because the other benchmark files are currently also very specific:

diff_benchmark_test.go
leafref_benchmark_test.go

ytypes/string_type.go
wenovus: Done.

ytypes/string_type.go
wenovus: Done.

ytypes/string_type.go
wenovus: Done with some tweaks.

wenovus:
f96e04f has the same performance:
BenchmarkPOSIXPattern-12              13          89388300 ns/op
PASS
ok      github.com/openconfig/ygot/ytypes/schema_tests  5.033s

robshakir (approved):

robshakir:
Thanks for all the changes, LGTM :-)

wenovus:
Thanks for the instructive review!

PR #518, opened 2021-04-20
Implement TogNMISetRequest render function

sthesayi:
This function is similar to TogNMINotifications but returns a SetRequest proto message instead.

### DIFF
--- a/ygot/render.go
+++ b/ygot/render.go
@@ -298,9 +298,9 @@ func (g *gnmiPath) StripPrefix(pfx *gnmiPath) (*gnmiPath, error) {
 	return newPathElemGNMIPath(g.pathElemPath[len(pfx.pathElemPath):]), nil
 }
 
-// GNMINotificationsConfig specifies arguments determining how the
+// GNMIConfig specifies arguments determining how the
 // gNMI output should be created by ygot.
-type GNMINotificationsConfig struct {
+type GNMIConfig struct {
 	// UsePathElem specifies whether the elem field of the gNMI Path
 	// message should be used for the paths in the output notification. If
 	// set to false, the element field is used.
@@ -317,6 +317,10 @@ type GNMINotificationsConfig struct {
 	PathElemPrefix []*gnmipb.PathElem
 }
 
+// GNMINotificationsConfig specifies arguments determining how the
+// gNMI output should be created by ygot.
+type GNMINotificationsConfig = GNMIConfig
+
 // TogNMINotifications takes an input GoStruct and renders it to slice of
 // Notification messages, marked with the specified timestamp. The configuration
 // provided determines the path format utilised, and the prefix to be included
@@ -327,7 +331,6 @@ type GNMINotificationsConfig struct {
 // abstraction. It can also be refactored to simply use the findSetleaves function
 // which has a cleaner implementation using the reworked iterfunction util.
 func TogNMINotifications(s GoStruct, ts int64, cfg GNMINotificationsConfig) ([]*gnmipb.Notification, error) {
-
 	var pfx *gnmiPath
 	if cfg.UsePathElem {
 		pfx = newPathElemGNMIPath(cfg.PathElemPrefix)
@@ -346,6 +349,25 @@ func TogNMINotifications(s GoStruct, ts int64, cfg GNMINotificationsConfig) ([]*
 	}
 
 	return msgs, nil
+}
+
+// TogNMISetRequest takes an input GoStruct and renders it to a single
+// SetRequest message. The configuration provided determines the path format
+// utilised, and the prefix to be included in the message if relevant.
+func TogNMISetRequest(s GoStruct, cfg GNMIConfig) (*gnmipb.SetRequest, error) {
+	var pfx *gnmiPath
+	if cfg.UsePathElem {
+		pfx = newPathElemGNMIPath(cfg.PathElemPrefix)
+	} else {
+		pfx = newStringSliceGNMIPath(cfg.StringSlicePrefix)
+	}
+
+	leaves := map[*path]interface{}{}
+	if err := findUpdatedLeaves(leaves, s, pfx); err != nil {
+		return nil, err
+	}
+
+	return leavesToSetRequest(leaves, pfx)
 }
 
 // findUpdatedLeaves appends the valid leaves that are within the supplied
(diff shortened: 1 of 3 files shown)


ygot/render.go
```diff
@@ -298,9 +298,9 @@ func (g *gnmiPath) StripPrefix(pfx *gnmiPath) (*gnmiPath, error) {
 	return newPathElemGNMIPath(g.pathElemPath[len(pfx.pathElemPath):]), nil
 }
 
-// GNMINotificationsConfig specifies arguments determining how the
+// GNMIConfig specifies arguments determining how the
 // gNMI output should be created by ygot.
-type GNMINotificationsConfig struct {
+type GNMIConfig struct {
```
robshakir, line 303: I would create an alternate type here -- since the 'string slice' path is something that is to be deprecated in gNMI, so for the `Set` case, where we're just implementing this API I think we do not need to support this.

ygot/render.go
```diff
@@ -348,6 +351,25 @@ func TogNMINotifications(s GoStruct, ts int64, cfg GNMINotificationsConfig) ([]*
...
 
+// TogNMISetRequest takes an input GoStruct and renders it to a single
+// SetRequest message. The configuration provided determines the path format
+// utilised, and the prefix to be included in the message if relevant.
+func TogNMISetRequest(s GoStruct, cfg GNMIConfig) (*gnmipb.SetRequest, error) {
+	var pfx *gnmiPath
+	if cfg.UsePathElem {
+		pfx = newPathElemGNMIPath(cfg.PathElemPrefix)
+	} else {
+		pfx = newStringSliceGNMIPath(cfg.StringSlicePrefix)
```
robshakir, line 362: Per the above comment, let's not support the string slice path here.

ygot/render.go
```diff
@@ -348,6 +351,25 @@ func TogNMINotifications(s GoStruct, ts int64, cfg GNMINotificationsConfig) ([]*
...
+	}
+
+	leaves := map[*path]interface{}{}
+	if err := findUpdatedLeaves(leaves, s, pfx); err != nil {
```
robshakir, line 366: In a gNMI `SetRequest` we can send a request that is JSON marshalled from the path that we have - so we can just  encode the GoStruct at a specified path using `ygot.Marshal7951`. This is likely to be preferable since it means that we have a structured document that can be read through rather than all `<path, value>` pairs it contains.

ygot/render.go
```diff
@@ -348,6 +351,25 @@ func TogNMINotifications(s GoStruct, ts int64, cfg GNMINotificationsConfig) ([]*
 	return msgs, nil
 }
 
+// TogNMISetRequest takes an input GoStruct and renders it to a single
+// SetRequest message. The configuration provided determines the path format
+// utilised, and the prefix to be included in the message if relevant.
+func TogNMISetRequest(s GoStruct, cfg GNMIConfig) (*gnmipb.SetRequest, error) {
```
robshakir, line 357: In this case, we're saying that the `SetRequest` we're generating /only/ does `replace` - and not update. I think for a general purpose API like this we should probably think about whether we want to be able to allow the user to specify what kind of `SetRequest` they'd like to cover here.

There are a few options for doing this:

* We could take in a current and then new tree, and have the function calculate how to get between them (this is akin to doing something like `ygot.Diff`, but probably with subtree awareness) - this might be quite expensive.
* We could take in an input that maps a struct to a specified update/replace/delete method.
* We could support just a single method for all of the struct's contents.

In the case that we're implementing for -- I guess we need replace primarily. @wenovus is working on some other code corresponding to `Set` in ypathgen (or equivalent). @wenovus -- what are your use cases in that scenario, and can this API be commonly used across the two?

robshakir (changes requested):

ygot/render.go
wenovus: I have a couple questions on the PR:
1. The current implementation dismantles the GoStruct and serializes the leaves individually into `Update` messages. Since [doing a replace on a leaf path is the same as doing an update](https://github.com/openconfig/reference/issues/135#issuecomment-811172277), does this mean that for the current implementation, it doesn't matter which operation is being done?
2. Therefore, it seems to me that since we're using JSON anyways, it might make sense to call `ygot.EncodeTypedValue` on the GoStruct (@robshakir it would be great if you can confirm this does the same thing as `ygot.Marshal7951`), and just populate the `SetRequest` manually.

wenovus: 2. is exactly what I'm planning to do for my use case.

robshakir: @wenovus - yes, `EncodeTypedValue` with `enc = gnmipb.JSON_IETF` is equivalent to `Marshal7951`.

sthesayi: @wenovus If you are planning to do (2) then can I take the approach of "We could take in an input that maps a struct to a specified update/replace/delete method." This would make the signature more like this

func TogNMISetRequest(s GoStruct, req *gnmipb.SetRequest, cfg GNMIConfig) error

In the cfg object we can take an enum to say what field we are trying to populate. In my case I will just specify replace. We will take the approach of swapping all of the contents with new content from the GoStruct to keep things simple. This would put the onus on the caller to potentially call multiple times with different structs to populate all of the SetRequest. This API can be easily extended in future to take current SetRequest as additional input in the GNMISetRequestConfig and if this is not null then we will perform the expensive diff to figure out the delete, update, replace trees in the method. 

What do you think?

wenovus: Since you only require `replace` for your use case, I think it might be enough to simply populate the `SetRequest` manually. As a result, I'm wondering if you could do without this function? I'm thinking something like this:
```
          typedValue, err := ygot.EncodeTypedValue(goStruct, gpb.Encoding_JSON_IETF)
          if err != nil {
                  t.Fatalf("Could not encode value into JSON format")
          }

          sr := &gpb.SetRequest{
                  Replace: &gpb.Update{
                          Path: path, // I assume you have a way of getting this value.
                          Val:  typedValue,
                  },
          }
```
One potential issue with calling `TogNMISetRequest` multiple times to populate a `SetRequest` is that what if the user specifies a prefix in `GNMISetRequestConfig` that's different from the prefix already populated into the input `SetRequest` by a previous call to `TogNMISetRequest` with a different `GNMISetRequestConfig`? `TogNMISetRequest` would need to define a behaviour for this conflict. Not providing a helper forces the user to understand a little more of ygot's infrastructure and gNMI types, but I think the above snippet is simple to write.

As for doing the expensive diff, I'm wondering what the general use case for it is? From how I understand it, `ygot.Diff` computes an `update` operation (as oppose to `replace`), that's equivalent to just doing a `replace` without the diff. So, I think just doing a `replace` operation is sufficient.

sthesayi: @wenovus I am good with this approach. The tree walk to populate the SetRequest is not that complicated for just the replace tree population and we can skip the library function. I was shortsighted with my PR... To do this correctly, we want the library to be quite generic and there are complications as you suggested. Also there are some situations where I would need more fine grained control on the replace paths so I can create second/third level paths instead of just a top level "/" because we don't want to always replace the entire config but are also not able to generate a diff since the previous SetRequest is not available.

@robshakir If you are in agreement then I can close this PR and take the manual approach in my component.

robshakir: I tend to agree that if this function is just a wrapper as Wen suggested, then it's not clear that it has huge value. However, I do think that this kind of library function is useful. We should discuss if there is a generic handler that does more than just what Wen suggested that would be required -- as Suresh says, there are cases where one might have a populated tree (GoStruct) but care about specifically how the request is going to be populated. 

My proposal that I made in a different discussion was whether to make the API look more like:

```golang
type RequestMap map[*gnmipb.Path]gnmipb.UpdateResult.Operation // map between a path and the method that should be used in the SetRequest

// TogNMISetRequest generates a gNMI SetRequest from the contents of the input GoStruct.
func TogNMISetRequest(input GoStruct, m RequestMap, opt ...SetRequestOpt) (*gnmipb.SetRequest, error)
```

This would mean that we could do something like:

```golang
ygot.TogNMISetRequest(device, ygot.RequestMap{
 ygot.StringToStructuredPath("/system/ntp"):   gnmipb.UpdateResult_REPLACE,
 ygot.StringToStructuredPath("/network-instances"): gnmipb.UpdateResult_UPDATE,
})
```

to marshal a specific subtree into a particular field of the `SetRequest`. The value proposition of this as a library function would be that we'd be able to hide the complexity of walking the tree to get specific contents out of the fakeroot that was provided from the calling code.

This would also allow us to handle cases where there was a need to exclude a specific subtree from the contents by having options like:

```golang
ygot.TogNMISetRequest(device, ygot.RequestMap{
  ygot.StringToStructuredPath("/"): gnmipb.UpdateResult_REPLACE,
}, ygot.ExcludePath("/system/aaa"))
```

I think we agreed offline to abandon this PR in favour of (2) being done in the calling libraries until we understand what the general solution might look like.

sthesayi:
Per the PR discussion thread, closing this PR until a time when we have a clear need for this library and then implement it to satisfy all use cases with thorough testing.

PR #529, opened 2021-05-12
Add new path generation flag `-simplify_wildcard_paths`.

wenovus:
This flag omits the keys from the generated paths if all keys are wildcarded for a list node. This is an equivalent representation for the path.

e.g.
`map[string]interface{}{"key1": "*", "key2": "*", "union-key": "*"},` -> `map[string]interface{}{},`
`map[string]interface{}{"key1": "foo", "key2": "*", "union-key": "*"},` -> no change.

The use case for this would be to help a user where a vendor does not yet support the `key=*` format.


generator/generator.go
```diff
@@ -96,6 +96,7 @@ var (
 	// Flags used for PathStruct generation only.
 	schemaStructPath        = flag.String("schema_struct_path", "", "The Go import path for the schema structs package. This should be specified if and only if schema structs are not being generated at the same time as path structs.")
 	generateWildcardPaths   = flag.Bool("generate_wildcard_paths", true, "Whether to generate methods for constructing wildcard paths.")
+	simplifyWildcardPaths   = flag.Bool("simplify_wildcard_paths", false, "Whether to omit the keys in the generated paths if all keys are wildcarded for a list node.")
```
greg-dennis, line 99: maybe "if all keys for a list node are wildcards"
avoids using "wildcard" as a verb

greg-dennis: If you change here, also change similarly below.

greg-dennis (approved):

ypathgen/pathgen.go
```diff
@@ -956,7 +960,13 @@ func generateChildConstructorsForList(methodBuf *strings.Builder, listAttr *ygen
 		// Create the string for the method parameter list, docstrings, and ygot.NodePath's key list.
 		fieldData.KeyParamListStr = strings.Join(paramListStrs, ", ")
 		fieldData.KeyParamDocStrs = paramDocStrs
-		fieldData.KeyEntriesStr = strings.Join(keyEntryStrs, ", ")
+		if simplifyWildcardPaths && comboIndex == 0 {
```
robshakir, line 963: Suggest:

fieldData.KeyEntriesStr = strings.Join(keyEntryStrs, ", ")
if simplifyWildcardPaths && comboIndex == 0 {
   fieldData.KeyEntriesStr = "" // explanation...
}
```

File: ypathgen/pathgen.go [hunk 3]
```diff
@@ -956,7 +960,13 @@ func generateChildConstructorsForList(methodBuf *strings.Builder, listAttr *ygen
...
+			// The zeroth index has everything wildcarded, so it doesn't need to specify any key value.
+			// This is a tweak that helps with vendor compatibility.
```
robshakir, line 965: Which vendor compatibility? It's also not just for compatibility but also according to the [path spec](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md).

wenovus: apparently broadcom, I don't know the details. Technically both are equivalent right? But it looks like the key=* syntax isn't yet supported.

wenovus: I'll make it clear in the comments that it's part of the spec.

generator/generator.go
wenovus: Done.

ypathgen/pathgen.go
```diff
@@ -156,6 +156,10 @@ type GenConfig struct {
 	ListBuilderKeyThreshold uint
 	// GenerateWildcardPaths means to generate wildcard nodes and paths.
 	GenerateWildcardPaths bool
+	// SimplifyWildcardPaths causes non-builder-generated wildcard nodes
+	// where all keys are wildcarded to omit the [key="*"] in the generated
+	// path.
```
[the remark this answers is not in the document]
wenovus: Added this example, thanks!

ypathgen/pathgen.go
wenovus: Done.

ypathgen/pathgen.go
wenovus: Improved comment.

PR #530, opened 2021-05-13
Add a new JSON marshal option that prefers the shadow path name.

wenovus:
This allows a user who is looking to multiplex config and telemetry operations on the same set of GoStructs to treat the GoStructs as either a config struct or a telemetry struct by simplying tweaking this marshal flag.

### DIFF
--- a/ygot/struct_validation_map.go
+++ b/ygot/struct_validation_map.go
@@ -39,17 +39,26 @@ const (
 	indentString string = "   "
 )
 
-// structTagsToLibPaths takes an input struct field as a reflect.Type, and determines
+// structTagToLibPaths takes an input struct field as a reflect.Type, and determines
 // the set of validation library paths that it maps to. Returns the paths as a slice of
 // empty interface slices, or an error.
-func structTagToLibPaths(f reflect.StructField, parentPath *gnmiPath) ([]*gnmiPath, error) {
+func structTagToLibPaths(f reflect.StructField, parentPath *gnmiPath, preferShadowPath bool) ([]*gnmiPath, error) {
 	if !parentPath.isValid() {
 		return nil, fmt.Errorf("invalid path format in parentPath (%v, %v)", parentPath.stringSlicePath == nil, parentPath.pathElemPath == nil)
 	}
 
-	pathAnnotation, ok := f.Tag.Lookup("path")
-	if !ok {
-		return nil, fmt.Errorf("field did not specify a path")
+	var pathAnnotation string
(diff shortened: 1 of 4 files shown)


ygot/struct_validation_map_test.go
```diff
@@ -172,10 +260,37 @@ func TestStructTagToLibPaths(t *testing.T) {
...
+		want: []*gnmiPath{{
+			pathElemPath: []*gnmipb.PathElem{{Name: "existing"}, {Name: "far"}, {Name: "boo"}},
+		}},
 	}}
 
 	for _, tt := range tests {
-		got, err := structTagToLibPaths(tt.inField, tt.inParent)
+		got, err := structTagToLibPaths(tt.inField, tt.inParent, tt.inPreferShadowPath)
 		if (err != nil) != tt.wantErr {
 			t.Errorf("%s: structTagToLibPaths(%v, %v): did not get expected error status, got: %v, want err: %v", tt.name, tt.inField, tt.inParent, err, tt.wantErr)
```
greg-dennis, line 295: add the additional argument in the error message

ygot/struct_validation_map.go
```diff
@@ -39,17 +39,26 @@ const (
...
 	if !parentPath.isValid() {
 		return nil, fmt.Errorf("invalid path format in parentPath (%v, %v)", parentPath.stringSlicePath == nil, parentPath.pathElemPath == nil)
 	}
 
-	pathAnnotation, ok := f.Tag.Lookup("path")
-	if !ok {
-		return nil, fmt.Errorf("field did not specify a path")
+	var pathAnnotation string
+	var ok bool
+	switch preferShadowPath {
```
greg-dennis, line 52: switch statement on a boolean seems non-standard to me. I'd probably do:

var pathAnnotation string
var ok bool
if preferShadowPath {
  pathAnnotation, ok = f.Tag.Lookup("shadow-path")
}
if !ok {
   if pathAnnotation, ok = f.Tag.Lookup("path"); !ok {
      return nil, fmt.Errorf("field did not specify a path")
    }
}

greg-dennis (changes requested):

ygot/struct_validation_map_test.go
wenovus: Done.

ygot/struct_validation_map.go
wenovus: Done.

greg-dennis (approved):

PR #531, opened 2021-05-14
Add option for [Set/Get/GetCreate/Delete]Node for reversing "shadow-path" and "path" behaviour.

wenovus:
This allows a user who is multiplexing a compressed `GoStruct` to choose which of either config or state data to unmarshal, and which one to silently ignore.

Note that if only "path" is present, the behaviour is unaffected.

Due to the nature of how options work for these functions I had to create an option for all of these functions. Also, it seems to me that since these functions form a toolset, that it is prudent to add this option for all of them (e.g. in one use case all but GetNode would be using this option).


ytypes/node.go
```diff
@@ -195,18 +194,41 @@ func retrieveNodeContainer(schema *yang.Entry, root interface{}, path *gpb.Path,
...
 
-		for _, p := range schPaths {
-			if !util.PathMatchesPrefix(path, p) {
-				continue
+		args := args
+		switch args.reverseShadowPath {
+		case true:
+			schPaths := util.ShadowSchemaPaths(ft)
+			for _, p := range schPaths {
+				if !util.PathMatchesPrefix(path, p) {
```
greg-dennis, line 202: invert the polarity:
  if util.PathMatchesPrefix(path, p) {
    return checkPath(p, args)
  }

ytypes/node.go
```diff
@@ -195,18 +194,41 @@ func retrieveNodeContainer(schema *yang.Entry, root interface{}, path *gpb.Path,
 			return retrieveNode(cschema, fv.Interface(), util.TrimGNMIPathPrefix(path, p[0:to]), np, args)
 		}
 
-		for _, p := range schPaths {
-			if !util.PathMatchesPrefix(path, p) {
-				continue
+		args := args
+		switch args.reverseShadowPath {
```
greg-dennis, line 198: I would use an if-statement rather than a switch on a boolean

ytypes/node.go
```diff
@@ -195,18 +194,41 @@ func retrieveNodeContainer(schema *yang.Entry, root interface{}, path *gpb.Path,
...
+			if len(schPaths) != 0 {
+				args.shadowPath = true
+			}
+			fallthrough
+		case false:
+			schPaths, err := util.SchemaPaths(ft)
+			if err != nil {
+				return nil, status.Errorf(codes.Unknown, "failed to get schema paths for %T, field %s: %s", root, ft.Name, err)
+			}
+			for _, p := range schPaths {
```
greg-dennis, line 217: could potentially factor out the loop into a helper function assigned immediately above the switch:

findPrefix := func (prefixes [][]string) []string {
  for _, p := range prefixes {
    if util.PathMatchesPrefix(path, p) {
      return p
    }
  }
  r

greg-dennis (changes requested):

ytypes/node.go
wenovus: Done for the 3 instances.

ytypes/node.go
wenovus: Done.

ytypes/node.go
wenovus: I think this is short enough (and there are just two of them) that I didn't decide to do it. I also didn't want to reason about what happens to `args`, since it might be modified just prior to the call.

greg-dennis (approved):

ytypes/node.go
```diff
@@ -59,6 +59,10 @@ type retrieveNodeArgs struct {
 	// values are skipped (non-existing GoStructs are still initialized),
 	// and any value retrieved is always nil.
 	shadowPath bool
+	// reverseShadowPath reverses the meaning of the "path" and
+	// "shadow-path" tags when both are present while processing a
+	// GoStruct.
+	reverseShadowPath bool
```
robshakir, line 65: `preferShadowPath` perhaps given that we're not really "reversing" a path (this would say to me that `/a/b/c` becomes `/c/b/a`.

ytypes/node.go
```diff
@@ -195,18 +194,35 @@ func retrieveNodeContainer(schema *yang.Entry, root interface{}, path *gpb.Path,
 			return retrieveNode(cschema, fv.Interface(), util.TrimGNMIPathPrefix(path, p[0:to]), np, args)
 		}
 
-		for _, p := range schPaths {
-			if !util.PathMatchesPrefix(path, p) {
-				continue
+		args := args
+		if args.reverseShadowPath {
+			schPaths := util.ShadowSchemaPaths(ft)
```
robshakir, line 199: I found this logic a bit hard to follow. 

iCan you explain why on line 221 when we have `!args.reverseShadowSchemaPaths` we check the shadow paths?

robshakir (changes requested):

ytypes/node.go
wenovus: The meaning of "reverse" here is to indicate that the path tags of "path" and "shadow-path" are effectively swapped during processing. The reason why I didn't use `preferShadowPath` is that it sounds somewhat unclear to me. What this flag does is that during unmarshalling, "shadow-path" is now unmarshalled, and "path" is silently ignored as a no-op. While `preferShadowPath` is a fine description for this behaviour, I think `reverseShadowPath` is very clear on what the behaviour is.

ytypes/node.go
wenovus: I added some comments, hopefully it clarifies the logic.

ytypes/node.go
robshakir: SGTM, thanks for the clarification here.

robshakir (approved):
Thanks for the changes here, LGTM.

PR #533, opened 2021-05-26
Add PathAndStruct helper for generated non-leaf PathStruct types.

wenovus:
A common use case is where a user would like to construct a path and populate its corresponding `GoStruct` at the same time. This helper obviates the need to retrieve the struct using the `GetOrCreateXXX` functions on top of using the path API.

#### Example
```
p, addr := pathAPI.Interface("Ethernet0").Subinterface(0).Ipv4().Address("10.10.10.10").PathAndStruct()
addr.PrefixLength = ygot.Uint8(24)
```

This is not generated for leaves, since their types can be accessed using `ygot` helper functions such as `ygot.Int32()`.

### DIFF
--- a/ypathgen/pathgen.go
+++ b/ypathgen/pathgen.go
@@ -501,6 +501,15 @@ type {{ .TypeName }}{{ .WildcardSuffix }} struct {
 {{- end }}
 `)
 
+	// goPathAndStructHelperTemplate generates a helper for non-leaves that returns
+	// the PathStruct and an empty corresponding GoStruct.
+	goPathAndStructHelperTemplate = mustTemplate("goPathAndStructHelper", `
+// PathAndStruct returns the path struct and an empty {{ .GoStructTypeName }} for the path "{{ .YANGPath }}".
+func (n *{{ .TypeName }}) PathAndStruct() (*{{ .TypeName }}, *{{ .GoStructTypeName }}) {
+	return n, &{{ .GoStructTypeName }}{}
+}
+`)
+
 	// goPathChildConstructorTemplate generates the child constructor method
 	// for a generated struct by returning an instantiation of the child's
 	// path struct object.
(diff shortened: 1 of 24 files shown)


ypathgen/pathgen.go
```diff
@@ -718,14 +743,19 @@ func generateDirectorySnippet(directory *ygen.Directory, directories map[string]
...
 		if err := goPathFakeRootTemplate.Execute(&structBuf, structData); err != nil {
 			return GoPathStructCodeSnippet{}, util.AppendErr(errs, err)
 		}
-	} else if err := goPathStructTemplate.Execute(&structBuf, structData); err != nil {
-		return GoPathStructCodeSnippet{}, util.AppendErr(errs, err)
+	} else {
+		if err := goPathStructTemplate.Execute(&structBuf, structData); err != nil {
+			return GoPathStructCodeSnippet{}, util.AppendErr(errs, err)
+		}
+		if err := goPathAndStructHelperTemplate.Execute(&structBuf, structData); err != nil {
```
robshakir, line 756: Should generating the path and struct helper be optional? In general we've tried to have these other features that are new types of output flag-controlled in the library - even if we have it as on by default, it'd probably be nice to have something whereby a user can turn this off if they don't want it to save on generated code size.

ypathgen/pathgen_test.go
```diff
@@ -2040,6 +2075,11 @@ type List struct {
 	*ygot.NodePath
 }
 
+// PathAndStruct returns the path struct and an empty oc.List for the path "/root-module/list-container/list".
```
robshakir, line 2078: This is interesting -- for YANG `container` types it seems pretty clear to just return the struct, but in the case I call this with `pathlib.ListContainer().ListAny().PathAndStruct()` should I get back `oc.List{}` or `map[...]*oc.List{}`?

greg-dennis:
One alternative is for Replace on an intermediate node to always construct a new value and accept a lambda that can modify it, e.g. 

```
pathAPI.Interface("Ethernet0").Subinterface(0).Ipv4().Address("10.10.10.10").Replace(t, func(addr A) {
  addr.PrefixLength = ygot.Uint8(24)
})
```

Not advocating for it, just thinking through the options.

robshakir:
Wen, is this change defunct at this point -- or should we try and merge it?

greg-dennis (commented):
I think maybe we should include this in the space of changes that will be considered for a merged/revised gNMI API in H1. At this point, this feels a little too much of a piecemeal contribution to me, and we might benefit from incorporating it into the more holistic look that is planned.

wenovus:
When generics is ready a better design for this might be just to provide the function
```go
func ygot.GoStruct[T GoStruct](PathStruct[T]) T {}
```

PR #534, opened 2021-05-28
Add Go 1.16 to CI testing.

robshakir:
Increase coverage to include go1.16.

Going forward, it'd be great to figure out whether there's a go-latest or similar
such that we can include that. Although we say that we support $latest-3 versions,
I think it'd be nice to just keep testing until such time as we find a reason to
need to deprecate the oldest version.

Thoughts?﻿

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: ['1.13', '1.14', '1.15']
+        go: ['1.13', '1.14', '1.15', '1.16']
 
     steps:
       - name: Install protobuf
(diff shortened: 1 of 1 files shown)


wenovus (changes requested):
Given that the GitHub Actions team themselves don't have such a $latest pattern, I doubt there is official support: https://github.com/actions/setup-go/blob/main/.github/workflows/versions.yml#L22

However, I see ['1.x'](https://github.com/github/certstore/blob/main/.github/workflows/test-windows-recent-go-versions.yml#L7) which might help with us forgetting about the newest release.

I'm fine with keeping older versions supported, since we explicitly say the versions we support: https://github.com/openconfig/ygot#getting-started-with-ygot

robshakir:
Thanks! I updated to add `1.x` too. :-)

wenovus (approved):

PR #536, opened 2021-06-08
Add initial implementation of unmarshalling gNMI Notifications->pb.

robshakir:
``` 
 * (M) proto/proto.go
 * (M) proto/proto_test.go
  - Add support for unmarshalling a map of gNMI.Path->value into
    a ygot-generated protobuf. This is done by ranging over the
    protobuf message and using the annotated paths.
  - Initial implementation just supporting string type.
```

This implementation just handles a simple string type, but TypedValue is to be added later such that we can do this directly from gNMI Notifications.

This change starts to implement the 2nd half of the pipeline of:

(new, this PR) `GoStruct->gNMI Notifications->Protobuf`
 - GoStruct -> gNMI Notifications is `ygot.TogNMINotifications`
 - gNMI Notifications -> Protobuf is `protomap.PathsFromProto`

GoStruct<-gNMI Notifications<-Protobuf
 - Protobuf <- gNMI Notifications is this code.
 - GoStruct <- gNMI Notifications is `ytypes.SetNode`

### DIFF
--- a/protomap/proto.go
+++ b/protomap/proto.go
@@ -429,6 +429,7 @@ func makeWrapper(msg protoreflect.Message, fd protoreflect.FieldDescriptor, val 
 			return nil, false, fmt.Errorf("got non-string value for string field, field: %s, value: %v", fd.FullName(), val)
 		}
 		return (&wpb.StringValue{Value: nsv}).ProtoReflect(), true, nil
+	default:
+		return nil, false, nil
 	}
-	return nil, false, nil
 }
--- a/protomap/proto.go
+++ b/protomap/proto.go
@@ -359,3 +359,77 @@ func resolvedPath(basePath, annotatedPath *gpb.Path) *gpb.Path {
 	np.Elem = append(np.Elem, annotatedPath.Elem[len(basePath.Elem):]...)
 	return np
 }
+
+// ProtoFromPaths takes an input proto.Message and adds the values that are specified in the map vals to
+// it, using basePath as the prefix to any paths within the vals map. The message, p, is modified in place.
+// The map, vals, is keyed by the gNMI path to the field which is annotated in the ygot generated protobuf,
+// the complete path is taken to be basePath + the key found in the map.
+func ProtoFromPaths(p proto.Message, vals map[*gpb.Path]interface{}, basePath *gpb.Path) error {
+	m := p.ProtoReflect()
+
+	directCh := map[*gpb.Path]interface{}{}
+	for p, v := range vals {
+		// TODO(robjs): needs fixing for compressed schemas.
+		l := 0
+		if basePath != nil {
+			l = len(basePath.Elem)
+		}
+		if len(p.Elem) == l+1 {
+			directCh[p] = v
+		}
+	}
+
+	var rangeErr error
+	unpopRange{m}.Range(func(fd protoreflect.FieldDescriptor, v protoreflect.Value) bool {
+		annotatedPath, err := annotatedSchemaPath(fd)
+		if err != nil {
+			rangeErr = err
+			return false
+		}
+
+		for _, ap := range annotatedPath {
+			for chp, chv := range directCh {
+				if proto.Equal(ap, chp) {
+					if fd.Kind() == protoreflect.MessageKind {
+						v, isWrap, err := makeWrapper(m, fd, chv)
+						if err != nil {
+							rangeErr = err
+							return false
+						}
+						if !isWrap {
+							// TODO(robjs): recurse into the message if it wasn't a wrapper
+							// type.
+							rangeErr = errors.New("unimplemented: child messages")
+							return false
+						}
+						m.Set(fd, protoreflect.ValueOfMessage(v))
+					}
+				}
+			}
+		}
+		return true
+	})
+
+	if rangeErr != nil {
+		return rangeErr
+	}
+
+	return nil
+}
+
+// makeWrapper generates a new message for field fd of the proto message m with the value set to val.
+// The field fd must describe a field that has a message type. An error is returned if the wrong
+// type of payload is provided for the message. The second, boolean, return argument specifies whether
+// the message provided was a known wrapper type.
+func makeWrapper(msg protoreflect.Message, fd protoreflect.FieldDescriptor, val interface{}) (protoreflect.Message, bool, error) {
+	newV := msg.NewField(fd)
+	switch newV.Message().Interface().(type) {
+	case *wpb.StringValue:
+		nsv, ok := val.(string)
+		if !ok {
+			return nil, false, fmt.Errorf("got non-string value for string field, field: %s, value: %v", fd.FullName(), val)
+		}
+		return (&wpb.StringValue{Value: nsv}).ProtoReflect(), true, nil
+	}
+	return nil, false, nil
+}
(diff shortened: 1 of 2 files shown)


protomap/proto.go
```diff
@@ -359,3 +359,77 @@ func resolvedPath(basePath, annotatedPath *gpb.Path) *gpb.Path {
...
+	})
+
+	if rangeErr != nil {
+		return rangeErr
+	}
+
+	return nil
+}
+
+// makeWrapper generates a new message for field fd of the proto message m with the value set to val.
```
wenovus, line 420: nit: of the proto message **msg**

protomap/proto.go
```diff
@@ -359,3 +359,77 @@ func resolvedPath(basePath, annotatedPath *gpb.Path) *gpb.Path {
...
+	newV := msg.NewField(fd)
+	switch newV.Message().Interface().(type) {
+	case *wpb.StringValue:
+		nsv, ok := val.(string)
+		if !ok {
+			return nil, false, fmt.Errorf("got non-string value for string field, field: %s, value: %v", fd.FullName(), val)
+		}
+		return (&wpb.StringValue{Value: nsv}).ProtoReflect(), true, nil
+	}
+	return nil, false, nil
```
wenovus, line 434: Since all cases in the switch end in a return, suggest to put the last return in a default statement. This is one condition where a missing return won't be flagged by the compiler.

protomap/proto.go
```diff
@@ -359,3 +359,77 @@ func resolvedPath(basePath, annotatedPath *gpb.Path) *gpb.Path {
...
+					if fd.Kind() == protoreflect.MessageKind {
+						v, isWrap, err := makeWrapper(m, fd, chv)
+						if err != nil {
+							rangeErr = err
+							return false
+						}
+						if !isWrap {
+							// TODO(robjs): recurse into the message if it wasn't a wrapper
+							// type.
+							rangeErr = errors.New("unimplemented: child messages")
```
wenovus, line 402: Maybe say which type the unsupported field is in the error message?

protomap/proto_test.go
```diff
@@ -289,3 +289,53 @@ func TestPathsFromProtoInternal(t *testing.T) {
 		})
 	}
 }
+
+func TestProtoFromPathsInternal(t *testing.T) {
```
wenovus, line 293: Why is this called Internal?

protomap/proto_test.go
```diff
@@ -289,3 +289,53 @@ func TestPathsFromProtoInternal(t *testing.T) {
...
+		inVals           map[*gpb.Path]interface{}
+		inBasePath       *gpb.Path
+		wantProto        proto.Message
+		wantErrSubstring string
+	}{{
+		desc:    "string field",
+		inProto: &epb.ExampleMessage{},
+		inVals: map[*gpb.Path]interface{}{
+			mustPath("/string"): "hello",
+		},
```
wenovus, line 306: Add other longer/shorter paths to test the filtering?

protomap/proto.go
```diff
@@ -359,3 +359,77 @@ func resolvedPath(basePath, annotatedPath *gpb.Path) *gpb.Path {
...
+		annotatedPath, err := annotatedSchemaPath(fd)
+		if err != nil {
+			rangeErr = err
+			return false
+		}
+
+		for _, ap := range annotatedPath {
+			for chp, chv := range directCh {
+				if proto.Equal(ap, chp) {
+					if fd.Kind() == protoreflect.MessageKind {
```
wenovus, line 393: What should happen if there is a child that can't be found in the message? Should we error out?

wenovus: I see in the next PR you address this.

wenovus:
A question: why do we want to support simple types?

protomap/proto.go
```diff
@@ -359,3 +359,77 @@ func resolvedPath(basePath, annotatedPath *gpb.Path) *gpb.Path {
 	np.Elem = append(np.Elem, annotatedPath.Elem[len(basePath.Elem):]...)
 	return np
 }
+
+// ProtoFromPaths takes an input proto.Message and adds the values that are specified in the map vals to
+// it, using basePath as the prefix to any paths within the vals map. The message, p, is modified in place.
+// The map, vals, is keyed by the gNMI path to the field which is annotated in the ygot generated protobuf,
```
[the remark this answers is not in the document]
robshakir: Fixed. Thanks.

protomap/proto.go
```diff
@@ -359,3 +359,77 @@ func resolvedPath(basePath, annotatedPath *gpb.Path) *gpb.Path {
 	np.Elem = append(np.Elem, annotatedPath.Elem[len(basePath.Elem):]...)
 	return np
 }
+
+// ProtoFromPaths takes an input proto.Message and adds the values that are specified in the map vals to
```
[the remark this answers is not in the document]
robshakir: Done, thanks. Also renamed `basePath` -> `prefix` to be more consistent.

protomap/proto.go
```diff
@@ -359,3 +359,77 @@ func resolvedPath(basePath, annotatedPath *gpb.Path) *gpb.Path {
...
+// it, using basePath as the prefix to any paths within the vals map. The message, p, is modified in place.
+// The map, vals, is keyed by the gNMI path to the field which is annotated in the ygot generated protobuf,
+// the complete path is taken to be basePath + the key found in the map.
+func ProtoFromPaths(p proto.Message, vals map[*gpb.Path]interface{}, basePath *gpb.Path) error {
+	m := p.ProtoReflect()
```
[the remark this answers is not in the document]
robshakir: Done, and done.

protomap/proto.go
robshakir: Thanks, fixed!

protomap/proto.go
```diff
@@ -359,3 +359,77 @@ func resolvedPath(basePath, annotatedPath *gpb.Path) *gpb.Path {
...
+	directCh := map[*gpb.Path]interface{}{}
+	for p, v := range vals {
+		// TODO(robjs): needs fixing for compressed schemas.
+		l := 0
+		if basePath != nil {
+			l = len(basePath.Elem)
+		}
+		if len(p.Elem) == l+1 {
+			directCh[p] = v
+		}
```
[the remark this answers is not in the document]
robshakir: Thanks! done :-)

protomap/proto_test.go
robshakir: Fixed -- this was a hangover from a previous design.

protomap/proto.go
robshakir: I'm not sure it matters for the moment, right? Any child message is unsupported. I've added it, but I'm not sure it's that useful :-)

protomap/proto_test.go
robshakir: Done, thanks.

robshakir:
Supporting simple types leaves it open that someone doesn't use `ygot.TogNMINotifications`, and rather pulls things from a JSON file -- if we insist on `TypedValue` then it means that they would need to do other work (`value.FromScalar`) just for us to undo that work.

Thanks for the review, PTAL at the changes.

protomap/proto.go
robshakir: Done - the static analysis/linter does a good job on this - but I think your suggestion makes sense.

wenovus (approved):

PR #541, opened 2021-06-10
Add JoinPaths utility function

greg-dennis:


greg-dennis:
I'll add unit tests and update the PR

util/gnmi.go
```diff
@@ -196,3 +196,27 @@ func FindModelData(mods []*yang.Entry) ([]*gpb.ModelData, error) {
 
 	return modelData, nil
 }
+
+// JoinPaths joins an prefix and suffix paths, returning an error if their
+// target or origin fields don't match.
```
wenovus, line 201: ... returning an error if their
// target or origin fields **are both non-empty and** don't match.

util/gnmi_test.go
```diff
@@ -766,3 +766,61 @@ func TestFindModelData(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			got, err := JoinPaths(tt.prefix, tt.suffix)
+			if diff := errdiff.Substring(err, tt.wantErrSubstring); diff != "" {
+				t.Errorf("JoinPaths(%v, %v) got unexpected error diff: %s", tt.prefix, tt.suffix, diff)
+			}
+			if err != nil {
+				return
+			}
+			if diff := cmp.Diff(tt.want, got, cmp.Comparer(proto.Equal)); diff != "" {
```
wenovus, line 821: optional: Can use `protocmp.Transform()` instead of proto.Equal. It provides a better diff message since it can traverse the proto message.

wenovus (approved):

util/gnmi.go
greg-dennis: Done.

util/gnmi_test.go
greg-dennis: Done. I had copied the style of the other unit tests. Probably worth doing a larger cleanup at some point to move everything to protocmp.Transform().

PR #543, opened 2021-06-11
Don't initialize leaf values when unmarshalling valid shadow paths.

wenovus:
Currently when `SetNode` is called on a path, missing elements are initialized as the function traverses through the `GoStruct`. This happens even for paths leading to a leaf as well as shadow paths, causing the default value to always be populated into these elements.

This change prevents `SetNode` from initializing a leaf value when traversing a shadow path.

`SetNode`'s test is also augmented to test the parent struct's value. Previously it didn't test it, which is why this case was missed.


ytypes/node.go
```diff
@@ -144,7 +144,9 @@ func retrieveNodeContainer(schema *yang.Entry, root interface{}, path *gpb.Path,
 				to--
 			}
 
-			if args.modifyRoot {
+			// If args.modifyRoot is true, then initialize the field before possibly searching further.
+			// However, if the field is a leaf, we don't want to modify it if shadowPath is true.
+			if args.modifyRoot && !(args.shadowPath && cschema != nil && cschema.IsLeaf()) {
```
robshakir, line 149: Do we want to modify it when `shadowPath` is false? It seems like this is an oversight in general, such that we're initialising leaves even though we don't need to?

robshakir: I'd prefer to make the behaviour consistent across shadow and non-shadow paths.

ytypes/node.go
```diff
@@ -399,6 +401,9 @@ type GetOrCreateNodeOpt interface {
 // along the path if they are nil.
 // Function returns the value and schema of the node as well as error.
 // Note that this function may modify the supplied root even if the function fails.
+// Note that this function may create containers or list entries even if the input path is a shadow path.
+// TODO(wenbli): a traversal should remember what containers or list entries were created so that a shadow path
+//               can later undo this. This applies to SetNode as well.
```
robshakir, line 406: If a container == nil, or a map has no members, is there a reason to delete it? I presume that any API (`GetNode`, `Diff`, `TogNMINotifications`, `Marshal7951`) all have to handle this case anyway?

It's the same as doing `&oc.Device{}.GetOrCreateSystem().GetOrCreateNtp()` -- in this scenario even though there was nothing to be marshalled, we set the containers to non-nil.

robshakir (changes requested):

ytypes/node.go
wenovus: `modifyRoot` is set by `GetOrCreateNode` and `SetNode`. For `GetOrCreateNode` on a leaf path, do we want to create a default value?

ytypes/node.go
wenovus: I thought about whether this matters and my concern is that if you do `&oc.Device{}.GetOrCreateInterface("eth0")`, the key leaf gets populated. Otherwise I agree there is no reason to delete it.

wenovus: I meant if you pass in the path `/interfaces/interface[name="eth0"]` to `GetOrCreateNode`.

greg-dennis (commented):
I'll leave the review to Rob.

ytypes/node.go
robshakir: The key leaf gets created, and its value set though right, so at this point the `map` entry is not actually empty. 

 I think the context that I was missing here is why we need this for shadow paths but not for the normal paths?

ytypes/node.go
robshakir: I guess I'm fine with creating the default value -- what I'm less certain of here is why there is a different for shadow path vs. regular path. If we were to call `GetOrCreateNode` with `shadowPath` = `true` should there be a difference to if we had `shadowPath=false`?

wenovus: If `-ignore_shadow_schema_paths` is not specified during generation, then the shadow paths aren't in the annotation, which means such a call errors out. The original intent of adding this option was to gracefully ignore these paths instead of erroring out. A use case for this option is when we receive both config and state values for telemetry, and need to unmarshal them into compressed structs where we care about only the config or state values, but don't want to see a bunch of error messages because the device streamed both kinds of values.

If we always created the value here, it would deviate from the original intent of this option, and isn't in line with the application aforementioned. The intention with the difference in behaviour is to continue with this original intent, and support this use case correctly.

ytypes/node.go
wenovus: There is a short explanation of the purpose of shadow paths where the flag is defined: https://github.com/openconfig/ygot/blob/master/generator/generator.go#L72

robshakir: Sorry -- I understand what a shadow path is. Why do we delete empty entries for shadow paths and not for normal paths though? i.e., what is the problem that you're trying to solve here, and why is it only a problem for shadow paths?

ytypes/node.go
robshakir: In what scenario does it error out - if I were to make a `GetOrCreateNode`/`SetNode` call that is for a shadow path that doesn't exist? (It's not quite clear to me what this means because `-ignore_shadow_schema_paths` presumably means that (by default) we do add these annotations?)

So is the case that we're discussing here the one whereby there are no annotations for shadow paths, and we still call `SetNode` for that case?

I'm not that comfortable with the idea that we make a change where shadow paths is true and then don't make it for the other case -- I'm still not clear what the "application aforementioned" is in this context. Are you saying that in some case where you get `/interfaces/interface[name=eth0]/config/name` then you don't want `/interfaces/interface[name=eth0]/state/name` to be set in this case?

wenovus: First of all, just making sure, `args.shadowPath` here is *NOT* a flag from the user. This is an internal tracking variable that tracks whether the current path is a shadow path or not. As the code recurses and descends into the input `GoStruct`, the annotations will reveal whether the input path is a non-shadow path or a shadow-path. This variable only helps the code decide on the right thing to do given the path type.

With that aside, the example you have is right: the reason I implemented `-ignore_shadow_schema_paths` is to silence errors that were coming from sibling config/state paths when using a compressed schema that prioritized either state/config leaves. If the goal is to only silence the error when given a shadow path, it would be wrong to also initialize the leaf. After all, prior to this flag, you don't get either of these, but an error instead that says the shadow path doesn't exist -- because there was no notion of a shadow path.

More history on what led to this issue: Originally, shadow paths were treated just like any bogus path, since compressed `GoStruct`s has only room for one of the siblings. However, the errors became distracting. We could solve this by simply having the application filter out paths that end with either 'state/xxx' or 'config/xxx'. Instead, I chose to have ygot annotate this as a shadow path. This means that the code goes through the trouble of trying to match the path to each element of the `GoStruct` as it descends through it, while being careful to not touch anything. This is the source of this problem.

ytypes/node.go
wenovus: The problem is that even if the function call didn't fail for a shadow path, it shouldn't have created a list entry since it made it not empty. Unmarshalling a shadow path should not modify the struct. See above for an explanation, happy to discuss this offline if it's still unclear.

ytypes/node.go
robshakir: Thanks for the explanation, but unfortunately I'm not sure this is getting much clearer for me. Let me try and break this down:

1. You want to be able to ignore the error in ygot when someone calls `Set` or `Get` to a path, because they're using a compressed schema and this schema has removed that field.
2. In this case, someone sends you a path to a leaf and today, you find the schema for it from `util.ChildSchema` from the input path - and it gets initialised to a non-nil value because it was found even though it was only found through the 'shadow path`.

This has brought a lot of complexity into the code, IMHO:

 * In https://github.com/openconfig/ygot/blob/dcfe4c7ca781fe9ad400cd28c4ed4b5e3e753239/ytypes/node.go#L152 -- all the branches from here down seem to do nothing if we have `args.shadowPath` set to true.
 * Shadow paths (compression) cannot exist for anything other than a leaf in `config` or `state` AFAICT, so it's not clear to me that we can ever have a container that has a non-nil shadow path.
     * So in `retrieveNodeContainer` (called for both list entries, and containers) I would recommend added something like:

```
if args.shadowPath {
   if !cschema.IsLeaf() { 
      return nil, fmt.Errorf("cannot have a non-leaf schema that has a shadow path")
   }
   return nil, nil
}
```
   *  This will at least encapsulate where `shadowPath` is referred to.
   *  It also means that this doesn't need to be something that we use when we traverse further down the tree.

Can you explain to me why we need `reverseShadowPath`? This seems like the use case would be that we're doing a `SetNode` or `GetNode` and somehow need to prefer the one that was less preferred -- but it's not clear to me how we ever get into this case given that it must be for a leaf -- we'll fail to find `config/foo` and rather find `state/foo` and then write it.

I'm happy to set something up for us to talk through this -- `retrieveNode` and friends are already complex, and I'm not convinced that what we've taken on here to fix this case is actually that maintainable.

ytypes/node.go
robshakir: Per my other response, I'm not sure that we're aligned on when these shadow paths can occur, and what needs to be handled -- let's get to the bottom of that and then come back to this thread.

robshakir (changes requested):

ytypes/node.go
robshakir: So, as we discussed -- we agreed to:

1. we'll just check whether the `cschema` is a leaf. If it is, and the path is a `shadowPath` - this is a NOOP (i.e., don't create it). If it is not a leaf, return an error. We'll remove `shadowPath` from `checkpath`.
2. going forward, we'll keep a note of the first entry that we initialise when doing `SetNode` or `GetOrCreate....`, and if the final call fails or is ignored (because it is a shadowpath), we'll uninitialise the first initialised node to undo our work. This will be for all use cases of this function.

wenovus: I have made changes in this PR to realize 1. Created https://github.com/openconfig/ygot/issues/544 to track 2.

robshakir (approved):
Thank you for all the changes here! This is a lot clearer! :-D

PR #545, opened 2021-06-17
Add support for rewriting the appended module name.

robshakir:
```
 * (M) ygot/render.go
 * (M) ygot/render_test.go
  - This change allows the contents of the `module` tag to be rewritten
    when outputting RFC7951-compatible JSON. Particularly, this is
    useful when a node that was removed from the schema is re-added
    using an augmentation, and the user wants to specify that ygot
    should pretend that the node is still in the original module.
```

### DIFF
--- a/ygot/render_test.go
+++ b/ygot/render_test.go
@@ -1794,14 +1794,15 @@ func (t *unmarshalableJSON) UnmarshalJSON(d []byte) error {
 
 func TestConstructJSON(t *testing.T) {
 	tests := []struct {
-		name         string
-		in           GoStruct
-		inAppendMod  bool
-		wantIETF     map[string]interface{}
-		wantInternal map[string]interface{}
-		wantSame     bool
-		wantErr      bool
-		wantJSONErr  bool
+		name                     string
+		in                       GoStruct
+		inAppendMod              bool
+		inRewriteModuleNameRules map[string]string
+		wantIETF                 map[string]interface{}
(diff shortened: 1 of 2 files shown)


ygot/render_test.go
```diff
@@ -1873,6 +1874,67 @@ func TestConstructJSON(t *testing.T) {
...
+		inAppendMod: true,
+		inRewriteModuleNameRules: map[string]string{
+			// rewrite m1 to m2
+			"m1": "m2",
+		},
+		wantIETF: map[string]interface{}{
+			"m2:foo": map[string]interface{}{
+				"value-one":    "one",
+				"m3:value-two": "two",
+				"value-three":  "three",
```
wenovus, line 1900: Should "value-three" be "m1:value-three"?

robshakir: No - since we rewrote m1->m2 then this becomes `m2:value-three`, but since we're in `m2` then we write this back to `m1`.

wenovus: Ahh I see, thanks for clarifying.

robshakir: No problem! Was there anything else you wanted to clarify here? Thanks for looking!

wenovus (approved):

ygot/render_test.go
wenovus: Nope everything looks good, approved!

robshakir:
Thanks!

PR #546, opened 2021-06-18
Always pair the leafref path with only the referred leaf in the struct tags for compressed schemas.

wenovus:
Currently for compressed schema GoStructs, we always put the key path in the `path` annotation of the struct tags, even if the leafref refers to the node under the config, and preferOperationalState=true, e.g. `path:"index|state/index" shadow-path:"config/index"`. However, if `index` points to `../config/index`, `index` doesn't really belong with `state/index`.

Instead, a more natural annotation would be `path:"state/index" shadow-path:"index|config/index"`.

This PR proposes to pair the leafref path with only the node that it points to.

This change leads to these behavioural changes:
- When unmarshalling datapoints to the annotation above, paths to the config-referring "index" are ignored. This makes sense since this isn't a state data anyways.
- When marshalling datapoints with the above annotation, and with reverseShadowPaths=true, the key gets marshalled, which is correct since it is part of the configuration. Conversely, when marshalling with reverseShadowPaths=false, the key path is omitted.

I updated all three `exampleoc`s, but only `opstateoc`'s generated code was affected by this PR alone.

### DIFF
--- a/ygen/codegen_test.go
+++ b/ygen/codegen_test.go
@@ -1563,16 +1563,6 @@ func TestGetDirectoriesAndLeafTypes(t *testing.T) {
 				},
 				Path: []string{"", "enum-module", "parent", "child"},
 			},
-			"/enum-module/a-lists/a-list": {
-				Name:   "AList",
-				Fields: map[string]*yang.Entry{}, // Key is only part of state and thus is excluded.
-				Path:   []string{"", "enum-module", "a-lists", "a-list"},
-			},
-			"/enum-module/b-lists/b-list": {
-				Name:   "BList",
-				Fields: map[string]*yang.Entry{},
-				Path:   []string{"", "enum-module", "b-lists", "b-list"},
-			},
 		},
 		wantTypeMap: map[string]map[string]*MappedType{
 			"/enum-module/parent": {
@@ -1586,8 +1576,6 @@ func TestGetDirectoriesAndLeafTypes(t *testing.T) {
 				"id2":         {NativeType: "E_EnumTypes_ID", IsEnumeratedValue: true},
 				"inline-enum": {NativeType: "E_Child_InlineEnum", IsEnumeratedValue: true},
 			},
-			"/enum-module/a-lists/a-list": {},
-			"/enum-module/b-lists/b-list": {},
 		},
 	}, {
 		name:           "simple openconfig test with openconfig-simple module excluded",
--- a/ygen/gogen.go
+++ b/ygen/gogen.go
@@ -2116,7 +2116,10 @@ func findMapPaths(parent *Directory, fieldName string, compressPaths, shadowSche
 	}
 	// Only for compressed data schema paths for list fields do we have the
 	// possibility for a direct leafref path as a second path for the field.
-	if !compressPaths || parent.ListAttr == nil || shadowSchemaPaths {
+	// Furthermore, if we don't even have a first path, then we cannot have
+	// a second path either. This can happen for shadow paths, which may
+	// not exist for some leaves.
+	if !compressPaths || parent.ListAttr == nil || childPath == nil {
 		return mapPaths, nil
 	}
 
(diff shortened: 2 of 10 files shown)


ygen/gogen.go
```diff
@@ -2146,6 +2149,15 @@ func findMapPaths(parent *Directory, fieldName string, compressPaths, shadowSche
 		// resolved it the field that the leafref points to. So, we
 		// compare their absolute paths for equality.
 		if k.Parent.Parent.Dir[k.Name].Type.Kind == yang.Yleafref && cmp.Equal(util.SchemaPathNoChoiceCase(k), fieldSlicePath) {
+			if util.IsConfig(k.Parent.Parent.Dir[k.Name]) != (util.StripModulePrefix(childPath[len(childPath)-2]) == "config") {
```
robshakir, line 2152: Make a helper function for "isInConfig" ?

ygen/gogen.go
```diff
@@ -2146,6 +2149,15 @@ func findMapPaths(parent *Directory, fieldName string, compressPaths, shadowSche
...
+				// The key node is a second path to this leaf if both are either config or state.
+				// For example, if the leafref is a config value, then it must refer to the config
+				// version of the child leaf, and vice versa for a state value, in order to pair the key
+				// with the leaf. The reason why leafref's "Path" field cannot be used
```
robshakir, line 2156: What is the transformation that `TransformEntry` does to the `Path` field here that makes it unreliable?

ygen/testdata/structs/openconfig-withlist-opstate.formatted-txt
```diff
@@ -185,8 +185,8 @@ func (t *Model) RenameSingleKey(oldK, newK string) error {
 
 // Model_MultiKey represents the /openconfig-withlist/model/b/multi-key YANG schema element.
 type Model_MultiKey struct {
-	Key1	*uint32	`path:"state/key1|key1" module:"openconfig-withlist"`
-	Key2	*uint64	`path:"state/key2|key2" module:"openconfig-withlist"`
+	Key1	*uint32	`path:"state/key1" module:"openconfig-withlist"`
```
robshakir, line 188: This doesn't look right... At this point you're saying that we don't serialise this value into the direct child which would be invalid for the schema - i.e., for schema

```yang
list foo
   config false;
   key "a";

   leaf a {
       type leafref {
          path "../state/a";
       }
   }
   
   container state {
       leaf a { type string; }
   }
}
```

we need to produce:

```
/foo/a
/foo/state/a
```

In this case, it seems like you removed the ability to produce /foo/a - which is invalid, since now `key "a"` is nil under the list.

ygot/schema_tests/schema_test.go
```diff
@@ -187,9 +187,6 @@ func TestDiff(t *testing.T) {
 		}(),
 		want: &gnmipb.Notification{
 			Update: []*gnmipb.Update{{
-				Path: mustPath("neighbors/neighbor[neighbor-address=192.0.2.1]/neighbor-address"),
```
robshakir, line 190: This isn't right -- I think related to the above problem, this should be produced.

ygot/schema_tests/testdata/notification_union_int64_opstate.txtpb
```diff
@@ -1,23 +1,3 @@
-update: <
```
robshakir, line 1: Same as above.

ytypes/schema_tests/testdata/bgp-example-opstate.json
```diff
@@ -4,7 +4,6 @@
       "afi-safis": {
         "afi-safi": [
           {
-            "afi-safi-name": "openconfig-bgp-types:IPV4_UNICAST",
```
robshakir, line 7: All these are not right - there's definitely a bug here.

robshakir (changes requested):
It seems like there's some misunderstanding around how `config false` lists are handled -- please ping me if you'd like to chat real-time about this.

r.

ygen/codegen_test.go
```diff
@@ -1563,16 +1563,6 @@ func TestGetDirectoriesAndLeafTypes(t *testing.T) {
 				},
 				Path: []string{"", "enum-module", "parent", "child"},
 			},
-			"/enum-module/a-lists/a-list": {
```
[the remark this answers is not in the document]
wenovus: This is because I added `config false;` to testdata/modules/enum-module.yang in order to make its leafref make sense, since it's referring to a state value. A config leafref can't refer to a state node.

Once this is added, this list is now excluded since it's state-only.

ygen/gogen.go
wenovus: `TransformEntry` mutates the `Path` field by changing from `config` to `state` in its path string. This happens if `PreferOperationalState=true` (see https://github.com/openconfig/ygot/pull/424). It seemed that in one or more places, the code depended on `Path` actually referring to the "prioritized" sibling, so we can't use this field to determine to which one the leafref points.

ygen/testdata/structs/openconfig-withlist-opstate.formatted-txt
wenovus: Yes this change does remove `/foo/a` if the leafref points to `"../config/a"`, but I believe in your example it should keep it since it points to the node under state.

enum-module.yang has this snippet:
```
  container a-lists {
    list a-list {
      config false;
      key "value";

      leaf value {
        type leafref {
          path "../state/value";
        }
      }

      container state {
        config false;
        uses common-value;
      }
    }
  }
```

enum-module.formatted-txt:
```
// AList represents the /enum-module/a-lists/a-list YANG schema element.
type AList struct {
    Value   AList_Value_Union   `path:"state/value|value" module:"enum-module"`
}
```

wenovus:
So another way to do this would be to keep the current `PreferOperationalState=true, ShadowPaths=false` output the same, and only remove the key name from the `path:` annotation in the struct tags when the leafref points to the shadowed sibling node when `ShadowPaths=true`.

This way if the user doesn't want shadow paths, at least a key will be there.

wenovus:
As discussed, will abandon this solution in favour of just duplicating the key into both `path:` and `shadow-path:` annotations.

wenovus:
Superseded by #547

PR #547, opened 2021-06-18
Add key to shadow-path annotation.

wenovus:
Currently for compressed schema GoStructs, we only put the key path in the `path` annotation of the struct tags, e.g. `path:"index|state/index" shadow-path:"config/index"`. However, when a compressed GoStruct is to be used either as a config object or a state object, we need to be able to marshal the key path as well, e.g. in order to create a valid JSON.

This PR proposes to make the path tags `path:"index|state/index" shadow-path:"index|config/index"`.

This change only affects behaviour when `shadow-path` is ever used for marshalling or unmarshalling. e.g. when `PreferShadowPath` is turned on when marshalling JSON, or when `ReverseShadowPath` is turned on for Set/GetNode.

I updated all three `exampleoc`s, but only `opstateoc`'s generated code was affected by this PR alone.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/google/go-cmp v0.5.5
 	github.com/kylelemons/godebug v1.1.0
 	github.com/openconfig/gnmi v0.0.0-20200508230933-d19cebf5e7be
-	github.com/openconfig/goyang v0.2.2
+	github.com/openconfig/goyang v0.2.5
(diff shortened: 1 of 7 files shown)


wenovus:
@robshakir Can we rename `ReverseShadowPath` to PreferShadowPath in a follow-up PR to keep things consistent? I think the original reason why I named it "reverse" no longer applies after this change.

robshakir (approved):
Thanks for the change -- this looks much more parsable and maintainable.

robshakir:
I'm absolutely OK with changing the option name.

PR #549, opened 2021-06-21
Support explicitly setting proto field tags.

wenovus:
Credits to @jasonewang for posting the first implementation at #339. I've modified this a bit to read the extension values from the new openconfig-codegen-extensions.yang model that's now in openconfig/public, and also to clarify the corner cases that were discussed in the PR.

Spec:
https://github.com/openconfig/ygot/blob/master/docs/yang-to-protobuf-transformations-spec.md#field-numbering

Corner cases not mentioned in the spec:
1. multiple `field-number` extensions.
   - Error out.
1. multiple `field-number-offset` extensions.
   - Add them up.
1. `field-number` or `field-number-offset` extensions on non-leaves.
   - Silently ignore.
1. `field-number-offset` on non-uses statements, including leafs.
   - Allow.
1. Conflicts.
   - Allow. The proto compilation will of course fail. Automated checking can be added here if we'd prefer.

I made some simple choices for the above items. Erroring out for 3 and 4 would likely require changing `goyang` to specifically support checking these cases, which I'm not sure is worth it here.


wenovus:
@jasonewang would you be able to consent to this PR being submitted based on your prior commits? (please see googlebot's post)

wenovus:
@googlebot I consent

wenovus:
@jasonewang Do you think you can consent to your commits being contributed to this project? If so could you consent like I did above? Thanks a lot.

wenovus:
Ok, I figured out you can push to the fork repository. So I just did that instead.

PR #551, opened 2021-06-29
[Go]: Resolve conflict when a multi-keyed list contains a container named "key"

wenovus:
Do this by keeping the current naming, but change to "%s_%s_key" instead of "%s_%s_Key" as the back up name if a conflict like this occurs. This should always work since child non-leaves are named using camel cases, so I'm making it error out if the backup is already taken.

Tested:
- Added a new integration test for this case.


ygen/gogen.go
```diff
@@ -2008,6 +2008,13 @@ func yangListFieldToGoType(listField *yang.Entry, listFieldName string, parent *
 		// generatedGoMultiKeyListStruct struct, which is then expanded by a template to the struct
 		// definition.
 		listKeyStructName = fmt.Sprintf("%s_%s_Key", parent.Name, listFieldName)
+		if gogen.definedGlobals[listKeyStructName] {
+			listKeyStructName = fmt.Sprintf("%s_%s_key", parent.Name, listFieldName)
```
robshakir, line 2012: I think this is non-obvious to the user, the name difference `Foo_Key` vs `Foo_key` in a long name seems insufficient, as well as not auto-complete friendly.

I would suggest that maybe use something like `YANGListKey` at this stage -- we tried to do the nice thing of having it just be `Key`, but given that the schema wouldn't let us do that then going with a longer-but-more-explicit name might be best?

robshakir (commented):
The change looks good -- just one discussion for us to have about how to name things in this case.

ygen/gogen.go
wenovus: Makes sense, the user might be confused what the difference between "_key" and "_Key" are, but a more explicit name makes that clear.

robshakir: Great, thanks -- can we update [the docs](https://github.com/openconfig/ygot/blob/master/docs/design.md#yang-lists) to specify that this is the case, just so there's some reference for users?

wenovus: Done.

robshakir (approved):
Thanks for the changes!

PR #552, opened 2021-07-26
Add DiffOption to only consider original paths.

dan-lepage:
Useful when you just want to check that block X is a subset of block Y.

### DIFF
--- a/ygot/diff.go
+++ b/ygot/diff.go
@@ -353,6 +353,25 @@ type DiffOpt interface {
 	IsDiffOpt()
 }
 
+// IgnoreAdditions is a DiffOpt that indicates newly-added fields should be
+// ignored. The returned Notification will only contain the updates and
+// deletions from original to modified.
+type IgnoreAdditions struct{}
+
+func (*IgnoreAdditions) IsDiffOpt() {}
+
+// hasIgnoreAdditions returns the first IgnoreAdditions from an opts slice, or
+// nil if there isn't one.
+func hasIgnoreAdditions(opts []DiffOpt) *IgnoreAdditions {
+	for _, o := range opts {
+		switch v := o.(type) {
+		case *IgnoreAdditions:
+			return v
+		}
+	}
+	return nil
+}
+
 // DiffPathOpt is a DiffOpt that allows control of the path behaviour of the
 // Diff function.
 type DiffPathOpt struct {
@@ -432,7 +451,9 @@ func Diff(original, modified GoStruct, opts ...DiffOpt) (*gnmipb.Notification, e
 			n.Delete = append(n.Delete, origPath.gNMIPaths...)
 		}
 	}
-
+	if hasIgnoreAdditions(opts) != nil {
+		return n, nil
+	}
 	// Check that all paths that are in the modified struct have been examined, if
 	// not they are updates.
 	for modPath, modVal := range modLeaves {
--- a/ygot/diff_test.go
+++ b/ygot/diff_test.go
@@ -839,6 +839,34 @@ func TestDiff(t *testing.T) {
 			}},
 		},
 	}, {
+		desc: "one path each modified, deleted, and added with IgnoreNewPaths set",
+		inOrig: &renderExample{
+			IntVal:   Int32(5),
+			FloatVal: Float32(1.5),
+			Int64Val: Int64(100),
+		},
+		inMod: &renderExample{
+			IntVal:   Int32(10),
+			Str:      String("cabernet-sauvignon"),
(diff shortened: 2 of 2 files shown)


ygot/diff.go
```diff
@@ -365,6 +365,9 @@ type DiffPathOpt struct {
 	// generated structs, which can result in duplication of list key leaves in
 	// the diff output.
 	MapToSinglePath bool
+	// OriginalPathsOnly specifies that only paths in the original value should be
+	// considered - updates that set unset fields will not be included in the diff.
+	OriginalPathsOnly bool
```
wenovus, line 370: // IgnoreNewPaths specifies that only paths that are set in the original value are
// considered in the diff -- paths that exist only in the modified struct are disregarded
// by the diff. This means that the contents of the Update field of the returned
// Notification will only contain modifications to the original value.

How about this name? I think its meaning is clearer.

ygot/diff_test.go
```diff
@@ -838,6 +838,16 @@ func TestDiff(t *testing.T) {
...
+	}, {
+		desc:   "single path addition in modified with OriginalPathsOnly set",
+		inOrig: &renderExample{},
+		inMod: &renderExample{
+			Str: String("cabernet-sauvignon"),
+		},
+		inOpts: []DiffOpt{&DiffPathOpt{
+			OriginalPathsOnly: true,
+		}},
+		want: &gnmipb.Notification{},
```
wenovus, line 850: Could you add some fields for modify & delete to make sure that the behaviour for those are unaffected by this flag?

wenovus (commented):
Implementation looks good, a naming and a testing suggestion.

ygot/diff.go
dan-lepage: Done.

ygot/diff_test.go
dan-lepage: Done.

wenovus (approved):
Thanks! @robshakir any concerns here?

ygot/diff.go
```diff
@@ -365,6 +365,11 @@ type DiffPathOpt struct {
 	// generated structs, which can result in duplication of list key leaves in
 	// the diff output.
 	MapToSinglePath bool
+	// IgnoreNewPaths specifies that only paths that are set in the original value are
```
robshakir, line 368: I think elsewhere we've tried to make there be a single field in the options rather than have multiple -- but I guess `DiffPathOpt` is a fairly generic name. I'd assume that this option should control how paths are interpreted, so we should have a new option that can just be `struct {}` that corresponds to `ModifiedOnly`.

ygot/diff.go
```diff
@@ -365,6 +365,11 @@ type DiffPathOpt struct {
...
+	// considered in the diff -- paths that exist only in the modified struct are disregarded
+	// by the diff. This means that the contents of the Update field of the returned
+	// Notification will only contain modifications to the original value.
+	IgnoreNewPaths bool
```
robshakir, line 372: I'd suggest `ModifiedOnly` or `OnlyModifiedFields` might be a clearer name.

ygot/diff.go
```diff
@@ -432,13 +439,14 @@ func Diff(original, modified GoStruct, opts ...DiffOpt) (*gnmipb.Notification, e
...
 		}
 	}
-
-	// Check that all paths that are in the modified struct have been examined, if
-	// not they are updates.
-	for modPath, modVal := range modLeaves {
-		if !matched[modPath] {
-			if err := appendUpdate(n, modPath, modVal); err != nil {
-				return nil, err
+	if pathOpt == nil || !pathOpt.IgnoreNewPaths {
```
robshakir, line 442: Suggest to reverse this logic -- i.e., `if pathOpt != nil && pathOpt.IgnoreNewPaths { return ... }`

robshakir (changes requested):

ygot/diff.go
dan-lepage: I'd worry that this would imply it only returns updated fields, not deleted ones. Would IgnoreAdditions be clearer?

robshakir: SGTM.

ygot/diff.go
dan-lepage: Ok, I made it standalone.

ygot/diff.go
dan-lepage: Done.

ygot/diff.go
dan-lepage: Done.

robshakir (approved):
LGTM, thanks for the change!

PR #560, opened 2021-08-10
Remove testutil/ from coverage report.

wenovus:
The code in this directory are used for ygot tests, so they naturally don't have tests.


wenovus:
Nvm this package is actually tested.

PR #568, opened 2021-08-12
Update goyang to v0.2.8

wenovus:
- Updated test files as https://github.com/openconfig/goyang/issues/131 is now fixed.


exampleoc/opstateoc/update.sh
```diff
@@ -35,7 +35,6 @@ go run ../../generator/generator.go -path=public,deps -output_file=oc.go \
   public/release/models/policy/openconfig-routing-policy.yang \
   public/release/models/lacp/openconfig-lacp.yang \
   public/release/models/system/openconfig-system.yang \
-  public/release/models/lldp/openconfig-lldp.yang \
```
robshakir, line 38: Why was this removed please?

robshakir (changes requested):
LGTM, but I want to understand why LLDP was removed.

exampleoc/opstateoc/update.sh
wenovus: Because it's already there (see last YANG file). There was a goyang change to error out for duplicate modules.

robshakir (approved):
Thanks!

exampleoc/opstateoc/update.sh
robshakir: Ah, apologies. I missed this.

PR #569, opened 2021-08-12
Add go:generate directives for (u)exampleoc/... generated code

wenovus:
We can now use `go generate` to generate these files, although note we still need to run the command [explicitly](https://pkg.go.dev/cmd/go/internal/generate), since go build, go test, etc. don't call the directive.


greg-dennis (approved):
Approve assuming the changes to the oc.go and ocpath.go files are intentional.

wenovus:
Yes it's a duplicate YANG file. Accidentally put it here instead of the base PR so doesn't make a difference.

PR #572, opened 2021-08-13
Support JSON Marshalling augments to compressed schemas

wenovus:
#### ygen: gogen.go, directory.go
* Extend `findSchemaPath` and `findMapPaths` to return a path of module names that correspond with the paths.
  e.g. "(shadow-)path:config/name|name" -> "(shadow-)module:modA/modB|modA"
* Removed a small existing hack whereby error from `InstantiatingModule` was being ignored.
#### ygot: render.go
* Created `appmodsJSON` to assist `structJSON` to append module names. It determines the module names that should be prepended to path elements when the JSON is assembled by `structJSON`.
* Change `structJSON` to use the output from `appmodsJSON` when determining whether to prefix node names with corresponding module names. This allows the logic to be simplified somewhat.
* Added helper `structTagToLibModules` sister to existing `structTagToLibPaths` to help with extracting the modules tag.

### Testing
* Added compressed augment unit test for `findMapPaths` and JSON marshalling.
* Added integration test for generating a compressed augment.
* Other integration tests updated to test `module` and `shadow-module` generation.

### Other Testing Changes
* Changed tests for `findSchemaPath` and `findMapPaths` to create `yang.Entry` objects using `Modules.Process` instead of manually creating them.
* Populated `.Modules` field for other tests in order to avoid errors due to above-mentioned hack being removed.


ygot/render.go
```diff
@@ -1098,7 +1126,14 @@ func structJSON(s GoStruct, parentMod string, args jsonOutputConfig) (map[string
 			continue
 		}
 
-		value, err := jsonValue(field, pmod, args)
+		// s is the fake root if its path tag is empty. In this case,
+		// we want to forward the parent module to the child nodes.
+		isFakeRoot := len(mapPaths) == 1 && mapPaths[0].Len() == 0
+		if isFakeRoot {
+			chMod = parentMod
+		}
```
wenovus, line 1134: This is essentially an alternative solution to this corner case solved here: https://github.com/openconfig/ygot/pull/73/files#diff-6acf78876eda5b3f170f96871eca978db5c8071ee952918b55d489148244e139

ygen/directory.go
```diff
@@ -137,38 +137,47 @@ func GetOrderedDirectories(directory map[string]*Directory) ([]string, map[strin
...
-	return findSchemaPath(parent, fieldName, false, absolutePaths)
+	schemaPaths, _, err := findSchemaPath(parent, fieldName, false, absolutePaths)
+	return schemaPaths, err
 }
 
 // findSchemaPath finds the relative or absolute schema path of a given field
 // of a Directory, or the shadowed field path (field duplicated and
 // deprioritized via compression) of a Directory. The Field is specified as a
 // name in order to guarantee its existence before processing.
 // NOTE: If shadowSchemaPaths is true, no error is returned if fieldName is not found.
```
robshakir, line 148: What is the 2nd return argument here that you added? It looks like the docstring here did not get updated?

ygot/render.go
robshakir: Yes, this looks a better solution.

ygot/render.go
```diff
@@ -1112,85 +1147,55 @@ func structJSON(s GoStruct, parentMod string, args jsonOutputConfig) (map[string
...
+
+		for i, p := range mapPaths {
+			if appmods != nil && p.Len() != len(appmods[i]) {
+				errs.Add(fmt.Errorf("number of paths and modules elements not the same: (paths: %v, modules: %v)", p, appmods[i]))
+				continue
+			}
+
+			parent := jsonout
+			j := 0
+			for ; j != p.Len()-1; j++ {
```
robshakir, line 1174: If `j` is being used outside of the loop context I'd call it something different. I think here something like `pathInMod`?

ygot/render_test.go
```diff
@@ -1660,33 +1660,65 @@ type invalidGoStructMap struct {
...
 type mapKey struct {
 	F1 string `path:"fOne"`
 	F2 string `path:"fTwo"`
 }
 
 type structMultiKeyChild struct {
-	F1 *string `path:"config/fOne|fOne" shadow-path:"state/fOne|fOne"`
-	F2 *string `path:"config/fTwo|fTwo" shadow-path:"state/fTwo|fTwo"`
+	F1 *string `path:"config/fOne|fOne" module:"fmod/f1mod|f1mod" shadow-path:"state/fOne|fOne" shadow-module:"fmod/f1mod|f1mod"`
+	F2 *string `path:"config/fTwo|fTwo" module:"fmod/f2mod|f2mod" shadow-path:"state/fTwo|fTwo" shadow-module:"fmod/f2mod-shadow|f2mod-shadow"`
```
robshakir, line 1705: Is this something that can actually happen? i.e., how does a compressed path end up being in two modules? Is this just to check that the `shadow-module` parsing works correctly?

ygot/struct_validation_map.go
```diff
@@ -82,6 +82,44 @@ func structTagToLibPaths(f reflect.StructField, parentPath *gnmiPath, preferShad
 	return mapPaths, nil
 }
 
+// structTagToLibModules takes an input struct field as a reflect.Type, and
+// determines the set of validation library modules that it maps to. Returns
```
robshakir, line 86: What does `validation library` mean in this context? Is it schema modules?

robshakir (changes requested):

ygen/directory.go
wenovus: Done.

ygot/render.go
wenovus: I still like `j` better because I think it's easier to understand. The reason why it needs to be outside is because the last element is a special case (it needs to set the child JSON value), so I think its natural meaning as an index is not lost.

ygot/struct_validation_map.go
wenovus: Fixed to be clearer. I copied over the other docstring and didn't proofread.

ygot/render.go
robshakir: SGTM :-)

ygot/render_test.go
robshakir: Interesting -- I hadn't thought about this case, it'd be weird, but could indeed happen.

ygot/struct_validation_map.go
robshakir: Thanks!

robshakir (approved):
LGTM, thanks for the change Wen.

PR #575, opened 2021-08-17
Hide EscapeHTML behind an option instead of default behaviour.

wenovus:
Alternatively I can remove this option altogether, let me know, I'm fine either way.


robshakir (approved):
I'm happy to leave this in place, just in case someone is relying on the content being escaped.

PR #579, opened 2021-09-02
Allow generation of multiple path struct packages

DanG100:
* Add logic to generate separate package at Directory level
* Writing files and full import resolution isn't implemented yet


ypathgen/pathgen.go
```diff
@@ -761,15 +787,28 @@ func generateDirectorySnippet(directory *ygen.Directory, directories map[string]
 		}
 		goFieldName := goFieldNameMap[fieldName]
 
-		if es := generateChildConstructors(&methodBuf, directory, fieldName, goFieldName, directories, schemaStructPkgAccessor, pathStructSuffix, listBuilderKeyThreshold, generateWildcardPaths, simplifyWildcardPaths); es != nil {
+		fieldDirectory := directories[field.Path()]
+		childAccessor := ""
+
+		if fieldDirectory != nil {
```
wenovus, line 793: if fieldDirectory := directories[field.Path()]; fieldDirectory != nil {

Also I'm not so sure about having a general functionality to find out the `childPkgAccessor`. Might it be better to only do this handling at the root, because we know that's the only place we'll ever have this handling?

ypathgen/pathgen.go
```diff
@@ -291,18 +291,31 @@ func (cg *GenConfig) GeneratePathCode(yangFiles, includePaths []string) (*Genera
...
 		}
 		structSnippets = append(structSnippets, structSnippet)
 	}
+	packages := map[string]*GeneratedPathCode{}
+	for _, snippet := range structSnippets {
+		if _, ok := packages[snippet.Package]; !ok {
+			packages[snippet.Package] = &GeneratedPathCode{}
+		}
+		packages[snippet.Package].Structs = append(packages[snippet.Package].Structs, snippet)
+		packages[snippet.Package].Deps = append(packages[snippet.Package].Deps, snippet.Deps...)
```
wenovus, line 306: not deduped?

wenovus (commented):
I'm wondering what the input to the generator will be and how that might affect your implementation. The input will be a list of files, and their imports will also need to be generated (this is possibly a bug actually, I wouldn't have expected this, I'll discuss with Rob later, so keep your previous PR).

You will then need to split the generated output into different directories. How will you do that if the input files are in completely different locations? (e.g. a/b/c, a/b/d, e/f/g?)

Would it be simpler to just output one package per module? This is easily done by checking `directory.Path[0]`.

ypathgen/pathgen.go
```diff
@@ -205,7 +205,7 @@ type GoImports struct {
 // node in the generated code, which may help callers add customized
 // augmentations to the basic generated path code.
 // If errors are encountered during code generation, they are returned.
-func (cg *GenConfig) GeneratePathCode(yangFiles, includePaths []string) (*GeneratedPathCode, NodeDataMap, util.Errors) {
+func (cg *GenConfig) GeneratePathCode(yangFiles, includePaths []string) (map[string]*GeneratedPathCode, NodeDataMap, util.Errors) {
```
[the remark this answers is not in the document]
DanG100: add

generator/generator.go
```diff
@@ -431,6 +431,9 @@ func main() {
 	if errs != nil {
 		log.Exitf("ERROR Generating PathStruct Code: %s\n", errs)
 	}
+	if len(pathCode) > 1 {
+		log.Exitf("Multiple packages aren't fully supported yet")
+	}
```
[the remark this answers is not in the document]
DanG100: removed

ypathgen/pathgen.go
```diff
@@ -722,6 +740,13 @@ type goPathFieldData struct {
 	KeyParamListStr string           // KeyParamListStr is the parameter list of the field's accessor method.
 	KeyEntriesStr   string           // KeyEntriesStr is an ordered list of comma-separated ("schemaName": unique camel-case name) for a list's keys.
 	KeyParamDocStrs []string         // KeyParamDocStrs is an ordered slice of docstrings documenting the types of each list key parameter.
+	ChildAccessor   string           // ChildAccessor is used if the child path struct exists in another package.
```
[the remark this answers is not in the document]
DanG100: fixed

ypathgen/pathgen.go
```diff
@@ -761,15 +787,28 @@ func generateDirectorySnippet(directory *ygen.Directory, directories map[string]
...
+			parentPackge := getPackageName(cg, directory, directories)
+			childPackage := getPackageName(cg, fieldDirectory, directories)
+			if parentPackge != childPackage {
+				deps = append(deps, childPackage)
+				childAccessor = "." + childPackage
+			}
+		}
+
+		if es := generateChildConstructors(&methodBuf, directory, fieldName, goFieldName, directories, schemaStructPkgAccessor,
+			cg.PathStructSuffix, listBuilderKeyThreshold, cg.GenerateWildcardPaths, cg.SimplifyWildcardPaths, childAccessor); es != nil {
```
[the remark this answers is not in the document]
DanG100: undone

ypathgen/pathgen.go
DanG100: deduped later

DanG100:
Added the rest of the logic to drive discussion. Generates one go package per yang module

ypathgen/pathgen.go
```diff
@@ -731,15 +756,15 @@ type goPathFieldData struct {
 // the fields of the struct. directory is the parsed information of a schema
 // node, and directories is a map from path to a parsed schema node for all
 // nodes in the schema.
-func generateDirectorySnippet(directory *ygen.Directory, directories map[string]*ygen.Directory, schemaStructPkgAccessor, pathStructSuffix string, listBuilderKeyThreshold uint, generateWildcardPaths, simplifyWildcardPaths bool) (GoPathStructCodeSnippet, util.Errors) {
+func generateDirectorySnippet(directory *ygen.Directory, directories map[string]*ygen.Directory, schemaStructPkgAccessor string, listBuilderKeyThreshold uint, cg *GenConfig) (GoPathStructCodeSnippet, util.Errors) {
```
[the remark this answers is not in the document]
DanG100: agreed, changed.

ypathgen/pathgen.go
DanG100: fieldDirectory is used outside the scope of that if statement, can't make that change.

i could add a check for ygen.IsFakeRoot for this

wenovus: `fieldDirectory` is only used within the if statement right? where else is it being used?

generator/generator.go
```diff
@@ -99,6 +99,10 @@ var (
 	simplifyWildcardPaths   = flag.Bool("simplify_wildcard_paths", false, "Whether to omit the keys in the generated paths if all keys for a list node are wildcards.")
 	listBuilderKeyThreshold = flag.Uint("list_builder_key_threshold", 0, "The threshold equal or over which the path structs' builder API is used for key population. 0 means infinity. This flag is only meaningful when wildcard paths are generated.")
 	pathStructSuffix        = flag.String("path_struct_suffix", "Path", "The suffix string appended to each generated path struct in order to differentiate their names from their corresponding schema struct names.")
+	splitByModule           = flag.Bool("split_by_module", false, "Whether to split path struct generation by module.")
+	fakeRootPackageName     = flag.String("fake_root_package_name", "device", "Name for the fake root package, when spilting generation by module.")
```
wenovus, line 103: pathstruct_fakeroot_package_name
Also can copy fakeroot_name above and make the default "". In the code can declare a const `defaultRootName = "device"`.

generator/generator.go
```diff
@@ -99,6 +99,10 @@ var (
 	simplifyWildcardPaths   = flag.Bool("simplify_wildcard_paths", false, "Whether to omit the keys in the generated paths if all keys for a list node are wildcards.")
 	listBuilderKeyThreshold = flag.Uint("list_builder_key_threshold", 0, "The threshold equal or over which the path structs' builder API is used for key population. 0 means infinity. This flag is only meaningful when wildcard paths are generated.")
 	pathStructSuffix        = flag.String("path_struct_suffix", "Path", "The suffix string appended to each generated path struct in order to differentiate their names from their corresponding schema struct names.")
+	splitByModule           = flag.Bool("split_by_module", false, "Whether to split path struct generation by module.")
```
wenovus, line 102: split_pathstructs_by_module

generator/generator.go
```diff
@@ -99,6 +99,10 @@ var (
...
+	fakeRootPackageName     = flag.String("fake_root_package_name", "device", "Name for the fake root package, when spilting generation by module.")
+	trimOCPackage           = flag.Bool("trim_oc_package", false, "Whether to trim openconfig- from generated package names.")
```
wenovus, line 104: trim_path_package_oc_prefix

generator/generator.go
```diff
@@ -99,6 +99,10 @@ var (
...
+	baseImportPath          = flag.String("base_import_path", "", "Base import path used concatenated with module for path struct imports.")
```
wenovus, line 105: ... used to concatenate with module package relative paths for ...

generator/generator.go
```diff
@@ -433,6 +441,17 @@ func main() {
 	}
 
 	switch {
+	case *splitByModule:
+		for packageName, code := range pathCode {
+			os.Mkdir(filepath.Join(*outputDir, packageName), 0755)
```
wenovus, line 446: Should it be 644?

generator/generator.go
```diff
@@ -433,6 +441,17 @@ func main() {
...
+			path := filepath.Join(*outputDir, packageName, "pathstructs.go")
```
wenovus, line 447: Just make it packageName.go?

ypathgen/pathgen.go
```diff
@@ -173,6 +173,11 @@ type GenConfig struct {
 	// If any key is not a wildcard, then this flag doesn't apply, since
 	// all key values must now be specified in the path.
 	SimplifyWildcardPaths bool
+	// SplitByModule controls whether to generate a go package for each yang module.
+	SplitByModule       bool
+	FakeRootPackageName string
+	TrimOCPackage       bool
+	BaseImportPath      string
```
wenovus, line 180: I'm assuming you will add comments for these?

ypathgen/pathgen.go
```diff
@@ -191,21 +196,23 @@ type GoImports struct {
...
+// a map of GeneratedPathCode struct containing all the generated code to
 // support the path-creation API. The important components of the generated
 // code are listed below:
 //	1. Struct definitions for each container, list, or leaf schema node,
 //	as well as the fakeroot.
 //	2. Next-level methods for the fakeroot and each non-leaf schema node,
 //	which instantiate and return the next-level structs corresponding to
 //	its child schema nodes.
+// The key of the map is the package name and the value is the information
+// needed to generate code for that package.
```
wenovus, line 208: Maybe this should be part of the first sentence so that it's clear from the outset what the key is?

ypathgen/pathgen.go
```diff
@@ -262,6 +266,21 @@ func (cg *GenConfig) GeneratePathCode(yangFiles, includePaths []string) (*Genera
...
+			return cg.PackageName
+		}
+		if ygen.IsFakeRoot(dir.Entry) {
+			return cg.FakeRootPackageName
+		}
+		name := util.SchemaTreeRoot(dir.Entry).Name
+		if cg.TrimOCPackage {
+			name = strings.TrimPrefix(name, "openconfig-")
+		}
+		name = strings.ReplaceAll(name, "-", "")
```
wenovus, line 280: rfc7950:
```
   identifier          = (ALPHA / "_")
                         *(ALPHA / DIGIT / "_" / "-" / ".")
```
Could you replace "_" and "." as well?

ypathgen/pathgen.go
```diff
@@ -262,6 +266,21 @@ func (cg *GenConfig) GeneratePathCode(yangFiles, includePaths []string) (*Genera
 		errs = util.AppendErrs(errs, es)
 	}
 
+	namer := func(dir *ygen.Directory, dirs map[string]*ygen.Directory) string {
```
wenovus, line 269: I'm assuming this will be made into a function?

wenovus (commented):
Thanks, this solution looks good. It would be nice next time if you just mentioned in the comments that you are splitting by modules and are no longer considering where the input YANG files reside. This speeds up the review.

I made some style suggestions but I'm assuming you will work on that now.

ypathgen/pathgen.go
wenovus: https://go.dev/blog/package-names It seems we need to make the package names lowercase.

generator/generator.go
DanG100: done

generator/generator.go
DanG100: done

generator/generator.go
DanG100: fixed

generator/generator.go
DanG100: directories need to be executable as well

generator/generator.go
DanG100: fixed

ypathgen/pathgen.go
DanG100: added

ypathgen/pathgen.go
DanG100: agreed

ypathgen/pathgen.go
DanG100: switched and regex and toLower

ypathgen/pathgen.go
DanG100: fixed

ypathgen/pathgen.go
```diff
@@ -746,13 +818,15 @@ func generateDirectorySnippet(directory *ygen.Directory, directories map[string]
...
+			return nil, util.AppendErr(errs, err)
 		}
 	} else if err := goPathStructTemplate.Execute(&structBuf, structData); err != nil {
-		return GoPathStructCodeSnippet{}, util.AppendErr(errs, err)
+		return nil, util.AppendErr(errs, err)
 	}
 
 	goFieldNameMap := ygen.GoFieldNameMap(directory)
+	deps := map[string]bool{}
+	builders := map[string]*strings.Builder{}
```
wenovus, line 829: I think this name is a bit too general. Maybe listBuilderAPIBufs?

generator/generator.go
```diff
@@ -100,6 +100,10 @@ var (
...
 	listBuilderKeyThreshold = flag.Uint("list_builder_key_threshold", 0, "The threshold equal or over which the path structs' builder API is used for key population. 0 means infinity. This flag is only meaningful when wildcard paths are generated.")
 	pathStructSuffix        = flag.String("path_struct_suffix", "Path", "The suffix string appended to each generated path struct in order to differentiate their names from their corresponding schema struct names.")
+	splitByModule           = flag.Bool("split_pathstructs_by_module", false, "Whether to split path struct generation by module.")
+	fakeRootPackageName     = flag.String("pathstruct_fakeroot_package_name", "", "Name for the fake root package, when spilting generation by module.")
+	trimOCPackage           = flag.Bool("trim_path_package_oc_prefix", false, "Whether to trim openconfig- from generated package names.")
+	baseImportPath          = flag.String("base_import_path", "", "Base import path used to concatenate with module package relative paths for path struct imports.")
```
wenovus, line 106: for path struct imports **when split_pathstructs_by_module=true**.

ypathgen/pathgen_test.go
```diff
@@ -1140,6 +1140,97 @@ func TestGeneratePathCodeSplitFiles(t *testing.T) {
...
+	}{{
+		name:    "oc simple",
+		inFiles: []string{filepath.Join(datapath, "openconfig-simple.yang")},
+		wantStructsCodeFiles: map[string]string{
+			"openconfigsimple": "testdata/modules/oc-simple/simple.txt",
+			"device":           "testdata/modules/oc-simple/device.txt",
+		},
+	}, {
+		name:         "oc simple and trim",
+		inFiles:      []string{filepath.Join(datapath, "openconfig-simple.yang")},
```
wenovus, line 1160: Do you want to test openconfig-import.yang in order to test multiple packages being written?

ypathgen/pathgen_test.go
```diff
@@ -2570,7 +3237,9 @@ func TestGenerateChildConstructor(t *testing.T) {
 		inPathStructSuffix        string
 		inGenerateWildcardPaths   bool
 		inSimplifyWildcardPaths   bool
-		want                      string
+		inChildAccessor           string
+		wantMethod                string
+		wantBuilder               string
```
wenovus, line 3242: wantListBuilderAPI?

generator/generator.go
```diff
@@ -430,6 +434,10 @@ func main() {
 		ListBuilderKeyThreshold: *listBuilderKeyThreshold,
 		GenerateWildcardPaths:   *generateWildcardPaths,
 		SimplifyWildcardPaths:   *simplifyWildcardPaths,
+		TrimOCPackage:           *trimOCPackage,
+		FakeRootPackageName:     *fakeRootPackageName,
+		SplitByModule:           *splitByModule,
+		BaseImportPath:          *baseImportPath,
```
wenovus, line 440: Maybe add a validation here that if `SplitByModule=true`, then the fake root name and the import path must not be empty.

generator/generator.go
```diff
@@ -430,6 +434,10 @@ func main() {
...
 		InputModulesOnly:        *inputModulesOnly,
```
wenovus, line 441: Add a flag validation that `splitByModule=true` only if `inputModulesOnly=true` as well.

wenovus (changes requested):
Overall looks good, some minor comments.

Thanks for adding the integration tests, I think these are pretty thorough and well tested! Did you try this on a real schema to see whether there are compilation issues? ygot doesn't yet have compilation tests for formatted.txt files for path structs.

robshakir (changes requested):
Hi Daniel, Wen -- thanks for the change here. I think there are still a number of concerns around this for me. Is there a design document that discusses the problem, and the proposed approach for the solution please (including considering marshalling, unmarshalling, and validation implications)?

generator/generator.go
DanG100: added check for import path. fake root has a default in ypathgen

ypathgen/pathgen_test.go
DanG100: added

DanG100:
@robshakir Wen and I talked offline this change doesn't affect marshaling or validating because It only affects path struct generation. I removed any dependencies on https://github.com/openconfig/ygot/pull/578. PTAL

ypathgen/pathgen.go
```diff
@@ -291,18 +303,63 @@ func (cg *GenConfig) GeneratePathCode(yangFiles, includePaths []string) (*Genera
...
+	return packages, nodeDataMap, errs
+}
+
+// packageNameReplacePattern matches all characters allowed in yang modules, but not go packages.
+var packageNameReplacePattern = regexp.MustCompile("[._-]")
+
+// goPackageName returns the go package to use when generating code for the input Directory.
+// If splitByModule is false, the pkgName is always returned. Otherwise,
+// a transformed version of the module that the directory belongs to is returned.
+func goPackageName(dir *ygen.Directory, splitByModule, trimOCPkg bool, pkgName, fakeRootPkgName string) string {
```
wenovus, line 349: Could you also mention the purpose of trimOCPkg and fakeRootPkgName?

wenovus, line 349: So today I've also received a complaint about the generated size of the GoStructs, which was an option you mentioned but we didn't investigate at the time due to the extra complexity of GoStructs. I took a look today and I think it's possible to also split GoStructs in the future while still having a top-level package that users can refer to. Since this is the case, maybe the package names for the PathStructs should have "path" appended to them to keep them separate from GoStructs. I like the idea of keeping them separate to further reduce the chances of a package that's too big.

ypathgen/pathgen.go
```diff
@@ -714,24 +785,30 @@ func getStructData(directory *ygen.Directory, pathStructSuffix string, generateW
...
+	ChildPkgAccessor string           // ChildPkgAccessor is used if the child path struct exists in another package.
 }
 
 // generateDirectorySnippet generates all Go code associated with a schema node
 // (container, list, leaf, or fakeroot), all of which have a corresponding
-// struct onto which to attach the necessary methods for path generation. The
-// code comprises of the type definition for the struct, and all accessors to
+// struct onto which to attach the necessary methods for path generation.
+// When generating code for the fakeroot, several structs may be returned,
+// one for snippet and one for list builder APIs in another package. Otherwise,
```
wenovus, line 803: one package containing the fake root struct and one package for each of the fake root's child lists that uses the builder API methods, since they must be defined in their respective child packages.

ypathgen/pathgen.go
```diff
@@ -791,21 +893,41 @@ func generateDirectorySnippet(directory *ygen.Directory, directories map[string]
...
+	}
+	return snippets, errs
 }
 
 // generateChildConstructors generates and writes to methodBuf the Go methods
-// that returns an instantiation of the child node's path struct object. It
-// takes as input the buffer to store the method, a directory, the field name
+// that returns an instantiation of the child node's path struct object.
+// When this is called on the fakeroot, the list Builder API methods need to be
+// in another package than the rest of the methods. In all other cases, methodBuf and
```
wenovus, line 923: the list builder API's methods need to be in the child package as opposed to the fakeroot's package.

wenovus (changes requested):

ypathgen/pathgen.go
wenovus: So essentially the idea is to split the GoStructs, and then use type aliasing at the top-level package to give the illusion that they're part of the same package.

ypathgen/pathgen.go
DanG100: done

ypathgen/pathgen.go
DanG100: done

ypathgen/pathgen.go
DanG100: added _path suffix

DanG100: added

ypathgen/pathgen.go
```diff
@@ -173,6 +177,17 @@ type GenConfig struct {
...
+	SplitByModule bool
+	// FakeRootPackageName is the name of package that contains the fakeroot, only
+	// used when splitting generated code by yang module.
+	FakeRootPackageName string
+	// TrimOCPackage controls whether to trim openconfig- from generated go package names.
+	TrimOCPackage bool
+	// BaseImportPath is used to create to full import path of the generated go packages.
+	BaseImportPath string
+	// InputModulesOnly configures whether to generate code for yang modules in the searched paths.
+	InputModulesOnly bool
```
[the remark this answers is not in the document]
DanG100: missed that when merging. removed

wenovus:
Looks like the integration .txt files need to be updated for the _path change.

ypathgen/pathgen_test.go
```diff
@@ -1288,6 +1387,7 @@ func getSchemaAndDirs() (*yang.Entry, map[string]*ygen.Directory, map[string]map
 			"container-with-config": nil,
 			"list":                  nil,
 			"list-with-state":       nil,
+			"import":                {NativeType: "string"},
```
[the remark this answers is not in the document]
DanG100: leftover from experimentation removed.

ypathgen/pathgen_test.go
```diff
@@ -1040,6 +1040,104 @@ func TestGeneratePathCodeSplitFiles(t *testing.T) {
...
 }
 
+func TestGeneratePathCodeSplitModules(t *testing.T) {
+	tests := []struct {
+		name                      string   // Name is the identifier for the test.
+		inFiles                   []string // inFiles is the set of inputFiles for the test.
+		inIncludePaths            []string // inIncludePaths is the set of paths that should be searched for imports.
+		inTrimOCPath              bool
+		inListBuilderKeyThreshold uint
+		wantStructsCodeFiles      map[string]string // wantStructsCodeFileDir map from package name to want source file.
```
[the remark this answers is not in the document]
DanG100: done

ypathgen/pathgen.go
```diff
@@ -883,10 +1009,12 @@ func generateChildConstructorsForLeafOrContainer(methodBuf *strings.Builder, fie
 
 // generateChildConstructorsForListBuilderFormat writes into methodBuf the
 // child constructor method snippets for the list represented by listAttr using
-// the builder API format. fieldData contains the childConstructor template
+// the builder API format. The builder API helpers are written into the builderBuf,
+// this allows the helpers to be written to package the child belongs to, not the parent.
```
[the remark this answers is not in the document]
DanG100: done

ypathgen/pathgen.go
```diff
@@ -62,6 +63,9 @@ const (
 	// NOTE: This cannot be "", as the builder method name would conflict
 	// with the child constructor method for the keys.
 	BuilderKeyPrefix = "With"
+	// defaultFakeRootPkgName is the default name for the package that contains
+	// the fake root struct (if splitting by module is enabled).
+	defaultFakeRootPkgName = "device"
```
wenovus, line 68: Sorry for not catching this earlier. I'm thinking instead of putting the fake root inside its own package, that we should still write the fake root in `ocPathStructsOutputFile`, which is the same file we'd be writing in if we didn't split packages. We should also make sure `generatePathStructsMultipleFiles` in generator.go is false so that we're guaranteed `ocPathStructsOutputFile` exists.

ypathgen/pathgen.go
```diff
@@ -173,6 +177,15 @@ type GenConfig struct {
 	// If any key is not a wildcard, then this flag doesn't apply, since
 	// all key values must now be specified in the path.
 	SimplifyWildcardPaths bool
+	// SplitByModule controls whether to generate a go package for each yang module.
+	SplitByModule bool
+	// FakeRootPackageName is the name of package that contains the fakeroot, only
+	// used when splitting generated code by yang module.
+	FakeRootPackageName string
```
wenovus, line 184: If the above comment is adopted, then we can drop this option and simply use `GenConfig.PackageName` for the fake root's package name.

ypathgen/pathgen.go
DanG100: i think that may be a bit confusing because if split by modules is true then you need to specify both output_dir and path_struct_output_file. Also fake root generation is optional. so `ocPathStructsOutputFile` needs be optional as well. The fake root is already its own package. What's the goal of writing the fake root `ocPathStructsOutputFile`?

wenovus: I agree it makes the flags even more complicated. However, I'm thinking there is value in separating out the API that's accessible to users (i.e. the fake root's package), and the API that's not accessible. I believe fakeroot generation for PathStructs is **mandatory**, indeed `ygot.ResolvePath` relies on this fact. I don't see how users would use PathStructs without the fakeroot, so I think it's reasonable to assume it'll always be there.

Therefore, I think we need to check that if split-by-modules is true, that both flags are set. What do you think?

DanG100: sgtm, added checks and update flag docs.

ypathgen/pathgen.go
DanG100: done

ypathgen/pathgen_test.go
```diff
@@ -1757,13 +1861,15 @@ func TestGenerateDirectorySnippet(t *testing.T) {
...
+		inSplitByModule           bool
+		want                      []GoPathStructCodeSnippet
+		wantNoWildcard            []GoPathStructCodeSnippet
 	}{{
 		name:        "container-with-config",
 		inDirectory: directories["/root-module/container-with-config"],
-		want: GoPathStructCodeSnippet{
+		want: []GoPathStructCodeSnippet{{
 			PathStructName: "ContainerWithConfig",
+			Package:        "device",
```
wenovus, line 1872: Can you also test with a different package name? I suspect there is some handling required in generator.go to write the fakeroot's package differently from the other packages. Maybe you can fallthrough to the next case.

wenovus (changes requested):

ypathgen/pathgen_test.go
DanG100: well that logic was missing, added

ypathgen/pathgen.go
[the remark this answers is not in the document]
DanG100: done

generator/generator.go
```diff
@@ -433,6 +445,24 @@ func main() {
...
+	case *splitByModule:
+		for packageName, code := range pathCode {
+			// The fake root package is written to ocPathStructsOutputFile.
+			// All other packages are written to outdir/<package>.
+			path := *ocPathStructsOutputFile
+			if packageName != pcg.PackageName {
+				if err := os.MkdirAll(filepath.Join(*outputDir, packageName), 0755); err != nil {
+					log.Exitf("failed to create directory for package %q: %v", packageName, err)
+				}
+				filepath.Join(*outputDir, packageName, fmt.Sprintf("%s.go", packageName))
```
wenovus, line 457: Is this a typo? `path = filepath.Join...`?

ypathgen/pathgen.go
```diff
@@ -791,21 +882,41 @@ func generateDirectorySnippet(directory *ygen.Directory, directories map[string]
...
+
+	for pkg, build := range listBuilderAPIBufs {
+		if code := build.String(); code != "" {
+			snippets = append(snippets, GoPathStructCodeSnippet{
+				PathStructName:    structData.TypeName,
+				Package:           pkg,
+				ChildConstructors: code,
+			})
+		}
+
```
wenovus, line 904: nit: extra newline

wenovus (approved):
Thanks for the work! This LGTM since you've tested this manually. Ideally we have a way of testing that the generated files work, I'm planning on taking a look at #584. I'll also work on #580 at some point to reduce the overhead of adding a change for ypathgen.

robshakir (approved):
Thanks for the changes and clarity @DanG100 and @wenovus - LGTM!

PR #590, opened 2021-09-16
Don't import ywrapper and yext proto files if not used.

### LINKED ISSUE #432: Remove protobuf unused imports automatically
`ygen` generated protobufs may have unused `ywrapper`, or `yext` imports.  In future versions of `protoc`, this may be considered an error. `ygen` should remove these imports automatically.

wenovus:
Fixes #432 

Implementation:
- Added `UsesYwrapperImport` and `UsesYextImport` fields to `Proto3Package` and `generatedProto3Message` to help accumulate possible places where these packages are used in the generated code.
- Inherit these fields from child messages for hierarchically-generated messages.

Testing:
- Manually verified all integration test proto files use the right imports.
- (corner case) Added a `decimal64` to a union (`oneof`) field for `union-list-key.yang` to test `ywrapper` import.

Test table:
| Imports        | enums.proto | others.proto |
| -----------       | ----------- | ------------ |
| neither           | proto-test-d.uncompressed.enums.formatted-txt | proto-test-a.compress.parent.formatted-txt |
| yext only         | enum-union.compress.enums.formatted-txt | enum-union.compress.formatted-txt |
| ywrapper only     |   N/A | proto-test-a.compress.parent.child.formatted-txt |
| yext and ywrapper |   N/A | nested-messages.nested_messages.formatted-txt |

Note that this table is not exhaustive, as there are also nested/oneof test cases. As it would create a lot of maintenance burden to add new YANG files to test the permutation of all input cases, I've tried to simplify the implementation as much as possible and then verify that commenting out any statement that set `UsesYxxxImport` causes at least one test to fail.


robshakir (approved):
Thanks for cleaning this up! LGTM.

PR #592, opened 2021-09-29
add GoPackage to NodeData

DanG100:
Add GoPackage field to path NodeData so that downstream generators can use the package name.

### DIFF
--- a/ypathgen/pathgen.go
+++ b/ypathgen/pathgen.go
@@ -288,7 +288,7 @@ func (cg *GenConfig) GeneratePathCode(yangFiles, includePaths []string) (map[str
 				HasDefault:            false,
 				YANGTypeName:          "",
 				YANGPath:              "/",
-				GoPackage:             goPackageName(directory, cg.SplitByModule, cg.TrimOCPackage, cg.PackageName),
+				GoPathPackageName:     goPackageName(directory, cg.SplitByModule, cg.TrimOCPackage, cg.PackageName),
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


ypathgen/pathgen.go
```diff
@@ -485,6 +486,8 @@ type NodeData struct {
 	YANGTypeName string
 	// YANGPath is the schema path of the YANG node.
 	YANGPath string
+	// GoPackage is the Go package name of a schema node.
```
wenovus, line 489: `NodeData` represents both GoStruct and PathStruct information. So maybe something more specific:

"GoPathPackageName is the Go package name containing the generated PathStruct for the schema node."

ypathgen/pathgen.go
```diff
@@ -616,9 +619,11 @@ func mustTemplate(name, src string) *template.Template {
 // schema field names of that directory entry (i.e. the same keys as the
 // "Fields" map of the Directory entry). Since ygen provides a *MappedType for
 // every leaf node only, leafTypeMap's value is nil for non-leaf nodes.
+// packageName, splitByModule, and trimOCPackage are used to generate the name of
+// Go package for each output node.
```
wenovus, line 623: are used to determine the generated Go package name for the generated PathStructs.

wenovus (approved):

ypathgen/pathgen.go
DanG100: done

ypathgen/pathgen.go
DanG100: done

PR #593, opened 2021-09-29
[ypathgen] ResolvePath should return a non-nil PathElem message.

wenovus:
Per the spec it should be `path := []*PathElem{}` rather than `nil` as
it currently is.
https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md#constructing-paths


greg-dennis (changes requested):
Add a unit test?

greg-dennis (approved):

PR #601, opened 2021-10-26
Add util function for wildcard path matching

DanG100:
* Enable less-strict path comparison for paths with wildcard keys


util/gnmi.go
```diff
@@ -53,12 +55,43 @@ func PathElemsEqual(a, b *gpb.PathElem) bool {
...
+		if vo, ok := b.Key[k]; !ok || v != vo {
+			return false
+		}
+	}
+	return true
+}
+
+// PathElemsEqualWildcard checks if two PathElem are the same allowing for wildcard keys.
+// If a.Key["foo"] == "*" and a.Key["foo"] == "bar" func returns true.
+// If a has no keys, it is to equal to b regardless of what keys b has and vice versa.
```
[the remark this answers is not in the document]
robshakir: We probably should add something to the spec that defines this, but my expectation would be that `null` and `*` are the same thing - so that we say, for the list:

```yang
list foo { 
  key "one two";
}
```

* `/foo[one=*][two=*]`
* `/foo[one=*]`
* `/foo[two=*]`
* `/foo`

all mean the same thing (essentially, all entries in `foo`).

I didn't quite understand Wen's comment of "they are not on exactly the same keys" -- Wen, could you expand here?

Dan -- keys in YANG are mandatory to be specified, does your "unset key" here refer to in the supplied user wildcarded path?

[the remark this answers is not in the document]
wenovus: It's irrelevant now. I believe I was saying how if the ordered pair of ("*", null, ...) between the two inputs don't match exactly, e.g. ("*", null) and (null, "*"), the current function would fail. But if we agree that `null` and `*` are the same thing, then they should be equal.

util/gnmi.go
```diff
@@ -53,12 +55,43 @@ func PathElemsEqual(a, b *gpb.PathElem) bool {
...
-		if vo, ok := b.Key[k]; !ok || vo != v {
+		if vo, ok := b.Key[k]; !ok || v != vo {
+			return false
+		}
+	}
+	return true
+}
+
+// PathElemsEqualWildcard checks if two PathElem are the same allowing for wildcard keys.
+// If a.Key["foo"] == "*" and a.Key["foo"] == "bar" func returns true.
```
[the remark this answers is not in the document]
DanG100: outdated now

util/gnmi.go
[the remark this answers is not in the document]
DanG100: thanks for clarification, i was talking about the user supplied path

DanG100:
Refactored this a bunch and simplified to follow the use case better. The goal of this func to check if a received gNMI path is valid based on the query path.

util/gnmi.go
```diff
@@ -43,6 +43,8 @@ func PathMatchesPrefix(path *gpb.Path, prefix []string) bool {
 }
 
 // PathElemsEqual replaces the proto.Equal() check for PathElems.
+// Note: Use PathElemsEquivalent for wildcards.
```
wenovus, line 46: PathElemsEquivalent -> PathMatchesQuery

util/gnmi_test.go
```diff
@@ -526,6 +526,184 @@ func TestPathMatchesPathElemPrefix(t *testing.T) {
...
+				Name: "six",
+			}},
+		},
+		inQuery: &gpb.Path{
+			Elem: []*gpb.PathElem{{
+				Name: "three",
+				Key:  map[string]string{"seven": "eight"},
+			}},
+		},
+	}}
```
wenovus, line 697: Another invalid test case: wildcard query but key is not present in return value.

util/gnmi.go
```diff
@@ -92,6 +97,31 @@ func PathMatchesPathElemPrefix(path, prefix *gpb.Path) bool {
...
+// Both paths must use the gNMI >=0.4.0 PathElem path format.
+func PathMatchesQuery(path, query *gpb.Path) bool {
+	if len(path.GetElem()) < len(query.GetElem()) || path.Origin != query.Origin {
+		return false
+	}
+	for i, queryElem := range query.Elem {
+		pathElem := path.Elem[i]
+		if queryElem == nil || pathElem == nil {
+			return false
+		}
```
wenovus, line 112: Could you document this corner case behaviour?

wenovus (approved):

util/gnmi.go
DanG100: adde to godoc

util/gnmi_test.go
DanG100: added

util/gnmi.go
DanG100: moved  that comment to the PathMatchesPathElemPrefix

PR #607, opened 2021-11-25
Add Go generation flag `generate_populate_defaults`.

### LINKED ISSUE #101: Mechanism to access default values

wenovus:
```
If set to true, a PopulateDefault method will be generated for all GoStructs which recursively
populates default values.
```

`(&exampleoc.Device{}).PopulateDefaults()` Benchmark:
BenchmarkPopulateDefaults-12               18048             67670 ns/op

Seems to run within 100 us.

FIXED #101 (issue contains design discussion)


ygen/gogen.go
```diff
@@ -300,6 +300,20 @@ type generatedLeafGetter struct {
...
+// a PopulateDefaults method for a GoStruct that recursively populates default
+// values within the subtree.
+type generatedDefaultMethod struct {
+	// Receiver is the name of the receiver for the default method.
+	Receiver string
+	// ChildContainerNames are the names of the container fields of the GoStruct.
+	ChildContainerNames []string
+	// ChildContainerNames are the names of the list fields of the GoStruct.
+	ChildListNames []string
+	// Leaf represent the leaf fields of the GoStruct.
```
robshakir, line 313: Comment should use the same name as the field. Suggest `Leaves` or `Leaf`.

ygen/gogen.go
```diff
@@ -1298,10 +1347,12 @@ func writeGoStruct(targetStruct *Directory, goStructElements map[string]*Directo
 	var associatedListMethods []*generatedGoListMethod
 
 	// associatedLeafGetters is a slice of structs which define the set of leaf getters
-	// to generated for the struct. It is only populated if the GenerateLeafGetters option
-	// is set to true.
+	// to generated for the struct.
 	var associatedLeafGetters []*generatedLeafGetter
 
+	var associatedDefaultMethod generatedDefaultMethod
```
robshakir, line 1353:
```golang
associatedDefaultMethod = generatedDefaultMethod{
   Receiver: targetStruct.Name,
}
```

ygen/gogen.go
```diff
@@ -1517,19 +1570,17 @@ func writeGoStruct(targetStruct *Directory, goStructElements map[string]*Directo
 				enumTypeMap[schemapath] = append(enumTypeMap[schemapath], mtype.NativeType)
 			}
 
-			if goOpts.GenerateLeafGetters {
```
robshakir, line 1520: Do we always generate the getters in the output code, I think the answer here is yes - since these two features are now dependent upon each other? Should we just replace them with a single flag (i.e., should `generateLeafGetters` be deprecated?)

ygot/schema_tests/schema_benchmark_test.go
```diff
@@ -0,0 +1,14 @@
+package schematest
+
+import (
+	"testing"
+
+	"github.com/openconfig/ygot/exampleoc"
+)
+
+func BenchmarkPopulateDefaults(b *testing.B) {
```
robshakir, line 9: Thanks for adding the benchmark -- how does this perform?

robshakir (changes requested):

ygen/codegen_test.go
```diff
@@ -1237,6 +1243,10 @@ func TestSimpleStructs(t *testing.T) {
 			wantCode := string(wantCodeBytes)
 
 			if gotCode != wantCode {
+				// FIXME(wenbli): debug codegen or pathgen
+				if err := ioutil.WriteFile(fmt.Sprintf("/usr/local/google/home/wenbli/tmp/%s", path.Base(tt.wantStructsCodeFile)), []byte(gotCode), 0644); err != nil {
+					panic(err)
```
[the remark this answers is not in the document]
wenovus: Oops I forgot to delete this.

ygen/gogen.go
wenovus: Done.

ygen/gogen.go
wenovus: Done.

ygen/gogen.go
wenovus: We don't always generate the getters -- this particular guard is simply a run-time optimization, because this same guard is present when `generateLeafGetters`, i.e. the function that populates the template, is called.

I removed this optimization because this data is also required for `PopulateDefaults`, and it has no effect on getter generation.

ygot/schema_tests/schema_benchmark_test.go
wenovus: BenchmarkPopulateDefaults-12 18048 67670 ns/op

Seems to run within 100 us.

robshakir (approved):

PR #610, opened 2021-12-02
Field numbers for multi-key leafs pointing to the same field should be unique.

wenovus:
This OpenConfig snippet is causing duplicate field numbers to be generated due to the list keys both pointing to the same path:
```yang
list inter-area-propagation-policy {
  key "src-area dst-area";
  description
    "A list of connections between pairs of areas - routes are
    propagated from the source (src) area to the destination (dst)
    area according to the policy specified";

  leaf src-area {
    type leafref {
      path "../config/src-area";
    }
    description
      "Reference to the source area";
  }

  leaf dst-area {
    type leafref {
      path "../config/dst-area";
    }
    description
      "Reference to the destination area";
  }
...
...
...
leaf src-area {
  type leafref {
    // we are at ospf/global/inter-area-propagation-policies/...
    // inter-area-propagation-policy/config/src-area
    path "../../../../../areas/area/identifier";
  }
  description
    "The area from which prefixes are to be exported.";
}

leaf dst-area {
  type leafref {
    // we are at ospf/global/inter-area-propagation-policies/...
    // inter-area-propagation-policy/config/src-area
    path "../../../../../areas/area/identifier";
  }
  description
    "The destination area to which prefixes are to be imported";
}
```

This PR makes `genListKeyProto` use the non-resolved entry path when determining the field number of its oneof fields.

The new `proto-test-g.yang` file has the following diff after this change:

```diff
--- want
+++ got
@@ -28,12 +28,12 @@
   }
   message ElistKey {
     oneof one {
-      sint64 one_sint64 = 380657501;
-      string one_string = 434013286;
+      sint64 one_sint64 = 278603474;
+      string one_string = 515705169;
     }
     oneof two {
-      sint64 two_sint64 = 380657501;
-      string two_string = 434013286;
+      sint64 two_sint64 = 322077556;
+      string two_string = 157954079;
     }
     Elist elist = 3;
   }
```

I've verified that two fields in a container that both point to the same leaf doesn't have this problem.


ygen/protogen.go
```diff
@@ -1328,6 +1329,10 @@ func unionFieldToOneOf(fieldName string, e *yang.Entry, mtype *MappedType, annot
 	}
 	sort.Strings(typeNames)
 
+	if path == "" {
```
robshakir, line 1332: Rather than add this here - suggest that we just make the caller where you explicitly specify `""` use args.field.Path()?

ygen/testdata/proto/proto-test-e.uncompressed.proto-test-e.formatted-txt
```diff
@@ -21,8 +21,8 @@ message AnimalKey {
     SPECIESENUM_OVIS_CANADENSIS = 3;
   }
   oneof species {
-    SpeciesEnum species_speciesenum = 102559808;
-    string species_string = 236397324;
+    SpeciesEnum species_speciesenum = 97881385;
```
robshakir, line 24: This change seems to be unrelated to the description of the change - since now these `oneof` fields have changing field numbers. Is there a reason that's the case (it'll cause backwards incompatible protos to be generated).

robshakir (changes requested):

ygen/testdata/proto/proto-test-e.uncompressed.proto-test-e.formatted-txt
wenovus: Previously we used the leafref's resolved path (which points to a union) to determine the field numbers. Now we use just the leafref's path to avoid colliding. Since this is backwards incompatible, I'll change it so that only when this collision happens does the behaviour change.

ygen/protogen.go
wenovus: Done.

ygen/testdata/proto/proto-test-e.uncompressed.proto-test-e.formatted-txt
robshakir: I'm missing something here. Sorry.

So we're calling:

```golang
ft, err := fieldTag(fmt.Sprintf("%s_%s", e.Path(), strings.ToLower(tn)))
```

when we generate the `oneof` - in this case `e.Path()` would be the path to the field that defines the leafref, not the leafref's specified `path` argument -- so `/list/one` or `/list/two` if there was a `list` with keys called `one` and `two`. In this case, shouldn't the values be different here since we're hashing with `/list/one_string` and `/list/two_string` for the fields of the union?

What was the clash that was occurring previously?

I'm a little concerned about only doing it when there is a conflict since this has the potential to rely on the ordering of the schema (which we usually don't depend on).

wenovus: The issue occurs when `/list/one` and `/list/two` both point to the same path somewhere else, say `/container/foo`, as exemplified by the new integration test case. This is due to these lines, which calls `resolveLeafrefTarget`, causing both to hash on `/container/foo`, thus creating the collision:
https://github.com/openconfig/ygot/blob/299b0613d66380f125ebb1836392d6e8448bd4d1/ygen/protogen.go#L1170-L1182
The original fix is to use `/list/one` and `/list/two` as you pointed out, but it may introduces a backwards compatibility issue so I then proposed to use `/container/foo` and `/list/two`.

The OP's example has an indirect issue, where the leafref points to a leafref and which collides. This becomes an issue when `compress_paths=true`, because `FindAllChildren` promotes the config/state leaf to the directory.

robshakir: Thanks for the explanation. I'd missed that there was a different context for generating the key protobufs.

I think this fix should be stable, since the `key` field in YANG is ordered and must be kept in the same order, so we'll always hash on `two`. I prefer the option of keeping things backwards compatible, but I wonder if it's counter-intuitive in the future. Should we add a comment that just links to this PR discussion so we remember why?

Thanks again for walking me through the code :-)

wenovus: Np, added comment now with link to this thread.

wenovus:
@robshakir Ping on this, I've added comments per your last comment.

robshakir:
Apologies, thought I'd approved.

robshakir (approved):

PR #614, opened 2022-01-05
Add option to call `SetNode` with a JSON-encoded value.

wenovus:
Currently `SetNode` only allows gNMI typed values (aka `GNMIEncoding`)
to be unmarshalled; however, it is possible for a gNMI server implementation to receive
JSON-encoded leaves, which ygot currently doesn't provide the facility
for unmarshalling into a GoStruct. This change adds an option for JSON-encoded leaf
values to be unmarshalled using `SetNode`.

### DIFF
--- a/ytypes/node.go
+++ b/ytypes/node.go
@@ -51,10 +51,14 @@ type retrieveNodeArgs struct {
 	// If val is set to a non-nil value, leaf/leaflist node corresponding
 	// to the given path is updated with this value.
 	val interface{}
+	// useJSONEncoding indicates that the input value is in JSON encoding
+	// instead of gNMI's typed encoding.
+	useJSONEncoding bool
 	// tolerateJSONInconsistenciesForVal means to tolerate inconsistencies
 	// for val as if it were converted from JSON. As of right now, this is
 	// specifically to deal with uint values being streamed as positive int
-	// values.
+	// values. If useJSONEncoding is true, then this option is not
+	// applicable.
 	tolerateJSONInconsistenciesForVal bool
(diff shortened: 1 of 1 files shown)


ytypes/node.go
```diff
@@ -51,10 +51,14 @@ type retrieveNodeArgs struct {
 	// If val is set to a non-nil value, leaf/leaflist node corresponding
 	// to the given path is updated with this value.
 	val interface{}
+	// useJSONEncoding indicates that the input value is in JSON encoding
+	// instead of gNMI's typed encoding.
+	useJSONEncoding bool
```
robshakir, line 56: General question -- should we support this mode like this, or rather using the `TypedValue` field that corresponds to JSON?

robshakir (changes requested):
Two questions:

1. Are there any implementation changes needed to support this?
2. Do we have existing test coverage? It seems like we should at least have some test cases that cover the argument being provided?

Cheers!
r.

ytypes/node.go
wenovus: Thanks for catching this, I agree that we should just drop this mode and just add support for unmarshalling `TypedValue`s that contain json_ietf. This fits the current use case better as well.

The use case I'm looking at is that in the current Ondatra implementation, `Marshal7951` is used to marshal messages instead of `EncodeTypedValue`, which means leaves are also in JSON format. Do you think Ondatra should be using `EncodeTypedValue` instead?

wenovus: Since this implementation is very different from just supporting `json_ietf`, I will drop this and open another PR.

PR #618, opened 2022-02-05
Support leaf-list defaults using latest version of goyang.

wenovus:
Note: I haven't released the goyang commit this is dependent on yet
(hence the go.mod replace statement) since I want to release both repos
at the same time. This is to avoid a downstream issue:
https://github.com/openconfig/models-ci/issues/59

- Update ygot to use the updated `Entry.Default []string` and its
  corresponding helpers for storing defaults for leaf-lists.
- Change some helper function such as `yangDefaultValueToGo()` to return
  `string` instead of `*string`. There is no good reason for returning a
  pointer and makes the code less convenient.
- Factor out `generateGoDefaultValue()` from `writeGoStruct()`, and have it
  handle multiple default values.
- In protogen's logic to generate default enums, check whether an enum
  is not a leaf-list before setting a particular one as the default enum
  value.
  **Note**: I'm not 100% sure whether this is the correct handling -- it
  appears to me that having a default enum value doesn't make sense for
  leaf-lists as repeated fields.

Testing:
- Added integration tests for Go and proto generation.


robshakir (approved):

wenovus:
@dplore I added you as another pair of eyes. Feel free to leave some high-level comments/questions. The context here is that goyang had a backwards-incompatible change (allowing defaults for leaf-lists): https://github.com/openconfig/goyang/pull/139

This PR adapts ygot to be able to generate defaults, see this integration test:
- YANG file: https://github.com/openconfig/ygot/pull/618/files#diff-f45a11b9b5a0743f8a99241dd8f771371baaa07733949948120c10e31f2776e2
- generated Go file: https://github.com/openconfig/ygot/pull/618/files#diff-e5177422bfad27fda267f0aa3e622bcb7af8f6896ad0149b24ad05fdcb647bbe

wenovus:
I'm going to merge this with Rob's approval.

PR #619, opened 2022-02-07
Update ygot to use hermetic module parsing change from goyang.

wenovus:
Note: I haven't released the goyang commit this is dependent on yet
(hence the go.mod replace statement) since I want to release both repos
at the same time. This is to avoid a downstream issue:
openconfig/models-ci#59


dplore (approved):
LGTM

Issue #623, opened 2022-03-08
If a leaf-list has a relative path leafref, validation fails


hansthienpondt:
Hi,

Considering the following [yang model](https://github.com/openconfig/ygot/files/8207481/bug.yang.txt), validation fails with following debugLibrary output:

```
DataNodeAtPath got leafref with path elem:{name:".."} elem:{name:".."} elem:{name:".."} elem:{name:"two"} elem:{name:"targetlist"} elem:{name:"targetname"} from node path /device/root/one/referencelist/target, field name Target
going up data tree from type string to []string, schema path from parent is [target], remaining path elem:{name:".."} elem:{name:".."} elem:{name:"two"} elem:{name:"targetlist"} elem:{name:"targetname"}
going up data tree from type []string to *hansygot.Uncompressed_Root_One_Referencelist, schema path from parent is [target], remaining path elem:{name:".."} elem:{name:"two"} elem:{name:"targetlist"} elem:{name:"targetname"}
going up data tree from type map[string]*hansygot.Uncompressed_Root_One_Referencelist to *hansygot.Uncompressed_Root_One, schema path from parent is [referencelist], remaining path elem:{name:"two"} elem:{name:"targetlist"} elem:{name:"targetname"}
root element type *hansygot.Uncompressed_Root_One with remaining path elem:{name:"two"} elem:{name:"targetlist"} elem:{name:"targetname"}
. GetNode next path name:"two", value { map[key:0xc000199160] (map ptr) }
. getNodesContainer: schema one, next path name:"two", value { map[key:0xc000199160] (map ptr) }
childSchema for schema one, field Referencelist, tag referencelist
RelativeSchemaPath yields [referencelist]
traversing schema Dirs.../referencelist - found
. check field name referencelist, paths [[referencelist]]
. ERR: could not find path in tree beyond schema node one, (type *hansygot.Uncompressed_Root_One), remaining path elem:{name:"two"} elem:{name:"targetlist"} elem:{name:"targetname"}
. Validate with value { { { map[key:0xc000199160] (map ptr ptr ptr) }, { map[mypolicy:0xc000010c38] (map ptr ptr ptr) } } }, type *hansygot.Device, schema name device
. validateContainer with value { { { map[key:0xc000199160] (map ptr ptr ptr) }, { map[mypolicy:0xc000010c38] (map ptr ptr ptr) } } }, type *hansygot.Device, schema name device
childSchema for schema device, field Root, tag root
RelativeSchemaPath yields [root]
[...]

PR #625, opened 2022-03-10
Fix leaf-list validation when it has a relative path leafRef.

hansthienpondt:
@wenovus as per your comments in https://github.com/openconfig/ygot/issues/623 submitting the PR.

Thanks for the help!

### DIFF
--- a/ytypes/leafref.go
+++ b/ytypes/leafref.go
@@ -189,6 +189,12 @@ func dataNodesAtPath(ni *util.NodeInfo, path *gpb.Path, pathQueryNode *util.Path
 			if root.Parent == nil {
 				return nil, fmt.Errorf("no parent for leafref path at %v, with remaining path %s", ni.Schema.Path(), path)
 			}
+			if !util.IsCompressedSchema(root.Schema) && root.Parent.Schema.IsLeafList() {
+				// Workaround Leaflist, not sure if this is breaking anything?
+				root = root.Parent
+				pathQueryRoot = pathQueryRoot.Parent
+				continue
+			}
 			if !util.IsCompressedSchema(root.Schema) && root.Parent.Schema.IsList() && util.IsValueMap(root.Parent.FieldValue) {
 				// If we are in an uncompressed schema, then we have one more level of the data tree than
 				// the YANG expects, since our data tree layout is:
--- a/ytypes/leafref.go
+++ b/ytypes/leafref.go
@@ -189,13 +189,7 @@ func dataNodesAtPath(ni *util.NodeInfo, path *gpb.Path, pathQueryNode *util.Path
 			if root.Parent == nil {
 				return nil, fmt.Errorf("no parent for leafref path at %v, with remaining path %s", ni.Schema.Path(), path)
 			}
-			if !util.IsCompressedSchema(root.Schema) && root.Parent.Schema.IsLeafList() {
-				// Workaround Leaflist, not sure if this is breaking anything?
-				root = root.Parent
-				pathQueryRoot = pathQueryRoot.Parent
-				continue
-			}
-			if !util.IsCompressedSchema(root.Schema) && root.Parent.Schema.IsList() && util.IsValueMap(root.Parent.FieldValue) {
+			if !util.IsCompressedSchema(root.Schema) && ((root.Parent.Schema.IsList() && util.IsValueMap(root.Parent.FieldValue)) || (root.Parent.Schema.IsLeafList() && util.IsValueSlice(root.Parent.FieldValue))) {
 				// If we are in an uncompressed schema, then we have one more level of the data tree than
 				// the YANG expects, since our data tree layout is:
 				// struct (parent container)
--- a/util/debug.go
+++ b/util/debug.go
@@ -28,11 +28,11 @@ var (
 	// traversal. Since this setting causes global variables to be manipulated
 	// controlling the output of the library, it MUST NOT be used in a setting
 	// whereby thread-safety is required.
-	debugLibrary = true
+	debugLibrary = false
 	// debugSchema controls the debugging output from the library from schema
 	// matching code. Generates lots of output, so this should be used
 	// selectively per test case.
-	debugSchema = true
+	debugSchema = false
 	// maxCharsPerLine is the maximum number of characters per line from
 	// DbgPrint and DbgSchema. Additional characters are truncated.
 	maxCharsPerLine = 1000
--- a/util/debug.go
+++ b/util/debug.go
@@ -28,11 +28,11 @@ var (
 	// traversal. Since this setting causes global variables to be manipulated
 	// controlling the output of the library, it MUST NOT be used in a setting
 	// whereby thread-safety is required.
-	debugLibrary = false
+	debugLibrary = true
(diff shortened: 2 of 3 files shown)


ytypes/leafref.go
```diff
@@ -189,7 +189,7 @@ func dataNodesAtPath(ni *util.NodeInfo, path *gpb.Path, pathQueryNode *util.Path
 			if root.Parent == nil {
 				return nil, fmt.Errorf("no parent for leafref path at %v, with remaining path %s", ni.Schema.Path(), path)
 			}
-			if !util.IsCompressedSchema(root.Schema) && root.Parent.Schema.IsList() && util.IsValueMap(root.Parent.FieldValue) {
+			if !util.IsCompressedSchema(root.Schema) && ((root.Parent.Schema.IsList() && util.IsValueMap(root.Parent.FieldValue)) || (root.Parent.Schema.IsLeafList() && util.IsValueSlice(root.Parent.FieldValue))) {
```
wenovus, line 192: So I think it might actually be
```go
(!util.IsCompressedSchema(root.Schema) && (root.Parent.Schema.IsList() && util.IsValueMap(root.Parent.FieldValue)))
|| (root.Parent.Schema.IsLeafList() && util.IsValueSlice(root.Parent.FieldValue))
```
Since leaf-lists are never compressed.

We should also add a test for this in `TestValidateLeafRefDataCompressedSchemaListOnly` to make sure we test that the correct thing happens inside a compressed schema. I can do that in a supplemental PR if you want.

ytypes/leafref.go
```diff
@@ -189,7 +189,7 @@ func dataNodesAtPath(ni *util.NodeInfo, path *gpb.Path, pathQueryNode *util.Path
...
 				// If we are in an uncompressed schema, then we have one more level of the data tree than
 				// the YANG expects, since our data tree layout is:
 				// struct (parent container)
```
wenovus, line 195: Let's update the comment to something clearer. I had trouble understanding this.

```go
				// If we are in an uncompressed schema, then we have one more level of the Go data tree traversed by ForEachField than
				// the YANG representation, since our data tree layout is:
				// struct (parent container)
				//  --> map (the list), or slice (the leaf-list)
				//  --> struct (the list member), or scalar (the leaf-list member)
				//
				// In YANG, .. from the list member struct gets us to the parent container, but for us
				// we have only reached the map. This means that we end up over-consuming the ".."s. To
				// avoid this issue, if we are in an uncompressed schema, and in a list, and we find
				// that we're looking at the map, we consume another level of the data tree. This gets
				// us to the parent container with ".." as would be expected.
				//
				// This is NOT required for the compressed schema, because in this case, we have removed
				// a level of the data tree. So the parent container in the above example will have been
				// removed. This is enforced by ygen. In this case, we do want to consume the extra level
				// of ..s in the list case, such that we do not end up under-consuming them.
```

wenovus (changes requested):

ytypes/leafref.go
hansthienpondt: Ack, will submit tomorrow, will take a stab at adding a test in TestValidateLeafRefDataCompressedSchemaListOnly

ytypes/leafref.go
hansthienpondt: Ack, will submit tomorrow

ytypes/leafref.go
hansthienpondt: I've done some testing, and it seems when compression is enabled we need a special use-case for single item lists, as they seem to be compressed into a single struct level?

hansthienpondt: **Case1**
DataNodeAtPath got leafref with path elem:{name:".."} elem:{name:".."} elem:{name:".."} elem:{name:".."} elem:{name:"conf3"} from node path /device/root/examples/example/config/conf3-ref, field name Conf3Ref
going up data tree from type []string to *hansygot.Root_Example, schema path from parent is [config], remaining path elem:{name:".."} elem:{name:".."} elem:{name:".."} elem:{name:"conf3"}
going up data tree from type *hansygot.Root_Example to map[uint32]*hansygot.Root_Example, schema path from parent is [example], remaining path elem:{name:".."} elem:{name:".."} elem:{name:"conf3"}
going up data tree from type map[uint32]*hansygot.Root_Example to *hansygot.Root, schema path from parent is [examples], remaining path elem:{name:".."} elem:{name:"conf3"}
going up data tree from type *hansygot.Root to *hansygot.Device, schema path from parent is [root], remaining path elem:{name:"conf3"}
root element type *hansygot.Device with remaining path elem:{name:"conf3"}
. GetNode next path name:"conf3", value { { conf2 (string ptr ptr ptr), [ conf3 (string ptr ptr) ], map[42:0xc0002426c0] (map ptr ptr) } }
. getNodesContainer: schema device, next path name:"conf3", value { { conf2 (string ptr ptr ptr), [ conf3 (string ptr ptr) ], map[42:0xc0002426c0] (map ptr ptr) } }


**Case2**
DataNodeAtPath got leafref with path elem:{name:".."} elem:{name:".."} elem:{name:".."} elem:{name:".."} elem:{name:"conf3"} from node path /device/root/examples/example/config/conf3-ref, field name Conf3Ref
going up data tree from type []string to *hansygot.Root_Examples_Example, schema path from parent is [config], remaining path elem:{name:".."} elem:{name:".."} elem:{name:".."} elem:{name:"conf3"}
going up data tree from type *hansygot.Root_Examples_Example to map[uint32]*hansygot.Root_Examples_Example, schema path from parent is [example], remaining path elem:{name:".."} elem:{name:".."} elem:{name:"conf3"}
going up data tree from type map[uint32]*hansygot.Root_Examples_Example to *hansygot.Root_Examples, schema path from parent is [example], remaining path elem:{name:".."} elem:{name:"conf3"}
going up data tree from type *hansygot.Root_Examples to *hansygot.Root, schema path from parent is [examples], remaining path elem:{name:"conf3"}
root element type *hansygot.Root with remaining path elem:{name:"conf3"}
. GetNode next path name:"conf3", value { conf2 (string ptr ptr), [ conf3 (string ptr) ], { nil, map[42:0xc00027e6f0] (map ptr ptr) } }
. getNodesContainer: schema root, next path name:"conf3", value { conf2 (string ptr ptr), [ conf3 (string ptr) ], { nil, map[42:0xc00027e6f0] (map ptr ptr) } }

wenovus: Ok I think I've figured it out. I'm going to push to your branch with the fixes.

wenovus:
@hansthienpondt I've ported your PR and opened another one: #626, do you mind signing the CLA there? Thanks.

ytypes/leafref.go
wenovus: So regarding your YANG example, compression is specific to OpenConfig, which requires every list to be wrapped underneath a container with no other sibling nodes.

ytypes/leafref.go
wenovus: I've updated this comment in the new forked PR.

hansthienpondt:
closing this PR as the fix is provided by @wenovus in https://github.com/openconfig/ygot/pull/626

PR #626, opened 2022-03-14
Fix leaf-list validation when it is a relative path leafref

wenovus:
credits to @hansthienpondt for finding the issue and doing the initial debugging and fix experimentation in #625. I copied over his fork.

## Changes
- Fix leaf-list validation when it is a relative path leafref
- Simplify `util.forEachFieldInternal` to not trim compressed path elements for maps or slices.

Currently, we trim the path for a `map[]{}` or slice type in Go for lists in order to prevent the list's name from appearing twice in the data tree while traversing a list element. This is because `forEachFieldInternal` currently creates an extra level of `NodeInfo` elements, copying the list name, when traversing list elements, so it uses two levels of data nodes to represent a YANG list.

Based on the tests that fail, this behaviour apparently tries to solve the problem of how to process just a single "../" when calling `ytypes.dataNodesAtPath` when there are two levels of `NodeInfo`s, but the logic in there incorrectly identifies compression to be the issue: compression is indeed not the issue, but rather how slices and maps are processed in two levels -- slices and maps exist equally in compressed and uncompressed generated code.

This PR removes the dependency of existing logic on schema compression, and instead looks at whether the current element is a map or a slice which corresponds to a YANG list or leaf-list, and in such cases, it skips that extra level created by `forEachFieldInternal`. This results in slightly simpler, and IMO much more understandable logic with the comment updates.

### DIFF
--- a/ytypes/leafref.go
+++ b/ytypes/leafref.go
@@ -196,6 +196,9 @@ func dataNodesAtPath(ni *util.NodeInfo, path *gpb.Path, pathQueryNode *util.Path
 				// element level. Since it does this by creating a "fake", or extra NodeInfo for each
 				// element, we need to skip this level of NodeInfo and instead directly use the NodeInfo
 				// of the parent (i.e. the map or slice) to avoid processing this extra NodeInfo.
+				//
+				// Note here that since lists and leaf-lists are represented the same way in compressed
+				// vs. uncompressed code, this logic is the same regardless of compression.
 				root = root.Parent
 				pathQueryRoot = pathQueryRoot.Parent
 				continue
--- a/ytypes/leafref.go
+++ b/ytypes/leafref.go
@@ -189,6 +189,12 @@ func dataNodesAtPath(ni *util.NodeInfo, path *gpb.Path, pathQueryNode *util.Path
 			if root.Parent == nil {
 				return nil, fmt.Errorf("no parent for leafref path at %v, with remaining path %s", ni.Schema.Path(), path)
 			}
+			if !util.IsCompressedSchema(root.Schema) && root.Parent.Schema.IsLeafList() {
+				// Workaround Leaflist, not sure if this is breaking anything?
+				root = root.Parent
+				pathQueryRoot = pathQueryRoot.Parent
+				continue
+			}
 			if !util.IsCompressedSchema(root.Schema) && root.Parent.Schema.IsList() && util.IsValueMap(root.Parent.FieldValue) {
 				// If we are in an uncompressed schema, then we have one more level of the data tree than
 				// the YANG expects, since our data tree layout is:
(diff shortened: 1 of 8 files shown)


wenovus:
@hansthienpondt do you mind signing the CLA since I'm basing this PR off of your initial changes?

hansthienpondt:
@wenovus done, I guess I can close the initial PR i've submitted?

wenovus:
> @wenovus done, I guess I can close the initial PR i've submitted?

Yes that's right, thanks.

ytypes/leafref_test.go
```diff
@@ -152,7 +152,7 @@ func TestValidateLeafRefData(t *testing.T) {
 						Kind: yang.LeafEntry,
 						Type: &yang.YangType{
 							Kind: yang.Yleafref,
-							Path: "../../../leaf-list",
```
robshakir, line 155: Was this test case buggy? I guess it was the wrong error message, but this never actually did a valid lookup.

ytypes/leafref.go
```diff
@@ -189,23 +189,13 @@ func dataNodesAtPath(ni *util.NodeInfo, path *gpb.Path, pathQueryNode *util.Path
...
-				// we have only reached the map. This means that we end up over-consuming the ".."s. To
-				// avoid this issue, if we are in an uncompressed schema, and in a list, and we find
-				// that we're looking at the map, we consume another level of the data tree. This gets
-				// us to the parent container with ".." as would be expected.
-				//
-				// This is NOT required for the compressed schema, because in this case, we have removed
-				// a level of the data tree. So the parent container in the above example will have been
-				// removed. This is enforced by ygen. In this case, we do want to consume the extra level
-				// of ..s in the list case, such that we do not end up under-consuming them.
+			if (root.Parent.Schema.IsList() && util.IsValueMap(root.Parent.FieldValue)) || (root.Parent.Schema.IsLeafList() && util.IsValueSlice(root.Parent.FieldValue)) {
```
robshakir, line 192: Ah, this is interesting -- as long as the current test cases pass then I'm good with this. I think the iteration around this code has been that confusion as to whether this is the case in a compressed schema vs. an uncompressed one.

Inspecting the code, I'm not quite sure without reloading full context how we avoid the issue that we are at `/foo/surrounding-container/list/element` and have a reference to `/foo/bar`. In this case `element` can have a leafref `../../../bar` but in the generated compressed structs we don't have the `surrounding-container`. I /think/ this is handled because we're iterating the schema and avoiding that level, but I just want to confirm that this case is still handled with this change.

ytypes/leafref_test.go
```diff
@@ -712,12 +742,32 @@ func TestValidateLeafRefDataCompressedSchemaListOnly(t *testing.T) {
...
 			in: &Root{
 				Example: map[uint32]*RootExample{42: {Conf: Uint32(42), ConfRef: Uint32(42)}},
 			},
 		},
+		{
+			desc: "conf-ref-list",
+			in: &Root{
+				Example: map[uint32]*RootExample{42: {Conf: Uint32(42), ConfRefList: []*uint32{Uint32(42)}}},
+			},
+		},
```
robshakir, line 770: These cases look to cover what I was concerned about above.

robshakir (approved):

PR #632, opened 2022-03-23
[cosmetic]: Change relevant instances of append -> prepend

wenovus:
This is a cosmetic change that attempts to reduce the confusion of the usage of "append" when we're actually prepending the module names.

- Change all relevant instances of `appmod` and similar variables to `prependmod` and similar names.
- Change all relevant instances of "append" to "prepend" in doc comments when talking about prepending module names.


robshakir (approved):
👍🏼 for accuracy, thanks Wen!

PR #635, opened 2022-03-25
Fix leaf-list Unmarshal to replace instead of appending values.

wenovus:
This is because gNMI doesn't allow for specifying deleting any specific leaf-list element.

I put `clearSliceField` in `ytypes` to avoid exposing this as a public function.

### DIFF
--- a/ytypes/leaf_list.go
+++ b/ytypes/leaf_list.go
@@ -100,6 +100,11 @@ func unmarshalLeafList(schema *yang.Entry, parent interface{}, value interface{}
 		return err
 	}
 
+	fieldName, _, err := schemaToStructFieldName(schema, parent)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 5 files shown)


ytypes/leaf_list.go
```diff
@@ -119,6 +124,8 @@ func unmarshalLeafList(schema *yang.Entry, parent interface{}, value interface{}
 		if len(sa.LeaflistVal.GetElement()) == 0 {
 			return fmt.Errorf("unmarshalLeafList for schema %s: value %v: got empty leaf list, expect non-empty leaf list", schema.Name, util.ValueStr(value))
 		}
+		// A new leaf-list update specifies the entire leaf-list
```
robshakir, line 127: Suggest something like `// We are supplied the entire leaf-list, so we should clear its contents if it is non-nil`.

robshakir (approved):
LGTM, thanks @wenovus

ytypes/leaf_list.go
wenovus: thanks.

PR #638, opened 2022-04-07
Don't prepend module name for in-namespace fields of a non-root-level GoStruct

### LINKED ISSUE #637: JSON Marshal functions shouldn't prepend the module name for in-namespace fields of a non-root-level GoStruct
ConstructIETFJSON, Marshal7951, EmitJSON currently always initializes the `parentMod` to `""` when marshalling a GoStruct, regardless of whether it is the fake root or not. This means that for non-root-level GoStructs, all fields have their belonging-module's name prepended in their JSON object-member names. This appears to violate RFC7951:

```
   Every data node instance is encoded as a name/value pair where the
   name is formed from the data node identifier using the rules of
   [Section 4](https://datatracker.ietf.org/doc/html/rfc7951#section-4).
   The value depends on the category of the data node, as
   explained in the following subsections.
```
([ref](https://datatracker.ietf.org/doc/html/rfc7951#section-5))

One way to solve this is by adding a method (e.g. `ΛBelongingModule()`) to each GoStruct to determine its belonging-module, such that we can treat lower-level structs differently when marshalling GoStructs.

@robshakir would be good to get your thoughts on this one before we fix this. The RFC doesn't provide a very clear example of it but I think it's fair to interpret it this way.

wenovus:
For details see #637

- Added `ΛBelongingModule()` method for all `GoStruct`s to achieve this. Go doesn't have class-level attributes and methods seem a good way of getting this information.
- Added unit tests in ygot/.
- Fixed demo BGP IETF-JSON output.
- Fixed ytypes/schema_tests JSON: this is an unrelated fix due to the BGP models changing.

Fixes #637


wenovus:
Another thought I have is to return `GoStructInfo()` (yes without the lambda so that it's easily accessible) and have it return a bunch of information, of which belonging-module is just one. For example we may want to know the origin of the GoStruct as well. instantiating-module, list vs. container, op-state vs. config, absolute YANG path are just some of the things we could return.

robshakir:
I tend to prefer the lambda-prefixed unix-style extensions to get the subset of information we need, I think these API tend to be cleaner, and easier for consumers to use.

ygen/gogen.go
```diff
@@ -1379,6 +1392,17 @@ func writeGoStruct(targetStruct *Directory, goStructElements map[string]*Directo
 		YANGPath:   util.SlicePathToString(targetStruct.Path),
 	}
 
+	// Non-fakeroot GoStructs all have a module to which they belong per
+	// RFC7950 namespace rules. The fakeroot is assigned "".
+	if !IsFakeRoot(targetStruct.Entry) {
+		im, err := targetStruct.Entry.InstantiatingModule()
+		if err != nil {
```
robshakir, line 1399: Should this be fatal? If there is an entry that is defined that doesn't have an instantiating module, then this seems problematic - such that we can never actually render this correctly according to 7951.

robshakir: Suggest we address this by just returning the slice of errors here.

ygot/render.go
```diff
@@ -1025,6 +1042,28 @@ func Marshal7951(d interface{}, args ...Marshal7951Arg) ([]byte, error) {
 	return js, nil
 }
 
+func hasBelongingModule(s GoStruct) bool {
+	return reflect.ValueOf(s).MethodByName("ΛBelongingModule").IsValid()
+}
+
+// belongingModuleName returns the name of the belonging module of the input GoStruct.
```
robshakir, line 1049: If we added this to the `ValidatedGoStruct` interface, I think that you could then just call this directly. I guess the downside of doing this is that users need to regenerate their bindings code at the same time as upgrading, but it seems cleaner in general.  (The argument to say that bindings should be regenerated is that this defect is not resolved unless they both regenerate the code and upgrade ygot).

robshakir (changes requested):

wenovus:
> I tend to prefer the lambda-prefixed unix-style extensions to get the subset of information we need, I think these API tend to be cleaner, and easier for consumers to use.

What are unix-style extensions?

ygen/gogen.go
wenovus: I believe all nodes in YANG belong to a namespace, so I think it's not possible for this to return an error if everything is working properly: https://datatracker.ietf.org/doc/html/rfc7950#section-6.2.1

SGTM with returning the slice of errors directly.

robshakir:
> > I tend to prefer the lambda-prefixed unix-style extensions to get the subset of information we need, I think these API tend to be cleaner, and easier for consumers to use.
> 
> What are unix-style extensions?

Sorry - by unix-style I meant "small things that do one thing well" :-)

ygot/render.go
wenovus: I'm thinking that `ΛBelongingModule` is not required by validation. If so then should it still be part of the interface definition?

robshakir: Good question -- `ValidatedGoStruct` was created because at some point we had users that didn't care about validation, and `GoStruct` needed to coexist with it. At this point, I'm not sure that logic is correct -- essentially someone working with a `ValidatedGoStruct` really probably cares whether they are working with a type generated by ygot or not. If this is the intention of the interface (rather than it being intended to show that this is a validatable struct) then I'd suggest extending it for reasons that don't immediately link to validation seems fine.

That being said, I'm not sure `ValidatedGoStruct` today covers all the helper methods that we would generate by default.

WDYT here?

wenovus: I've always been confused by this interface so thanks for clarifying. I agree to add to the interface, it looks like the purpose of this interface is to expose more methods to the user rather than for validation only.

It makes sense that the user should always be using the same version of ygot to generate and operate on the GoStructs, so I don't see it as a problem that we require users to update both at the same time. The two are coupled and make assumptions about each other and this mutual dependency is hard to break and not necessary to break.

robshakir: SGTM :-)

If we're agreeing with this - we should perhaps add any method that we /always/ generate to this interface, and perhaps update its description to just say that it is implemented by any ygot-generated struct.

We could probably say that `GoStruct` is deprecated at this point.

wenovus: Done.

ygen/gogen.go
wenovus: Done.

robshakir (approved):
LGTM, thanks!

Issue #642, opened 2022-04-13
<validatedgostruct>.Validate() shadowing "Validate" yang model field


steiler:
The Nokia SROS yang model e.g. (https://github.com/nokia/7x50_YangModels) defines a validate leafs which causes issues with the chosen ygot naming.

Here is the reference to the leafs.
```
 find ./7x50_YangModels/latest_sros_22.2/ -name "*.yang" | xargs grep -A 5 -n "leaf validate "
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-groups.yang:190278:                                            leaf validate {
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-groups.yang-190279-                                                type boolean;
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-groups.yang-190280-                                                default "false";
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-groups.yang-190281-                                                description "Allow NETCONF validate operation";
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-groups.yang-190282-                                            }
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-groups.yang-190283-
--
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-system-security-aaa.yang:1044:                            leaf validate {
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-system-security-aaa.yang-1045-                                type boolean;
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-system-security-aaa.yang-1046-                                default "false";
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-system-security-aaa.yang-1047-                                description "Allow NETCONF validate operation";
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-system-security-aaa.yang-1048-                            }
./7x50_YangModels/latest_sros_22.2/nokia-submodule/nokia-conf-system-security-aaa.yang-1049-
```

It would be good to have a less generic name for the actual Validate() method, which would not that easy interfear with yang definitions.

robshakir:
[...]

PR #643, opened 2022-04-13
rename Validate to ΛValidate

steiler:
To overcome the issues regarding shadowing a Validate yang leaf as mentioned in #642.
This PR changes the name of the Validate function to YgotValidate(...). besides adjusting all the demos, tests and examples the following files contain the essential changes.

- ygot/types.go
- ygot/struct_validation_map.go
- ygen/gogen.go

Is there a chance this PR will be accepted? Or does anyone else have a different approach solving this issue without introducing this type of incompatibility?


robshakir:
I'm supportive of a change of this type - but only through a flag. `Validate` should be the default name for this function for simplicity and backwards compatibility.

I'd therefore:
 * declare a `string` variable that is set to a constant default value of `Validate`.
 * have a flag that is added to `generator.go` that sets this within the `GoOpts` that are handed to `GenerateGoCode`.

wenovus:
A problem with simply renaming `Validate` is that `ValidatedGoStruct` is an interface widely used in the marshalling/unmarshalling/validation helpers. This interface encompasses all methods that are generated for all Go structs regardless of generation flags. What this means is you wouldn't be able to use these helpers anymore. Indeed this handicap will become quite severe since the `GoStruct` interface is soon to be deprecated.

I agree with Rob on the addition of a flag, but I think we'll also need to change the definition of `ValidatedGoStruct` to use `ΛValidate(...ValidationOption) error` and then just have `Validate` or `YgotValidate` forward the call to this function. This allows the helpers to continue working while allowing the name change.

You would also need to update a bunch of places in existing code within ygot that calls `Validate()` to call `ΛValidate()` instead.

@robshakir Does this sound ok to you?

steiler:
So I'm not bound to YgotValidate, if it is `ΛValidate()` instead thats fine.
Question then is, should it be a string flag to accept a "proxy function name" or would it be alright, to go with a bool flag which swiches on the additional `Validate()` function which itself simply calls `ΛValidate()` this flag would be true by default and can be set to false to avoid creating the `Validate()` function.
I'm just thinking that I would like to avoid adding even more generated code, with regards to the size of e.g. the SROS models, then necessary.

wenovus:
I'd propose that if the flag is the empty string `""`, then you don't generate the proxy function at all.

steiler:
I have now implemented the validation method as ΛValidate(...) and the parameter for defining the proxy function name as `-validate_fn_name=Foo`

wenovus:
Thanks, could you fix the tests? If you run `go test ./...` in the root directory I think it should catch it without having to run GitHub Actions.

steiler:
Thanks for the hint Wen, all tests seem to work fine now.

wenovus (approved):
LGTM, thanks!

wenovus:
Actually could you run gofmt -s -w?

wenovus (changes requested):
Sorry, approved a bit optimistically. Could you augment one of the tests in gogen_test (e.g. "simple single leaf mapping test") with GoOpt you added so we have a unit test for it.

Also, I don't see the `Validate` proxy functions in exampleoc/, so could you run `go generate ./...` there and re-generate these schemas?

steiler:
Alright @wenovus, thanks.
I think I covered all comments now.

wenovus:
Thanks, LGTM could you fix the test here?

wenovus (approved):

PR #644, opened 2022-04-14
Add code to generate IR (not used).

wenovus:
`buildDirectoryDefinitions` and `findEnumSet` have now been converted to be callable by `GenerateIR`.

Some minor changes to the `IR`'s definition have been made. I will wait for the refactoring is done before modifying the design doc on the IR.

Later PRs will convert ypathgen, Go, and proto generation to use the IR instead of the Directory map.

This code is based on prototyping at https://github.com/openconfig/ygot/tree/restruct-ygen-proto, which itself is based on the prototype at https://github.com/openconfig/ygot/tree/restruct-ygen-interface

Testing:  
* A few integration tests have been added.


ygen/codegen.go
```diff
@@ -549,16 +575,15 @@ func (dcg *DirectoryGenConfig) GetDirectoriesAndLeafTypes(yangFiles, includePath
 		return nil, nil, errs
 	}
 
-	dirsToProcess := map[string]*yang.Entry(mdef.directoryEntries)
-
-	enumSet, _, errs := findEnumSet(mdef.enumEntries, cg.TransformationOptions.CompressBehaviour.CompressEnabled(), false, cg.ParseOptions.SkipEnumDeduplication, cg.TransformationOptions.ShortenEnumLeafNames, cg.TransformationOptions.UseDefiningModuleForTypedefEnumNames, cg.GoOptions.AppendEnumSuffixForSimpleUnionEnums, false, cg.TransformationOptions.EnumOrgPrefixesToTrim)
+	enumSet, _, errs := findEnumSet(mdef.enumEntries, opts.TransformationOptions.CompressBehaviour.CompressEnabled(), !opts.TransformationOptions.EnumerationsUseUnderscores, opts.ParseOptions.SkipEnumDeduplication, opts.TransformationOptions.ShortenEnumLeafNames, opts.TransformationOptions.UseDefiningModuleForTypedefEnumNames, opts.AppendEnumSuffixForSimpleUnionEnums, opts.UseConsistentNamesForProtoUnionEnums, opts.TransformationOptions.EnumOrgPrefixesToTrim)
```
robshakir, line 578: (discussion) There are a lot of arguments to this function -- I wonder whether it's worth creating a `findEnumSetOpts` struct that is handed as an argument to try and make calling it a bit neater. The only reason to do this is probably readability, but it then comes with some of the concern of what happens if you don't consume a field (which is why we have it this way today). 

What are your thoughts on introducing such a struct?

ygen/codegen.go
```diff
@@ -653,21 +678,33 @@ func (cg *YANGCodeGenerator) GenerateProto3(yangFiles, includePaths []string) (*
...
+	opts := IROptions{
+		ParseOptions:          cg.Config.ParseOptions,
+		TransformationOptions: cg.Config.TransformationOptions,
+		NestedDirectories:     true,
+		AbsoluteMapPaths:      true,
+		// If UseConsistentNamesForProtoUnionEnums=true, then also set
+		// appendEnumSuffixForSimpleUnionEnums=true for consistent union enum
+		// names.
+		AppendEnumSuffixForSimpleUnionEnums:  cg.Config.ProtoOptions.UseConsistentNamesForProtoUnionEnums,
+		UseConsistentNamesForProtoUnionEnums: cg.Config.ProtoOptions.UseConsistentNamesForProtoUnionEnums,
```
robshakir, line 691: Should we just make `UseConsistentNamesforProtoUnionEnums` always set `AppendEnumSuffixForSimpleUnionEnums` to be true -- and not really present this as two options? Is there any reason you would want inconsistently named enumerations?

ygen/codegen.go
```diff
@@ -653,21 +678,33 @@ func (cg *YANGCodeGenerator) GenerateProto3(yangFiles, includePaths []string) (*
...
-	// names.
-	enumSet, penums, errs := findEnumSet(mdef.enumEntries, cg.Config.TransformationOptions.CompressBehaviour.CompressEnabled(), true, cg.Config.ParseOptions.SkipEnumDeduplication, cg.Config.TransformationOptions.ShortenEnumLeafNames, cg.Config.TransformationOptions.UseDefiningModuleForTypedefEnumNames, cg.Config.ProtoOptions.UseConsistentNamesForProtoUnionEnums, cg.Config.ProtoOptions.UseConsistentNamesForProtoUnionEnums, cg.Config.TransformationOptions.EnumOrgPrefixesToTrim)
+
+	opts := IROptions{
+		ParseOptions:          cg.Config.ParseOptions,
+		TransformationOptions: cg.Config.TransformationOptions,
+		NestedDirectories:     true,
+		AbsoluteMapPaths:      true,
+		// If UseConsistentNamesForProtoUnionEnums=true, then also set
+		// appendEnumSuffixForSimpleUnionEnums=true for consistent union enum
```
robshakir, line 688:
```suggestion
		// AppendEnumSuffixForSimpleUnionEnums=true for consistent union enum
```

ygen/directory.go
```diff
@@ -133,6 +137,124 @@ func GetOrderedDirectories(directory map[string]*Directory) ([]string, map[strin
 	return orderedDirNames, dirNameMap, nil
 }
 
+// GetOrderedPathDirectories returns an alphabetically-ordered slice of
+// Directory names and a map of Directories keyed by their paths, so that each
+// directory can be processed in path-alphabetical order. This helps produce
+// deterministic generated code, and minimize diffs when compared with expected
+// output (i.e., diffs don't appear simply due to reordering of the Directory
+// maps).
+func GetOrderedPathDirectories(directory map[string]*Directory) []string {
```
robshakir, line 146: What are our thoughts on using generics here? I guess we should not until we do not support Go <1.18?

```go
func sortedStringMap[T any](m map[string]T) []string {
	s := []string{}
	for k := range m {
		s = append(s, k)
	}
	sort.Strings(s)
	return s
}
```

ygen/directory.go
```diff
@@ -114,6 +114,10 @@ func GoFieldNameMap(directory *Directory) map[string]string {
 // expected output (i.e., diffs don't appear simply due to reordering of the
 // Directory maps). If the names of the directories are not unique, which is
 // unexpected, an error is returned.
+// TODO(wenbli): this function's purpose is to check for name conflicts. This
```
robshakir, line 117: Is this implication here that we should remove this code after the IR refactor is complete (or at least the check that the length of the two maps is the same)?

ygen/directory.go
```diff
@@ -133,6 +137,124 @@ func GetOrderedDirectories(directory map[string]*Directory) ([]string, map[strin
...
+				Name: name,
+				YANGDetails: YANGNodeDetails{
+					Name:         field.Name,
+					Defaults:     field.DefaultValues(),
+					Module:       mod,
+					Path:         field.Path(),
+					ResolvedPath: target.Path(),
+				},
+				MappedPaths:         mp,
+				MappedModules:       mm,
```
robshakir, line 224: Question here, why `MappedModules`? I find the name here it a little confusing, since it's not clear that we "map" to a module - just that a node is defined by that module, right?

The "map" in paths was because this field is "mapped" to one or more output YANG paths (just remembering what the original intent there was).

ygen/enumgen.go
```diff
@@ -127,6 +128,17 @@ func (s *enumSet) enumeratedUnionEntry(e *yang.Entry, compressPaths, noUnderscor
 				return nil, err
 			}
 
+			var enumKind EnumeratedValueType
+			if len(enumNameSake.Type) > 0 {
+				if util.IsYANGBaseType(e.Type) {
```
robshakir, line 133: You can likely write this in a more succinct way:


```go
enumKind = DerivedEnumerationType
if len(enumNameSake.Type) > 0 {
  enumKind = DerivedUnionEnumerationType
  if util.IsYANGBaseType(e.Type) {
     enumKind = UnionEnumerationType
  }
}
```

or alternatively:

```go
switch {
  case len(enumNameSake.Type) > 0 && util.IsYANGBaseType(e.Type):
     enumKind = UnionEnumerationType
  case len(enumNameSake.Type) > 0:
     enumKind = DerivedUnionEnumerationType
  default:
     enumKind = DerivedEnumerationType
}
```

the latter seems cleaner to me.

ygen/genir.go
```diff
@@ -22,6 +32,32 @@ type IROptions struct {
...
+
+	// NestedDirectories specifies whether the generated directories should
+	// be nested in the IR.
+	NestedDirectories bool
+
+	// AbsoluteMapPaths specifies whether the path annotation provided for
+	// each field should be relative paths or absolute paths.
+	AbsoluteMapPaths bool
+
+	// AppendEnumSuffixForSimpleUnionEnums appends an "Enum" suffix to the
```
robshakir, line 44: Is it worth cutting our v1 at the point that we have integrated the IR? I'm just wondering whether we're carrying complexity that we could potentially avoid here?

ygen/ir.go
```diff
@@ -158,11 +197,30 @@ type NodeDetails struct {
...
 	// be mapped to in the output code - these annotations can be
 	// used to annotation the output code with the field(s) that it
 	// corresponds to in the YANG schema.
-	MapPaths [][]string
+	MappedPaths [][]string
+	// MappedModules describes the path elements' belonging modules that
+	// the output node should be mapped to in the output code - these
+	// annotations can be used to annotation the output code with the
+	// field(s) that it corresponds to in the YANG schema.
+	MappedModules [][]string
```
robshakir, line 209: Ah -- sorry, this is clearer as to why "mapped" is in the name. Maybe we can call this `MappedPathElementModules` - although that is quite wordy.... I'm OK with `MappedModules`.

ygen/ir.go
```diff
@@ -158,11 +197,30 @@ type NodeDetails struct {
...
+	// Shadow paths are paths that have sibling config/state values
+	// that have been compressed out due to path compression.
+	ShadowMappedPaths [][]string
+	// ShadowMappedModules describes the shadow path elements' belonging
+	// modules (if any) that the output node should be mapped to in the
+	// output code - these annotations can be used to annotation the output
+	// code with the field(s) that it corresponds to in the YANG schema.
+	// Shadow paths are paths that have sibling config/state values
+	// that have been compressed out due to path compression.
+	ShadowMappedModules [][]string
```
robshakir, line 223: Is there any case in which `ShadowMappedModules` is not equal to `MappedModules`? It seems like there should not be such a case because it would mean that we were doing compression across modules for OpenConfig (which wouldn't be allowable), but I guess this theoretically can happen through augmenting in a `state` leaf for a `config` leaf and with compression enabled?

robshakir (commented):
Thanks for this -- a few discussion comments in-line but generally this LGTM.

It's quite a big code review, but it seems like the minimum size you could have made it given this is a big refactor. I'm OK if you want to split up - e.g., the tests into a separate PR - since I don't really see any other way to split it. That being said -- we probably want the tests to be there so that we can validate the code that we're adding.

Thanks!

ygen/codegen.go
wenovus: `AppendEnumSuffixForSimpleUnionEnums` fixes a version of inconsistent union names for Go generated code only. The fixes for proto enum naming was done all at once, so that's why it's both or neither: refs:

* https://github.com/openconfig/ygot/pull/474
* https://github.com/openconfig/ygot/pull/483

ygen/directory.go
wenovus: Yes that's right. I'll add to the TODO here to deprecate this API once the refactor is complete.

ygen/directory.go
wenovus: Yeah I'm planning to not touch generics until 1.20 (next year) since ygot currently supports the latest 3 versions of Go. Another consideration is the Go team is saying generics is not completely production ready.

ygen/directory.go
wenovus: I'll change it to `MappedPathModules`. I named it `MappedModules` because of congruency, so that it's clear to the user these two are related, which I think might be the quickest way for the reader to understand their meanings.

I was considering what the naming should be here. I actually wanted to merge the two strings into a single struct but it was too cumbersome to change all the existing code, and I found I frequently needed to convert to `[]string` for processing. I could define a type to do that but I thought keeping it like this was simpler so I left it be.

ygen/enumgen.go
wenovus: Done.

ygen/genir.go
wenovus: Yes we are carrying complexity, but it's not a whole lot. My thinking on v1 right now is to get all the pending v1 work done so that we can drop as much complexity at that time and it'll also mean we'll likely maintain it for a while. Otherwise it either means we'll still have to carry the complexity from having to maintain those flags, or if we release v2 too soon I think it wouldn't give people a good sense of security about the codebase.

ygen/ir.go
wenovus: Ok I changed it to `MappedPathModules`, maybe it's a good compromise.

ygen/ir.go
wenovus: You mentioned one way. Another way is if the `ShadowMappedPaths` is different then the dimensions of the `ShadowMappedModules` will match that of `ShadowMappedPaths`. I thought keeping them independent also keeps this simpler to understand.

ygen/codegen.go
wenovus: I think the readability is OK, maybe I got used to it though. I do think it's an advantage when the compiler tells you a parameter is unused.

ygen/ir.go
robshakir: Sounds good to me. I was just wanting to check that we were on the same page here.

ygen/ir.go
robshakir: SGTM.

ygen/genir.go
robshakir: The cost here isn't particularly high as you say -- so I'm happy with just saying that this can stick around for some time.

ygen/directory.go
robshakir: SGTM.

ygen/directory.go
robshakir: SGTM.

ygen/directory.go
robshakir: Great - that sounds good to me too :-)

ygen/codegen.go
robshakir: ACK - understand that it is both or neither -- my suggestion was that if `UseConsistentNamesForProtoUnionEnums` is set, should we automatically set the other one? alternatively, should we error if they are inconsistently set?

ygen/codegen.go
robshakir: Sure, I'm good with that.

ygen/codegen.go
wenovus: I guess for Go generation `UseConsistentNamesForProtoUnionEnums` is always false, so that's why for Go they could be different. If I made them the same field in `IROptions`, then within `GenerateIR` the call to `findEnumSet` won't be able to separate them from each other, and this would then mean the proto-specific logic cannot be switched off. Now it's possible that somehow we can merge the two flags but I think it's really not much of a problem for them to exist side-by-side for < 1 year before ygot@v1 is released, at which time they would be deleted.

robshakir: Ah, I see. OK, separate is OK with me, despite it being a bit confusing. I'd rather just get towards v1 when we can deprecate this :-)

robshakir (approved):

PR #646, opened 2022-04-15
Rename goGenState to GoLangMapper.

wenovus:
This name is exported since we might want other users to be able to access it.

The genFakeRoot parameter is also removed since this property can be
checked through other means, as is done in
https://github.com/openconfig/ygot/tree/ygen-ir-langmapper


robshakir (approved):
LGTM, thanks for splitting this :-)

PR #649, opened 2022-04-20
Add ability to delete root node with ytypes.DeleteNode

DanG100:
If len(path.Elem) == 0, then the DeleteNode should clear it.


wenovus:
Could you comment on what use case you're intending to use this for? I'm thinking this has to do with the fakeroot but I think understanding the use case will make it easier to judge whether this is an appropriate solution.

wenovus:
Incoporated in https://github.com/openconfig/ygot/pull/904

PR #650, opened 2022-04-21
Update IR to Support Go Generation

wenovus:
Various updates to the IR for better supporting Go generation.


ygen/codegen.go
```diff
@@ -443,7 +443,9 @@ func (cg *YANGCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*
 	}
 
 	// Store the returned schematree and enumSet within the state for this code generation.
-	gogen := newGoLangMapper(mdef.schematree, enumSet)
+	gogen := NewGoLangMapper(cg.Config.GoOptions.GenerateSimpleUnions)
```
robshakir, line 446: Should we deprecate the more complex unions at this point?

ygen/directory.go
```diff
@@ -72,6 +72,10 @@ type YangListAttr struct {
 	// keyElems is a slice containing the pointers to yang.Entry structs that
 	// make up the list key.
 	KeyElems []*yang.Entry
+	// ListKeyYANGNames is the ordered list of YANG names specified in the
+	// YANG list per Section 7.8.2 of RFC6020. Rely on this
```
robshakir, line 76:
```suggestion
	// YANG list per Section 7.8.2 of RFC6020. ygot relies on this ordering
```

Is this the right suggestion? Alternatively, do we mean that the consumer of the IR can rely on it?

ygen/genir.go
```diff
@@ -72,7 +73,7 @@ type IROptions struct {
 // applied to the schema whilst generating the IR.
 //
 // GenerateIR returns the complete ygen intermediate representation.
-func GenerateIR(yangFiles, includePaths []string, newLangMapper NewLangMapperFn, opts IROptions) (*IR, error) {
+func GenerateIR(yangFiles, includePaths []string, langMapper LangMapper, opts IROptions) (*IR, error) {
```
robshakir, line 76: I guess the logic here is that you are using the LangMapper to store some state that is then used in its functionality? It seems like there is a risk here that you could end up with some state already being there that breaks the output -- should we have some `Reset()` function in the interface that allows us to remove all non-argument states so that callers don't risk this stale data?

ygen/goelements.go
```diff
@@ -538,6 +553,50 @@ func (s *GoLangMapper) goUnionSubTypes(subtype *yang.YangType, ctx *yang.Entry,
...
+	if len(defaultValues) == 0 && mtype.DefaultValue != nil {
+		defaultValues = []string{*mtype.DefaultValue}
+	}
+	for i, defVal := range defaultValues {
+		var err error
+		if defaultValues[i], _, err = gogen.yangDefaultValueToGo(defVal, resolveTypeArgs{yangType: field.Type, contextEntry: field}, len(mtype.UnionTypes) == 1, compressPaths, skipEnumDedup, shortenEnumLeafNames, useDefiningModuleForTypedefEnumNames, enumOrgPrefixesToTrim); err != nil {
+			return nil, err
+		}
+	}
+	// TODO(wenbli): In ygot v1, we should no longer
```
robshakir, line 571: This TODO makes me feel like we could deprecate this now -- but I'm open to discussing this again.

robshakir (approved):

ygen/codegen.go
wenovus: I'm open to doing that; although I'm wondering whether we should keep the backwards-incompatible changes all concentrated in a single release at v1 to minimize the number of times downstream users have to change?

ygen/genir.go
wenovus: I see why you created `NewLangMapperFn` now. I actually did this to avoid writing extra boilerplate code, but I think I'll change it back.

ygen/goelements.go
wenovus: As above I'm wondering if we should deprecate all at once at v1 release.

ygen/genir.go
wenovus: Actually I'm not so sure about adding it back again because if we pass in a function we're also saying we might instantiate two versions of them, which is also what we don't want the user to have to prepare for. I think the user just has to design their `LangMapper` in such a way that any deterministic ordering of calls always produces a deterministic set of naming -- i.e. we guarantee that the `LangMapper` functions are always called in a certain order. OR alternatively (which I prefer more) that we specifically say that the ordering in which methods are called is unstable and may change over time, such that the user should be prepared for this and design their langMapper in such a way such that it is robust to this, by, for example, handling naming uniquification in a later stage.

wenovus: [Added comments that LangMapper should be idempotent](https://github.com/openconfig/ygot/pull/650/commits/62eecf269f5fc76bd8a7daa56c1446cdd0f7f15c)

PR #651, opened 2022-04-21
Convert `GenerateGoCode` to use IR for Code Generation.

wenovus:


ygen/gogen.go
```diff
@@ -2019,85 +1981,94 @@ func generateGetListKey(buf *bytes.Buffer, s *Directory, nameMap map[string]*yan
...
 
 	var listType string
 	var multiListKey *generatedGoMultiKeyListStruct
 	var listKeys []goStructField
 	var listKeyStructName string
 
-	// Key name elements are ordered per Section 7.8.2 of RFC6020. Rely on this
-	// fact for determisitic ordering in output code and rendering.
-	keyElemNames := strings.Fields(listField.Key)
+	shortestPath := func(ss [][]string) [][]string {
```
robshakir, line 2004: Could we add a comment here (or at the call site) since I think it took me a second to realise that the shortest path must be the key's path because the key cannot be within a descendent container.

ygen/gogen_test.go
```diff
@@ -1366,11 +652,174 @@ func (*Tstruct) ΛBelongingModule() string {
 }
 `,
 		},
-		wantUncompressed: wantGoStructOut{
```
robshakir, line 1369: Did we decide that we should not have testing for uncompressed schemas here? I don't immediately see any tests that replaced this.

robshakir (changes requested):

ygen/gogen_test.go
wenovus: There is no such thing as compressed vs. uncompressed once we have the IR as far as `writeGoStruct` is concerned, that is handled all at IR generation time.

ygen/gogen.go
wenovus: Done.

ygen/gogen_test.go
robshakir: Makes sense :-D, thanks, apologies I missed this.

robshakir (approved):

PR #652, opened 2022-04-22
Remove log that is always generated for annotation fields.

robshakir:
```
 * (M) ytypes/container.go
  - Since schemas that are generated with annotations will always
    generate the log that says that unmarshal into them is unsupported
    the log message produced is extremely low value to a user. To this
    end - just remove it.
```

@wenovus -- I concluded that it was just higher-value to remove this entirely. Given that the user knows that they generated the code with annotations, then I'm not sure they ever will have value from this log.

An alternative would be to check that the annotation field was populated, and then produce this log, but I think this seems like we should just implement unmarshalling annotations. I can look at this as a future task.

### DIFF
--- a/ytypes/container.go
+++ b/ytypes/container.go
@@ -19,7 +19,6 @@ import (
 	"reflect"
 	"strings"
 
-	log "github.com/golang/glog"
 	"github.com/kylelemons/godebug/pretty"
 	"github.com/openconfig/goyang/pkg/yang"
 	"github.com/openconfig/ygot/util"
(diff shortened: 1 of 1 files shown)


wenovus (approved):
Makes sense, just outputting it unconditionally seems like more spam than value.

wenovus:
@robshakir How would we like to release ygot right now? ygot releases look to be blocked by https://github.com/openconfig/ygot/pull/648, should we wait until its resolution or revert #638?

robshakir:
I suggest that we rollback #638 -- I think that will unblock changes, whilst remaining backwards compatible.

Once we resolve the conversation in #648 (which today will contain #638) then we can choose the fate of that code based on that conclusion. 

WDYT?

wenovus:
SGTM.

PR #657, opened 2022-04-27
Generate Path API using IR.

wenovus:


wenovus:
@DanG100 just heads up that this change will impact ypathgen, once this PR is merged in ygot I'll be looking to bringing the change to what you're working on so it would be good to avoid any large changes to ypathgen where you're working.

I'll ask you to be a reviewer to this PR once it's rebased to master.

ygen/directory.go
```diff
@@ -89,19 +89,17 @@ func GetOrderedFieldNames(directory *Directory) []string {
 }
 
 // GoFieldNameMap returns a map containing the Go name for a field (key
-// is the field schema name). Camelcase and uniquification is done to ensure
-// compilation. Naming uniquification is done deterministically.
-func GoFieldNameMap(directory *Directory) map[string]string {
+// is the field schema name). Name Uniquification is done to ensure
```
robshakir, line 92: Perhaps I'm missing something here - `GoFieldNameMap` seems like it should return a map of a field to a Go field name -- if you don't create the camel case name here, then isn't this actually not creating a Go field name?

ygen/gogen.go
```diff
@@ -1281,9 +1281,9 @@ func writeGoHeader(yangFiles, includePaths []string, cfg GeneratorConfig, rootNa
 	return common.String(), oneoff.String(), nil
 }
 
-// isScalarField determines which fields should be converted to pointers when
+// IsScalarField determines which fields should be converted to pointers when
 // outputting structs; this is done to allow checks against nil.
-func isScalarField(field *NodeDetails) bool {
+func IsScalarField(field *NodeDetails) bool {
```
robshakir, line 1286: Should this be in `util` if it's public?

ypathgen/pathgen.go
```diff
@@ -631,47 +620,38 @@ func getNodeDataMap(directories map[string]*ygen.Directory, leafTypeMap map[stri
 				HasDefault:            false,
 				YANGTypeName:          "",
 				YANGPath:              "/",
-				GoPathPackageName:     goPackageName(dir.Entry, splitByModule, trimOCPackage, packageName, packageSuffix),
+				GoPathPackageName:     goPackageName(dir.RootModule, splitByModule, trimOCPackage, true, packageName, packageSuffix),
```
robshakir, line 623: I think I see why `RootModule` was defined here - since you want to do something based on the top-level parent of the node in the schema tree. I suggest that this is better done by looking at the path to avoid the confusing definition.

You can just do `strings.Split(path, "/")[1]` to find this - and this is significantly less confusing than having it contained in the IR IMHO, since today we're implying that this thing is meaningful in some manner for all different IR users, whereas it's really just nuance of how ypathgen has chosen to name packages (another approach could have been to split it based on the defining module, if we allow all possible splits to populate the IR, it is going to grow without bounds).

ygen/gogen.go
wenovus: This is operating on a `ygen` type though. It would later go into the `gogen` package.

ygen/directory.go
wenovus: It's now operating on `ParsedDirectory`, which has the candidate Go names already populated via GoLangMapper.
Here: `directory.Fields[fName].Name`

ypathgen/pathgen.go
wenovus: As I just commented in https://github.com/openconfig/ygot/pull/655#discussion_r859359783 I think we'll need this information somehow, so we can remove `Path` xor `RootModule`. I'm OK either way. I just realized we could call `SchemaPath` `DataPath` to distinguish it from `Path`. RFC7950 uses the term [here](https://datatracker.ietf.org/doc/html/rfc7950#section-4.2.7).

ygen/gogen.go
robshakir: SGTM.

ypathgen/pathgen.go
robshakir: But `DataPath` would imply that this is in the data tree - right? Such that we really would end up with things like `/interfaces/interface[name=eth0]/config/name` as the 'datatree' path for a leaf.

I wonder whether we should just create a namespace-qualified version of the path - such that we have `/openconfig-interfaces:interfaces/openconfig-interfaces:interface/...` and then this would include all of the information that one could possibly need here?

ygen/directory.go
robshakir: So here, what is the value of this map? All the examples seem to have the same values corresponding to one another. Can we eliminate it?

robshakir:
@wenovus I see some outstanding comments here, is there something that we should discuss about those ones, or are they resolved?

ygen/directory.go
wenovus: `ParsedDirectory.Fields` doc comment explains why a map is needed here. Does this answer your question?
https://github.com/openconfig/ygot/blob/6f722a0cce2a47949294afa0c3f23b080d51e501/ygen/ir.go#L209-L212

ypathgen/pathgen.go
wenovus: I'm fine with keeping the name as-is, although `SchemaPath` also doesn't seem right because it doesn't have the choice/case statements. I also see the value in creating a new `Path` type to contain both the defining module as well as the node name. For example it could be used to workaround this issue: https://docs.google.com/document/d/1EUzb_uXQWnDETFJkOm2Ad9fIJ_N4rytq-f3JwupWBu8/edit#heading=h.xdjxwhqgihsl

My proposal is to leave this be for now, figure out what we're going to do with `ParsedDirectory`'s path (a possibly backwards-incompatible change), and then create this new `Path` type.

I'm recording these TODOs in my own notes so I'm not losing track of them.

wenovus:
> @wenovus I see some outstanding comments here, is there something that we should discuss about those ones, or are they resolved?

Sorry I don't know why I didn't reply to these. PTAL I'm proposing no further action on this PR.

wenovus:
@robshakir When you have time PTAL -- I'd like to release this in order to prevent tech debt from accumulating in ygnmi. Thanks.

ygen/directory.go
robshakir: Ah, I see, thanks -- missed that there might be collisions here.

ypathgen/pathgen.go
robshakir: Sounds good to me.

robshakir (approved):
LGTM! 

Maybe we can file an issue with the collection of "things that we might want to clean up", and come back to these as discussion points? It might be good to have a real-time discussion on them.

PR #658, opened 2022-04-28
Rename `ResolvedPath` to `LeafrefTargetPath`

wenovus:


ygen/ir.go
```diff
@@ -349,9 +349,10 @@ type YANGNodeDetails struct {
 	Module string
 	// Path specifies the absolute YANG schema node path.
 	Path string
-	// ResolvedPath specifies the leafref-resolved absolute YANG schema
-	// node path.
-	ResolvedPath string
+	// LeafrefResolvedPath specifies the leafref-resolved absolute YANG
+	// schema node path. This is populated only if the YANG node was a
+	// leafref.
+	LeafrefResolvedPath string
```
robshakir, line 355: `LeafrefTargetPath`?

robshakir (changes requested):

ygen/ir.go
wenovus: Done, thanks -- I thought this was recursively resolved but it's actually only a singly-resolved path, I'll make this clear in the comment as well.

robshakir (approved):

PR #659, opened 2022-04-28
Improve ΛBelongingModule() comment

wenovus:


ygen/gogen.go
```diff
@@ -1033,8 +1033,8 @@ func (t *{{ .StructName }}) ΛEnumTypeMap() map[string][]reflect.Type { return 
 	// function that has a generated struct as receiver, and returns the
 	// name of the module in which namespace the generated struct belongs.
 	goBelongingModuleTemplate = mustMakeTemplate("belongingModuleMethod", `
-// ΛBelongingModule returns the name of the module in whose namespace
-// {{ .StructName }} belongs.
+// ΛBelongingModule returns the name of the module having the same XML
```
robshakir, line 1036: I wonder whether "returns the name of the module that defines the namespace of {{ .StructName }}" maybe? This wording reads to me like each node could define its own namespace, which I don't think is correct.

robshakir (changes requested):

ygen/gogen.go
wenovus: SGTM.

robshakir (approved):

PR #661, opened 2022-05-04
Add a new JSON unmarshal option that prefers the shadow path values.

wenovus:
It reuses the same `ytypes.PreferShadowPath` currently used as a `ytypes.SetNodeOpt`.

Testing:
* unit tests for leaf, leaf-list, container, list
* integration test in schema_tests/
  * Improved the integration test (bgp-example-*.json) so that the "ebgp" and "ibgp" containers now contain either "config" and "state" values (see [bgp-example-opstate-with-shadow.json](https://github.com/openconfig/ygot/compare/unmarshal-prefershadowpath?expand=1#diff-872186bfbb32435f77539e6b4ce1abea985dd74c52a09ae60e4c88e349134e99)), such that depending on the toggle value of `preferShadowPath` either one is unmarshalled. This is what makes [bgp-example-opstate-prefershadow.json](https://github.com/openconfig/ygot/compare/unmarshal-prefershadowpath?expand=1#diff-e6b6c244dfae7c365aaf2c4f7c1c01ae1d43dbeea30e881ca35c763dd46c54d7) different from [bgp-example-opstate.json](https://github.com/openconfig/ygot/compare/unmarshal-prefershadowpath?expand=1#diff-9bc38bff750fb321ac17e470e8b4f36f96d0fbca9966d38c2476e546c3b1db2b). The other difference between these two files is that these containers under IPV4_UNICAST store different values, which already existed prior to change.


wenovus:
@robshakir Thanks for reviewing #662, this PR is the prerequisite to it, pinging in case you missed it, thanks!

util/path.go
```diff
@@ -65,9 +65,37 @@ func ShadowSchemaPaths(f reflect.StructField) [][]string {
...
+// Paths are embedded in the "path" struct tag and can be either simple:
+//   e.g. "path:a"
+// or composite (if path compression is used) e.g.
+//   e.g. "path:config/a|a"
+// In the latter case, this function returns {"config", "a"}, because only the
+// longer path exists in the data tree and we want the schema for that node.
+// This case is found in OpenConfig leaf-ref cases where the key of a list is a
+// leafref; the schema *yang.Entry for the field is given by
+// schema.Dir["config"].Dir["a"].
+func RelativeSchemaPath2(f reflect.StructField, preferShadowPath bool) ([]string, error) {
```
robshakir, line 87: I don't think this naming makes sense, `RelativeSchemaPath2` doesn't really tell us why it differs.

I'd suggest either:

 * changing every use of `RelativeSchemaPath` to this new implementation.
 * using varargs such that you can hand `preferShadowPath`
 * or having `RelativeShadowSchemaPath` and `RelativeSchemaPath` as the public API, and have them both call `relativeSchemaPathInternal` that takes the `preferShadowPath` argument.

My vote would be for the third option.

util/reflect.go
```diff
@@ -458,9 +458,24 @@ func DeepEqualDerefPtrs(a, b interface{}) bool {
 // if the struct tag is invalid, or nil if tag is valid but the schema is not
 // found in the tree at the specified path.
 func ChildSchema(schema *yang.Entry, f reflect.StructField) (*yang.Entry, error) {
+	return ChildSchema2(schema, f, false)
+}
+
+// ChildSchema2 returns the schema for the struct field f, if f contains a valid
```
robshakir, line 464: Same issue with this function naming as above.

robshakir (changes requested):

util/path.go
wenovus: Done, I named it `RelativeSchemaPathPreferShadow` because it uses the "path" tag if the "shadow-path" tag doesn't exist.

util/reflect.go
wenovus: Done, named it `ChildSchemaPreferShadow`

robshakir (approved):
Thanks for the changes Wen!

PR #662, opened 2022-05-04
Allow JSON input for `ytypes.SetNode`

### LINKED ISSUE #660: Feature request: `SetNode` should accept JSON input for non-leaf nodes.
Today, we can call `Unmarshal` for a container to unmarshal JSON, but `SetNode` throws an error if given a `Set` for a non-leaf node.

Low priority - but we should fix this. Filing this bug for the backlog.

wenovus:
Both leafs, non-leafs, are supported with the `PreferShadowPath` option.

Fixes #660


wenovus:
@robshakir I prioritized this change since this (along with #661) will enable and make supporting `GetRequest` queries easier in Ondatra.

robshakir (approved):
Thanks for this change Wen, and especially adding the test coverage for both JSON and non-JSON.

Issue #663, opened 2022-05-05
`ygotreflect` package for easier access to GoStruct metadata and manipulations


wenovus:
With the recent rename of `Validate` to `ΛValidate`, the `ygot.ValidatedGoStruct` interface has become less usable, not to mention this being a breaking change in order to allow nodes named `validate` to be created. This breaking change was unavoidable due to the definition of `ygot.ValidatedGoStruct` having to be static.

In the future, we might very well add new "metadata" methods such as these to GoStructs, some of which will be controlled by generation flags. In order for these methods to be accessible, the current way of using `ValidatedGoStruct` has two problems:
1. It takes names away from the YANG namespace.
2. It cannot contain any method that's not generated unconditionally.

A more flexible way would be to write functions for accessing these metadata in a separate `gostruct` package that would not only avoid awkward names, but also allow metadata generated conditionally to be accessible (e.g. `gostruct.Validate(interface)`). This, however, comes at the expense of discoverability due to these being functions rather than methods.

A way to get all these benefits without sacrificing discoverability is to provide some sort of "ygotreflect" package for traversing GoStructs. An API could look something like this (inspired by https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect#Message):
```go
interface := &oc.Interfaces_Interface{}
interface.YgotReflect().Validate()
interface.YgotReflect().Name() // a string
interface.YgotReflect().Descriptor().BelongingModule // a string
interface.YgotReflect().Interface() // returns the ValidatedGoStruct
interface.YgotReflect().Type().New()
interface.YgotReflect().Type().Parent().New()
interface.YgotReflect().Range() // returns a set of ygotreflect.FieldDescriptor values.
```

Such a set of utilities would not only easily provide the functionalities to solve the problem above, but is also highly-extensible, while having a low threat to backwards-incompatibility.

I'm planning to gather some input and explore whether this API would help improve the ease of user development.

robshakir:
I think this is an interesting idea :-) The challenge of overlapping with potential YANG node names is always present for names that are user-friendly. Given that this is the case, one could observe that if I create a YANG model that has `ygot-reflect` as a name then this itself is going to overlap with fields.

I'm not quite sure that I see the full use case for a `reflect` like library -- what is the benefit to the user of having `Interface()` (especially chained onto a receiver that implements the interface itself)?

wenovus:
[...]

PR #664, opened 2022-05-07
Un-deprecate `ygot.GoStruct`, revert `ygot.ValidatedGoStruct` to pre-v0.18.1

wenovus:
This reverts the [backwards-incompatibility in v0.18.1](https://github.com/openconfig/ygot/releases/tag/v0.18.1) without reverting #643.

**This PR proposes in the near future to deprecate `ValidatedGoStruct` and instead use https://github.com/openconfig/ygot/issues/663 to access all functions currently in `ValidatedGoStruct`.**

The idea is to revert the `ValidatedGoStruct` definition in order to preserve backwards-compatibility and thus break the usage of `ValidatedGoStruct` only for those minority of users using a different name for Validate(). After #663 is rolled out, then these users will no longer be discriminated and a standard API will be used regardless of generation flags.

In order to allow these minority of users access to ygot/ytypes' helper functions, `ygot.GoStruct` has been un-deprecated and used throughout ygot, and a new internal interface containing `ΛValidate()` is used whenever necessary. There is a dearth of internal functions in ygot that actually requires calling these methods so this conversion was pretty hassle-free.

Un-deprecating `ygot.GoStruct` is necessary in order to allow those using a different name for `Validate()` to access ygot/ytypes helpers. This is because their generated structs do not satisfy the `ValidatedGoStruct` interface and hence it cannot be used as the input parameter type. `ygot.GoStruct` is the only other available interface.


robshakir:
This makes sense to me, thanks @wenovus. I think this keeps things cleaner, given that we know that we have ~0 users that are using a name other than `Validate`.

ygot/types.go
```diff
@@ -58,6 +56,17 @@ type ValidatedGoStruct interface {
 	ΛBelongingModule() string
 }
 
+// validatedGoStruct is an interface used for validating GoStructs.
+type validatedGoStruct interface {
```
robshakir, line 60: Essentially, this interface is internal-only such that it is *always* implemented by a generated struct, right?

robshakir (approved):

ygot/types.go
wenovus: Yes, `ΛValidate` and `IsYANGGoStruct()` are always generated.

robshakir: Great, thanks -- can we note this in the description?

wenovus:
```
 59 // validatedGoStruct is an interface used for validating GoStructs.                                                                                                                                                                                                                                                                                    
 60 // This interface is implemented by all Go structs (YANG container or lists),                                                                                                                                                                                                                                                                          
 61 // regardless of generation flag.                                                                                                                                                                                                                                                                                                                      
```
Done.

wenovus:
So this still is a backwards-incompatible change because `ytypes.UnmarshalFunc` and `ytypes.MergeStructs` now take in `ygot.GoStruct` instead of `ygot.ValidatedGoStruct` which may break existing code that expect a parameter or a return value of `ValidatedGoStruct`, although I believe these changes are what we want (since we want to deprecate this interface), and are tolerable enough.

https://github.com/openconfig/ygot/releases/tag/v0.20.0

robshakir:
I agree, this seems much more tolerable than needing to have users change their `.Validate()` calls -- I expect the number of places that folks are casting to `ValidatedGoStruct`.

PR #666, opened 2022-05-12
More accurate enum naming and classification according to namesake type.

wenovus:
Currently for unions it uses the outer-most union for the name and the type classification, but it should rather use the namesake union instead.

The integration test changes are due to **unused enums** now removed from the generated file because these were inline enums that should never have been generated in the global enum file.

### DIFF
--- a/testdata/modules/openconfig-complex.yang
+++ b/testdata/modules/openconfig-complex.yang
@@ -94,7 +94,9 @@ module openconfig-complex {
     }
 
     leaf leaf-default-override {
-      type cyclone-scales;
+      type union {
+        type cyclone-scales;
+      }
       default 3;
     }
   }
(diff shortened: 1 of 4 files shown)


robshakir (approved):
Should this be behind a flag, since it potentially changes generated protobuf union names in downstream code? (I'm not sure whether it actually does within OpenConfig, but I think with these kind of changes - and what we've seen recently, we need to give users a way to opt-out of default changes for at least a couple of releases.) Did you regenerate the schemas?

Change seems logical to me though -- thanks.

robshakir (changes requested):
Sorry, blocking comment on the flag above.

wenovus:
> Should this be behind a flag, since it potentially changes generated protobuf union names in downstream code? (I'm not sure whether it actually does within OpenConfig, but I think with these kind of changes - and what we've seen recently, we need to give users a way to opt-out of default changes for at least a couple of releases.) Did you regenerate the schemas?
> 
> Change seems logical to me though -- thanks.

I've put this PR after the backwards-incompatible enum flag removal PR such that it is clear there is no material changes in any output code.

robshakir (approved):
Approving based on the discussion in #667.

Issue #412, opened 2020-07-08
Restructure Code Generation in ygen


robshakir:
`ygen` started out as a library that was solely handling generation of Go code, subsequently it has transitioned into being one that generates three different outputs:

* Directories and types - which is to say that it generates an abstract object that maps to a Go `struct` or a Protobuf `message`.
* Go code
* Protobuf code

This means that we have a large multi-purpose package that is becoming complex to maintain. Since the only consumers of the public APIs are within our generator binaries, then we have a bit of scope to be able to shuffle this package around - and this should definitely be done prior to a 1.0 release of `ygot`.

To this end, I suggest that we restructure this such that we have:

* `ygen` that performs the base mapping of an input YANG set of files into an abstract representation.
* `ygot/gogen` - a specific package that generates Go code, this should call `ygen` public APIs to be able to do all the pre-processing and then have the logic to map that into Go.
* `ygot/protogen` - a specific package that generates Proto code, similar to `ygot/gogen` above.

This refactoring should likely be done in a few stages:

* [ ] Mechanical move of Go code generation into `gogen`.
* [ ] Mechanical move of Proto code generation into `protogen`
* [ ] Refactoring the generation of `Directories` to be disconnected from the generation of Go (today, there is a linkage to the Go generation since we do not abstract the state storage interface).
* [ ] Refactoring `ygot/gogen` to use the abstract Directories as input.
* [ ] Refactoring `ygot/protogen` to use the abstract Directories as input.

This is a fairly major restructuring, so we'll likely go through some changes to APIs that are incompatible with each other to start with, so we should be careful about figuring out when to cut the next version of ygot - and make sure that this is an RC for 1.0.

robshakir:
Looking at this, I think this is actually better to restructure the other way around than I suggested initially. So - we'd internally restructure the code to be able to use the same pattern that `ypathgen` uses - such that we use the pattern shown below:

![ygot](https://user-images.githubusercontent.com/4836821/86983428-b65f7b80-c140-11ea-9241-f9507f9802a4.png)

This will mean that the first set of changes restructure `GenerateGoCode` and `GenerateProtoCode` internally to be able to rely on this pattern - and then once this is done, we then pull out the code that is now language specific into the packages that I suggested above.

[...]

PR #667, opened 2022-05-14
[backwards-incompatible] Delete enum flags for `proto_generator`

wenovus:
These flags make IR generation more complex (maybe a day or so of work), so I want to remove them now a bit early ahead of v1 release to avoid the throwaway work.


robshakir:
So, what is the transition plan here with #666 too. Are we saying that we are going to make the IR change something that gates 1.0?

wenovus:
TLDR; I think the IR change will gate 1.0. I'm not sure how #666 is relevant since that to me is a bugfix.

Theoretically we don't need to wait for the IR changes to be complete, but in practice this work is being done prior to some of the other 1.0 blocking work so I think it ends up gating it. Perhaps we really should've cut 1.0 first, but the IR changes are significant, and there was a thinking to make it gate 1.0: https://github.com/openconfig/ygot/issues/412#issue-653545770

robshakir:
I think what I am saying is that if I am someone that is currently generating protobufs with ygot, and we make a change like in #666, or the default change that we've made here, then do I get locked into a situation where I cannot actually add fields to my protobufs without making a backwards-incompatible change to other fields?

If that's the case, then I think I need to pin at a version of ygot to keep using, which is fine -- but it's not that clear if I pin to 0.20.1 vs. 0.20.2 right? I was therefore thinking is there some way that we're having some "big bang" release that says to users "we have fixed a bunch of bugs and you can expect fewer/no backwards incompatible changes for some time"? To me, that is why it is important to understand when v1.0 is, and whether this is coupled to the IR, because theoretically, we could cut 1.0.0 right now, and then need to move to 2.0.0 with the IR -- and with no flags to control these changes any more, then we may get stuck in a world where folks are asking for changes in pre-IR code to fix other bugs.

wenovus:
So for any backwards-incompatible change I will increment the minor version (so @v0.21.0), if I state very clear in the release notes that this is backwards-incompatible for protobuf generation do you think it's hard to know which one to pin to?

I agree that a "big bang" release is preferred because it causes the fewest churn for users. I could've make the proto conversion (I will post the PRs soon) backwards-compatible, but the complexity just didn't seem worth it to me for a throw-away change (I'm estimating ygot@v1 will have to wait until EOY unless we prioritize it).

That being said, one possible "hack" here is to make the proto's backwards-incompatible changes happen right now, such that when @v1 occurs, the proto code is still backwards compatible. Almost all of the @v1 incompatible items deal with Go. What do you think?

robshakir:
I'm good with pinning to `v0.20.0`. If this is the only proto-backwards-incompatible change then this seems OK with me as a strategy, are there others that you're seeing?

wenovus:
> I'm good with pinning to `v0.20.0`. If this is the only proto-backwards-incompatible change then this seems OK with me as a strategy, are there others that you're seeing?

The only other one is `generate_fake_root`. We discussed last year that we'd delete this one for both Go and proto for simplicity and always generate the fake root message/struct. I just looked and I believe this is backwards-compatible except for the flag deletion.

robshakir:
Sounds great. This is good to go from my perspective.

PR #668, opened 2022-05-17
Rename protoGenState to ProtoLangMapper

wenovus:


robshakir (approved):
Thanks for making this an easy mechanical change to review!

PR #669, opened 2022-05-17
Change `IR.Enums` map's key to be a unique key instead of the enum's name.

wenovus:
Do this using the unique key created by `findEnumSet`. Otherwise there could be a name conflict (e.g. proto enum names) when looking up an enum from the `IR.Enums` map.

If there is a conflict, then an error will be raised; however, this will never happen because even though the key space is currently separate within `enumSet`:
https://github.com/openconfig/ygot/blob/6f722a0cce2a47949294afa0c3f23b080d51e501/ygen/enumgen.go#L72-L75
Due to the way each key type is constructed, it means the three types' keys will never conflict.


ygen/enumgen.go
```diff
@@ -172,30 +174,31 @@ func (s *enumSet) enumeratedUnionEntry(e *yang.Entry, compressPaths, noUnderscor
...
 // determining the name from an identityref leaf, rather than directly from the
 // identity.
-func (s *enumSet) identityrefBaseTypeFromLeaf(idr *yang.Entry) (string, error) {
+func (s *enumSet) identityrefBaseTypeFromLeaf(idr *yang.Entry) (string, string, error) {
 	return s.identityrefBaseTypeFromIdentity(idr.Type.IdentityBase)
 }
 
 // identityrefBaseTypeFromIdentity retrieves the generated type name of the
 // input *yang.Identity. The value returned is based on the defining module
 // followed by the CamelCase-ified version of the identity's name.
```
robshakir, line 183: Update the doc comment here?

robshakir (changes requested):

ygen/enumgen.go
```diff
@@ -172,30 +174,31 @@ func (s *enumSet) enumeratedUnionEntry(e *yang.Entry, compressPaths, noUnderscor
 // function since it covers the common case that the caller is interested in
 // determining the name from an identityref leaf, rather than directly from the
 // identity.
-func (s *enumSet) identityrefBaseTypeFromLeaf(idr *yang.Entry) (string, error) {
+func (s *enumSet) identityrefBaseTypeFromLeaf(idr *yang.Entry) (string, string, error) {
```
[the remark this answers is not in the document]
wenovus: Done.

ygen/enumgen.go
wenovus: Done.

wenovus:
@robshakir Pending comments are resolved.

robshakir (approved):
Thanks for the changes.

PR #670, opened 2022-05-18
Add `MappedType.UnionTypeInfos`

wenovus:
This will be used to store the key to the global enum map in the `IR` for looking up union enums in `IR.Enums`.

The next PR actually populates `EnumeratedYANGTypeKey`.


ygen/genstate.go
```diff
@@ -67,6 +70,16 @@ type MappedType struct {
...
 }
 
+// MappedUnionSubtype stores information associated with a union subtype within
+// a MappedType.
+type MappedUnionSubtype struct {
+	// EnumeratedYANGTypeKey stores a globally-unique key that can be
+	// used to key into IR's EnumeratedYANGTypes map containing all of the
+	// enumeration definitions. This value should only be populated when
+	// the union subtype is an enumerated type
+	//EnumeratedYANGTypeKey string
```
robshakir, line 80: Is this commented out to avoid an unused warning?

ygen/genstate.go
```diff
@@ -67,6 +70,16 @@ type MappedType struct {
 	DefaultValue *string
 }
 
+// MappedUnionSubtype stores information associated with a union subtype within
+// a MappedType.
+type MappedUnionSubtype struct {
+	// EnumeratedYANGTypeKey stores a globally-unique key that can be
+	// used to key into IR's EnumeratedYANGTypes map containing all of the
+	// enumeration definitions. This value should only be populated when
+	// the union subtype is an enumerated type
```
robshakir, line 79:
```suggestion
	// the union subtype is an enumerated type.
```

ygen/genstate.go
```diff
@@ -52,6 +52,9 @@ type MappedType struct {
 	// the generated code from the structs maintains only type validation,
 	// this is not currently a limitation.
 	UnionTypes map[string]int
+	// UnionTypeInfos stores more information about each union subtype that
+	// can be populated if needed during the code generation pass.
+	UnionTypeInfos map[string]MappedUnionSubtype
```
robshakir, line 57:
```suggestion
	UnionTypeInfo map[string]MappedUnionSubtype
```

nitty naming suggestion.

ygen/genstate.go
```diff
@@ -52,6 +52,9 @@ type MappedType struct {
 	// the generated code from the structs maintains only type validation,
 	// this is not currently a limitation.
 	UnionTypes map[string]int
```
robshakir, line 54: Should we refactor here to have `UnionTypes` be `map[string]MappedUnionSubType` and then make the `int` that is the value here be within that value? We could do it the other way around by having the `UnionTypeInfo` just store the int now and then we can remove this in a subsequent PR?

robshakir (changes requested):

ygen/genstate.go
wenovus: This is commented out since it's not being used so I wanted to make this clear. This is uncommented in the next PR: https://github.com/openconfig/ygot/pull/671

ygen/genstate.go
wenovus: I like having the plural because maps contain multiple units of `UnionTypeInfo`. I know 'info' is an uncountable noun, but I don't like English's exceptions when being used in a programming environment which is more about rules rather than exceptions. What do you think :)?

ygen/genstate.go
wenovus: Sounds good, do you mind if I do
1) Add the index to `UnionTypeInfo`
2) Remove `UnionTypes`
both in a later PR after the chain?

ygen/genstate.go
wenovus: So actually once `UnionTypes` is removed, I'll rename this to this name. How about that?

wenovus:
@robshakir PTAL

ygen/genstate.go
robshakir: That sounds reasonable to me.

robshakir (approved):

ygen/genstate.go
wenovus: @robshakir  Done, I put it at the end of the chain here: https://github.com/openconfig/ygot/pull/684

PR #680, opened 2022-05-19
Convert `TestWriteProtoMsg` to use IR

wenovus:


robshakir (approved):
🥳 🎉 

Awesome work on this chain of PRs @wenovus -- this was a massive in-place refactor for both of the Go and Proto libraries. This chain was also super-easy to review thanks to the care you'd taken to break it out! Great work!

wenovus:
> partying_face tada
> 
> Awesome work on this chain of PRs @wenovus -- this was a massive in-place refactor for both of the Go and Proto libraries. This chain was also super-easy to review thanks to the care you'd taken to break it out! Great work!

Thanks Rob! Np! I learned that the *work* to refactor a single complex layer into two layers involved having to understand this complex single layer (obviously!), and another observation is that from now on, writers of generators would have to reason about both the `IR`, as well as the goyang AST when they need something which the `IR` does not provide, instead of having to reason only about the AST. That being said, this will occur only infrequently so users wouldn't need to understand the AST as much, and for more complex processing of the AST (e.g. `LeafrefTargetPath`), the two layers create a separation of concerns which insulates the writer of the generator from having to know how to do this complex processing: all they need to do is to read the `IR` doc comments. On the test side, it simplified the code generation unit tests by requiring the simpler `IR` objects instead of AST objects as input, and arguably increased robustness by creating a middle test layer where mistakes can be found.

A problem that remains though is that the `IR` objects in test files are still difficult to create and maintain just because there is so much information, and using YANG files as input to tests remain the easiest way to ensure rigorous testing while doing our best to limit test complexity. Daniel found this as well while working on ygnmi (https://github.com/openconfig/ygnmi/pull/19), and perhaps there is no easy way around that.

In short, it was fun to see how the refactoring affected the codebase. I'll submit an issue with a list of the things I'd like to clean up or discuss. Oh yes, I didn't expect this to happen either, but next time I need to budget time for clean-up items.

robshakir:
> > partying_face tada
> > Awesome work on this chain of PRs @wenovus -- this was a massive in-place refactor for both of the Go and Proto libraries. This chain was also super-easy to review thanks to the care you'd taken to break it out! Great work!
> 
> Thanks Rob! Np! I learned that the _work_ to refactor a single complex layer into two layers involved having to understand this complex single layer (obviously!), and another observation is that from now on, writers of generators would have to reason about both the `IR`, as well as the goyang AST when they need something which the `IR` does not provide, instead of having to reason only about the AST. 

This is a good observation - but perhaps we can also think about how this separates concerns. If someone wants to add a new language/output format, they go away and look at the (much better documented, more compact, simpler to understand) IR and don't really need to know that much about how goyang works with YANG. This means that (hopefully) in the majority of cases the user outputting the code can do so without coming to the ygot-maintainers team. It's only when they find "hey, I want to have some other data that is missing" that they would necessarily need to come have that discussion.

> That being said, this will occur only infrequently so users wouldn't need to understand the AST as much, and for more complex processing of the AST (e.g. `LeafrefTargetPath`), the two layers create a separation of concerns which insulates the writer of the generator from having to know how to do this complex processing: all they need to do is to read the `IR` doc comments.

Yep!

> On the test side, it simplified the code generation unit tests by requiring the simpler `IR` objects instead of AST objects as input, and arguably increased robustness by creating a middle test layer where mistakes can be found.
> 
> A problem that remains though is that the `IR` objects in test files are still difficult to create and maintain just because there is so much information, and using YANG files as input to tests remain the easiest way to ensure rigorous testing while doing our best to limit test complexity. Daniel found this as well while working on ygnmi ([openconfig/ygnmi#19](https://github.com/openconfig/ygnmi/pull/19)), and perhaps there is no easy way around that.

This is interesting -- I think there is some limitation here in terms of the inherent complexity of YANG -- the IR itself does have some complexity, but that complexity is rather driven by YANG itself. There is a lower-limit as to how little information we can put there if we need to include each of the vagaries of YANG. We could discuss this going forward though -- is something like an "IR helper" the right thing to do? It seems like probably not -- since whatever we're doing here we likely need to have some exhaustive set of the "things YANG can do" inputs, which may just be obscured if we don't write them in YANG!
 
> In short, it was fun to see how the refactoring affected the codebase. I'll submit an issue with a list of the things I'd like to clean up or discuss. Oh yes, I didn't expect this to happen either, but next time I need to budget time for clean-up items.

Thanks again for all the work -- yep, budgeting a bit of time for cleanup is always good, since it means we can take a "wait and see whether this was the right decision" point of view during some code reviews rather than blocking.

Let me check the chain of PRs to make sure I'm up to date and hopefully you can merge them before the weekend :-D

PR #681, opened 2022-05-23
Add `ConfigFalse` field to IR's `ParsedDirectories`.

wenovus:
This value corresponds to
https://datatracker.ietf.org/doc/html/rfc7950#section-7.21.1

This is needed in order to finish refactoring ygnmi to use IR, which needs this value to know whether to allow a user to multiplex a path using a `dut.Telemetry().Interfaces("foo").Config()` and `dut.Telemetry().Interfaces("foo").State()`.

The reason I chose `ConfigFalse` instead of just `Config` is because "config true" is the default value and is much more common in the OpenConfig models.

### DIFF
--- a/ygen/directory.go
+++ b/ygen/directory.go
@@ -186,6 +186,7 @@ func getOrderedDirDetails(langMapper LangMapper, directory map[string]*Directory
 			BelongingModule:   belongingModule,
 			DefiningModule:    definingModuleName,
 			RootElementModule: rootModule,
+			ConfigFalse:       !util.IsConfig(dir.Entry),
 		}
 		switch {
 		case dir.Entry.IsList():
--- a/ygen/genir_test.go
+++ b/ygen/genir_test.go
@@ -1002,6 +1002,7 @@ func TestGenerateIR(t *testing.T) {
 					BelongingModule:   "openconfig-simple",
 					RootElementModule: "openconfig-simple",
 					DefiningModule:    "openconfig-simple",
+					ConfigFalse:       true,
 				},
 				"/openconfig-simple/remote-container": {
 					Name: "OpenconfigSimple_RemoteContainer",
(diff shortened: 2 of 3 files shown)


robshakir (commented):
I wonder whether we should call this `ReadOnly` rather than `ConfigFalse`? It would make the IR a little more accessible without having to understand `config` in YANG.

That being said, there's obviously some YANG that is already needing to be understood, but it'd be good to think about this.

wenovus:
> I wonder whether we should call this `ReadOnly` rather than `ConfigFalse`? It would make the IR a little more accessible without having to understand `config` in YANG.
> 
> That being said, there's obviously some YANG that is already needing to be understood, but it'd be good to think about this.

@robshakir So I'm using `ConfigFalse` because of this comment you had here: https://github.com/openconfig/ygot/pull/600#discussion_r744183015 where we decided to keep to the RFC's definition. It makes sense to me that we don't yet create another term into already the complex YANG language. Personally I also think forcing the user to understand what "config" is is the right thing to do, since it just involves a bit of reading and provides the user with a definitive definition. I've now added a link to the YANG RFC for this.

wenovus:
@robshakir So I'm looking to merge this one before the pending chain because the chain starts with a backwards-incompatible change that I'd like to be careful about when to release.

robshakir (approved):
Thanks for pointing out that we'd made the decision to stick with RFC-ish language - this makes sense to me, insofar that we're saying "know YANG, but don't necessarily know goyang" :-)

LGTM.

wenovus:
> Thanks for pointing out that we'd made the decision to stick with RFC-ish language - this makes sense to me, insofar that we're saying "know YANG, but don't necessarily know goyang" :-)
> 
> LGTM.

Makes sense to me too, goyang is IMO even harder to understand haha.

PR #682, opened 2022-05-23
allow an arbitrary prefix to be trimmed from package names

greg-dennis:


generator/generator.go
```diff
@@ -103,7 +103,7 @@ var (
 	listBuilderKeyThreshold = flag.Uint("list_builder_key_threshold", 0, "The threshold equal or over which the path structs' builder API is used for key population. 0 means infinity. This flag is only meaningful when wildcard paths are generated.")
 	pathStructSuffix        = flag.String("path_struct_suffix", "Path", "The suffix string appended to each generated path struct in order to differentiate their names from their corresponding schema struct names.")
 	splitByModule           = flag.Bool("split_pathstructs_by_module", false, "Whether to split path struct generation by module.")
-	trimOCPackage           = flag.Bool("trim_path_package_oc_prefix", false, "Whether to trim openconfig- from generated package names, when split_pathstructs_by_module=true.")
+	trimPackagePrefix       = flag.String("trim_package_prefix", "", "Prefix to trim from generated package names, when split_pathstructs_by_module=true.")
```
wenovus, line 106: So we're not creating a general flag alongside the OC flag? This is a backwards-incompatible change. I think I'm still in favour of adding the general flag alongside the OC-specific flag while telling the user that they should be aware of collisions between modules from different org names.

@robshakir you mentioned this concern originally here: https://github.com/openconfig/ygot/pull/453#issuecomment-700310248, would be good to get your comments too.

robshakir: I was concerned that having these be generally available would cause more overlaps in generated code names where there is not external checks that this is not the case (e.g., where we have vendor devices that support >1 different model, such as their own implementations and OpenConfig).

@greg-dennis Is there a specific set of modules that you're aiming to generate here that this might be safe for?

greg-dennis: My current use case is for Open Traffic Generator, where the package names are currently appearing as "opentrafficgeneratorbgp", "opentrafficgeneratordiscovery", "opentrafficgeneratorflow",  "opentrafficgeneratorisis",
 "opentrafficgeneratorport". That seemed kind of ugly to me. The yang models are here for reference:
 https://github.com/open-traffic-generator/models-yang/tree/main/models
 
I was thinking the burden was on the user/application to choose the codegen parameters that work best for their use case. Presumably they would choose the strip the prefix from their "primary" or most common namespace. Perhaps an alternative would be to name packages by their module prefix, as specific in the yang module, or for users to specify their own prefix replacement (e.g. "otg" in my case). Or as I said, just put the user in control? What are you comfortable with?

wenovus: I like the idea to add a feature to replace a custom prefix (e.g. "open-traffic-generator" with "otg"). Per the [RFC section](https://datatracker.ietf.org/doc/html/rfc7950#section-5.1),
```
   Developers of YANG modules and submodules are RECOMMENDED to choose
   names for their modules that will have a low probability of colliding
   with standard or other enterprise modules, e.g., by using the
   enterprise or organization name as a prefix for the module name.
   Within a server, all module names MUST be unique.
```
So for the organizations that follow this standard (e.g. openconfig, ietf, open-traffic-generator), this can be used to reduce verbosity. Obviously there is still a possible collision but it's much better than trimming, and there is the added benefit now that the same mechanism can actually help users avoid collisions as well.

The actual implementation will come in name pairs, if there is a Go flag mechanism for that this sounds like a good option to me.

robshakir: I'm OK with giving the user the rope to choose how to do this with flags. The "good" thing about a mistake here is that it breaks at compile time.

@wenovus -- you're imagining something like `-sub_module_prefix="openconfig," -sub_module_prefix="open-traffic-generator,otg"`?

wenovus: Sorry I missed that the proposal simply asks to trim the path struct package names, and not including the generated Go names. Given that we're only allowing trimming a single org's name, it should be the primary org being used in the codebase, and so IMO there should be a very low chance for collisions. So I'm ok with this change.

wenovus: I still like the prefix solution better, since it gives users more flexibility in case someone doesn't follow this naming convention. It still seems to me the two flags should be treated in a similar way even though I agree that the chances for collision is even lower since it requires the conflicting modules to both be instantiating nodes. Given we're aiming for version 1 release, I want to be careful about the flags we do introduce. @greg-dennis is there a reason you prefer trimming rather than replacing? @robshakir your thoughts are also welcome.

greg-dennis: I don't have a strong preference. So long as it shortens the opentrafficgenerator package names, I can live with whatever. Offhand, it seemed to me like trimming a prefix was a small, uncomplicated extension of what was already provided for openconfig, and something low risk that would only be used by the code generation for the primary namespace of its modules. A map of replacements is heavier weight, and that may be justified. I don't have a use case where to demonstrate that it's worth its weight, but if you and Rob anticipate those cases, then that's fine by me.

wenovus: Sounds good to me, let's just keep the trimming implementation for now.

wenovus, line 106: nit: "module prefix to trim from generated package names (e.g. openconfig-)...

wenovus (approved):

PR #683, opened 2022-05-23
Add `ShadowSchemaPath` field to IR's `YANGDetails`.

wenovus:
This is needed in order to finish refactoring ygnmi to use IR, which needs this value when allowing a user to multiplex a path using a `dut.Telemetry().Interfaces("foo").Config()` and `dut.Telemetry().Interfaces("foo").State()`.


ygen/ir.go
```diff
@@ -438,6 +438,10 @@ type YANGNodeDetails struct {
 	// SchemaPath specifies the absolute YANG schema node path. It does not
 	// include the module name nor choice/case elements in the YANG file.
 	SchemaPath string
+	// ShadowSchemaPath specifies the absolute YANG schema node path of the
+	// shadowed sibling node. It does not include the module name nor
+	// choice/case elements in the YANG file.
```
robshakir, line 443: I suggest that we include a definition here of what a shadow schema path is -- we can probably just say something like.


```
A shadow schema path is included when a leaf exists in both 'intended' and 'applied'
state of an OpenConfig schema (see https://datatracker.ietf.org/doc/html/draft-openconfig-netmod-opstate-01) and hence is within the 'config' and 'state' containers
of the schema. ShadowSchemaPath is populated only when the -compress generator flag
is used, and indicates the path of the node not represented in the generated IR based
on the preference to prefer intended or applied leaves.
```

wenovus: Thanks, incorporated.

wenovus: @robshakir So similar to the parent PR I'm looking to merge this one before the pending chain because the chain starts with a backwards-incompatible change that I'd like to be careful about when to release.

robshakir (approved):

ygen/ir.go
robshakir: No problems on the ordering - was just giving you heads-up on my availability :-)

PR #684, opened 2022-05-24
[ygen-backwards-incompatible] Merge `UnionTypeInfo` into `UnionTypes`.

wenovus:
For background see https://github.com/openconfig/ygot/pull/670#discussion_r880044988

The backwards incompatibility comes from the new definition for `MappedType`.


wenovus:
@robshakir PTAL This PR addresses this comment at the beginning of the chain: https://github.com/openconfig/ygot/pull/670#discussion_r880626787

robshakir (commented):
Looks like this PR got some additional changes pushed to it -- I'll review for the contents and then come back and look to make sure that it's OK with the stack of PRs after that.

ygen/protoelements.go
```diff
@@ -327,6 +395,11 @@ func (s *ProtoLangMapper) yangTypeToProtoScalarType(args resolveTypeArgs, pargs
 	}
 }
 
+type unionSubtypeInfo struct {
```
robshakir, line 398: Can we add a comment here or move the type definition into the function itself since it seems to only be used in one place.

robshakir (changes requested):

ygen/protoelements.go
wenovus: The diff was bad. Please take a look at the diff now.

wenovus:
> Looks like this PR got some additional changes pushed to it -- I'll review for the contents and then come back and look to make sure that it's OK with the stack of PRs after that.

I forgot about this PR. It's merged with master now PTAL. This is the motivating comment thread for this PR: https://github.com/openconfig/ygot/pull/670#discussion_r880626787

robshakir (approved):
Ah, thanks for fixing the diff. LGTM.

PR #685, opened 2022-05-25
Add MergeOpt `MergeEmptyMaps` to allow an empty map to be merged.

wenovus:
Some users may want to use this as a different meaning.

Added the following note in the doc comments:
```
// NOTE: Since YANG doesn't distinguish between a nil map and an empty map,
// please consider another approach before using this option.
```


ejbrever (approved):
Thanks!

PR #689, opened 2022-06-06
Add `LangMapperExt` and `Flags` fields to each level of the `IR`.

wenovus:
The idea is to to call a different method from the interface for each level of hierarchy of the IR to populate the `Flag` fields for that level of the IR. So within `getOrderedDirDetails` for each `ParsedDirectory` and `NodeDetails`, and within the loop of genEnums (in `GenerateIR`) for each `EnumeratedYANGType`. This way the caller of `GenerateIR` is able to produce extra information for any level of the `IR`.

Tests added in genir_test.go


ygen/ir.go
```diff
@@ -89,6 +89,45 @@ type LangMapper interface {
...
+
+// LangMapperExt contains extensions that the LangMapper instance should
+// implement if extra information from the IR is required. These flag values
+// are expected to contain information rarely used but needed from goyang's
+// AST. Values that are expected to be used more often should be placed in the
+// IR itself so that other users can get access to the same information without
+// implementing it themselves.
+type LangMapperExt interface {
+	// PopulateFieldFlags populates extra information given a particular
+	// field of a ParsedDirectory and the corresponding AST node.
```
robshakir, line 109: Is this function called for all fields whether they are a leaf, leaf-list, list or container? What about for choice/case? It'd be good to just ensure that we have a clear expectation here.

ygen/ir.go
```diff
@@ -364,10 +403,11 @@ type NodeDetails struct {
 	// Shadow paths are paths that have sibling config/state values
 	// that have been compressed out due to path compression.
 	ShadowMappedPathModules [][]string
-	// TODO(wenbli): Think about how to add this in an easily-usable way.
+	// LangMapper during IR generation to assist the code generation stage.
```
robshakir, line 406: Not sure this makes sense -- the map[string]string needs to implement the interface? Could you clarify the comment?

robshakir (changes requested):

ygen/ir.go
wenovus: Fixed comment.

ygen/ir.go
wenovus: Fixed comment.

ygen/ir.go
robshakir: Thanks.

ygen/ir.go
robshakir: Thanks, took me some time to find out how to see that in the GitHub UI, but LGTM.

robshakir (approved):

PR #691, opened 2022-06-06
Remove dependency of enumgen on MappedType

wenovus:
This is a nit change. I think this was originally a convenience change but I think it makes sense for it to not depend on it since right now we're modifying it after receiving it.


ygen/enumgen.go
```diff
@@ -204,40 +204,38 @@ func (s *enumSet) enumName(e *yang.Entry, compressPaths, noUnderscores, skipDedu
 // a typedef which is either an identityref or an enumeration). The resolved
 // name is prefixed with the prefix supplied. If the type that was supplied
 // within the resolveTypeArgs struct is not a type definition which includes an
-// enumerated type, the MappedType returned is nil, otherwise it should be
-// populated.
-func (s *enumSet) enumeratedTypedefTypeName(args resolveTypeArgs, prefix string, noUnderscores, useDefiningModuleForTypedefEnumNames bool) (*MappedType, error) {
+// enumerated type, an empty string is returned as the first returned value;
```
robshakir, line 207: This use of `""` feels a bit of an odd overload, could we consider having a `bool` returned that lets us know whether this was found to be valid that could be checked explicitly? The `""` approach seems like its open to folks just not checking this return value's contents and then finding out that they generated invalid code later on.

robshakir (changes requested):

ygen/enumgen.go
wenovus: Thanks, changed to a bool.

robshakir (approved):

PR #692, opened 2022-06-06
Move enumset and schematree into `LangMapperBase`.

wenovus:
Currently, the individual LangMapper implementation instances (GoLangMapper and ProtoLangMapper) are required to hold these types, which can be confusing to the user since these are not fundamentally part of the LangMapper API -- rather it's their methods (i.e. enum name look-up and leafref resolution) that are part of the LangMapper API interface.

Embedding a special interface (`LangMapperBase`) not only decouples the implementation from the interface, but also provides a clear forward compatibility contract of methods (incl. enum name look-up and leafref resolution) available to implementors should more functionality become available in the future.

The proposed vision is:
* `LangMapper`-defined methods need to be implemented by LangMapper implementors.
* `LangMapperBaseSetup`-defined methods are set-up methods that are called either within `GenerateIR`, or by the unit tests of LangMapper methods.
* `LangMapperBase`-defined methods NOT part of the overall `LangMapper` interface are helper methods available to `LangMapper` implementors in their method implementations.

I will modify the design doc once all is agreed-upon and merged.

### DIFF
--- a/ygen/goelements.go
+++ b/ygen/goelements.go
@@ -133,12 +133,10 @@ var _ LangMapper = &GoLangMapper{}
 // GoLangMapper contains the functionality and state for generating Go names for
 // the generated code.
 type GoLangMapper struct {
-	// enumSet contains the generated enum names which can be queried.
-	enumSet *enumSet
-
-	// schematree is a copy of the YANG schema tree, containing only leaf
-	// entries, such that schema paths can be referenced.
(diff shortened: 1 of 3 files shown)


robshakir:
I think this makes sense, it certainly makes the per-language code look a lot neater, and not access internals that the IR should have populated itself, so SGTM.

PR #693, opened 2022-06-06
Add `InjectEnumSet` and `InjectSchemaTree` to `LangMapperBase`.

wenovus:
This allows unit tests to be created by custom LangMappers on their methods (e.g. `FieldName`, `KeyLeafType`).


ygen/ir.go
```diff
@@ -98,12 +98,24 @@ type LangMapperBaseSetup interface {
...
+	setEnumSet(*enumSet)
 
-	// SetSchemaTree is used to supply a copy of the YANG schema tree to
+	// setSchemaTree is used to supply a copy of the YANG schema tree to
 	// the mapped such that leaves of type leafref can be resolved to
 	// their target leaves.
-	SetSchemaTree(*schemaTree)
+	setSchemaTree(*schemaTree)
+
+	// SetupEnumSet is intended to be called by unit tests in order to set up the
```
robshakir, line 108: Should we call this `InjectEnumSet` or something similar to show that it is being used specifically for that purpose? `Setup...` seems to indicate to me that this is something that would be required for non-testing purposes.

ygen/ir.go
```diff
@@ -119,36 +131,66 @@ type LangMapperBase struct {
 	schematree *schemaTree
 }
 
-// SetEnumSet is used to supply a set of enumerated values to the
+// setEnumSet is used to supply a set of enumerated values to the
 // mapper such that leaves that have enumerated types can be looked up.
 //
 // NB: This method is a set-up method that the user does not need to be invoked
```
robshakir, line 137: Just reading this back `a set-up method that the user _does not_ need to be invoked`? Does this mean that the user doesn't need to invoke it, or that they don't care whether this is called? It's a bit confusing.

robshakir (changes requested):

ygen/ir.go
wenovus: Done.

ygen/ir.go
wenovus: Fixed up comment, PTAL if this makes sense now.

PR #695, opened 2022-06-06
Split Go generation into `gogen` package.

wenovus:
The change includes the following notable changes:
* gogen/codegen.go: Instead of using `ygen.NewYANGCodeGenerator` as the struct for all
  code generation from the single `ygen` package, `gogen.NewGoCodeGenerator` is created for Go
  generation. `GoOpts` is moved in `gogen`.
* gogen/genir_test: Since IR generation requires a suitable `LangMapper` implementation,
  the tests using `GoLangMapper` are moved to here.
* gogen/gogen.go, goelements.go: These are moved to this package.
* gogen/helpers.go: Go generation-specific helpers are moved here.
* internal/igenutil/genutil.go: Non-Go-specific generation helpers are
  moved here.
* ygen/genstate_test.go: A fake LangMapper implementation is created for
  unit testing.


generator/generator.go
```diff
@@ -111,7 +112,7 @@ var (
 // writeGoCodeSingleFile takes a ygen.GeneratedGoCode struct and writes the Go code
 // snippets contained within it to the io.Writer, w, provided as an argument.
 // The output includes a package header which is generated.
-func writeGoCodeSingleFile(w io.Writer, goCode *ygen.GeneratedGoCode) error {
+func writeGoCodeSingleFile(w io.Writer, goCode *gogen.GeneratedGoCode) error {
```
robshakir, line 115: `gogen.GeneratedGoCode` --> `gogen.GeneratedCode` perhaps?

gogen/codegen.go
```diff
@@ -0,0 +1,429 @@
...
+
+	"github.com/openconfig/ygot/internal/igenutil"
+	"github.com/openconfig/ygot/util"
+	"github.com/openconfig/ygot/ygen"
+	"github.com/openconfig/ygot/ygot"
+)
+
+// GoCodeGenerator is a structure that is used to pass arguments as to
+// how the output Go code should be generated.
+type GoCodeGenerator struct {
```
robshakir, line 17:
```suggestion
type CodeGenerator struct {
```

gogen/codegen.go
```diff
@@ -0,0 +1,429 @@
...
+	// Config stores the configuration parameters used for code generation.
+	Config ygen.GeneratorConfig
+	// GoOptions stores a struct which stores Go code generation specific
+	// options for the code generaton.
+	GoOptions GoOpts
+}
+
+// GoOpts stores Go specific options for the code generation library.
+type GoOpts struct {
```
robshakir, line 26:
```suggestion
type Opts struct {
```

gogen/codegen.go
```diff
@@ -0,0 +1,429 @@
...
+// the header with package comment of the generated code is returned in Header, along with the
+// a slice of strings containing the packages that are required for the generated Go code to
+// be compiled is returned.
+//
+// For schemas that contain enumerated types (identities, or enumerations), a code snippet is
+// returned as the EnumMap field that allows the string values from the YANG schema to be resolved.
+// The keys of the map are strings corresponding to the name of the generated type, with the
+// map values being maps of the int64 identifier for each value of the enumeration to the name of
+// the element, as used in the YANG schema.
+type GeneratedGoCode struct {
```
robshakir, line 119:
```suggestion
type GeneratedCode struct {
```

gogen/codegen.go
```diff
@@ -0,0 +1,429 @@
...
+	// RawJSONSchema stores the JSON document which is serialised and stored in JSONSchemaCode.
+	// It is populated only if the StoreRawSchema GoCodeGenerator boolean is set to true.
+	RawJSONSchema []byte
+	// EnumTypeMap is a Go map that allows YANG schemapaths to be mapped to reflect.Type values.
+	EnumTypeMap string
+}
+
+// NewGoCodeGenerator returns a new instance of the GoCodeGenerator
+// struct to the calling function.
+func NewGoCodeGenerator(c *ygen.GeneratorConfig, goopts *GoOpts) *GoCodeGenerator {
```
robshakir, line 144:
```suggestion
func NewCodeGenerator(c *ygen.GeneratorConfig, goopts *Opts) *CodeGenerator {
```

gogen/codegen.go
```diff
@@ -0,0 +1,429 @@
...
+// modules that are included by the specified set of modules, or submodules of those
+// modules). It extracts the set of modules that are to be generated, and returns
+// a GeneratedGoCode struct which contains:
+//	1. A struct definition for each container or list that is within the specified
+//	    set of models.
+//	2. Enumerated values which correspond to the set of enumerated entities (leaves
+//	   of type enumeration, identities, typedefs that reference an enumeration)
+//	   within the specified models.
+// If errors are encountered during code generation, an error is returned.
+func (cg *GoCodeGenerator) GenerateGoCode(yangFiles, includePaths []string) (*GeneratedGoCode, util.Errors) {
```
robshakir, line 188:
```suggestion
func (cg *GoCodeGenerator) Generate(yangFiles, includePaths []string) (*GeneratedCode, util.Errors) {
```

robshakir: Now this is just Go, then I think we can just call it `Generate` to be pithier.

gogen/codegen.go
```diff
@@ -0,0 +1,429 @@
...
+	if cg.Config.TransformationOptions.GenerateFakeRoot {
+		rootName = cg.Config.TransformationOptions.FakeRootName
+		if rootName == "" {
+			rootName = igenutil.DefaultRootName
+		}
+		if r, ok := ir.Directories[fmt.Sprintf("/%s", rootName)]; ok {
+			rootName = r.Name
+		}
+	}
+	commonHeader, oneoffHeader, err := writeGoHeader(yangFiles, includePaths, cg, rootName, ir.ModelData)
```
robshakir, line 213: Should we make `writeGoHeader` just `writeHeader`?

gogen/codegen.go
```diff
@@ -0,0 +1,429 @@
...
+				}
+				// Sort the enumerated types into schema order.
+				sort.Slice(enumTypeMap[schemaPath], func(i, j int) bool {
+					return field.LangType.UnionTypes[enumTypeMap[schemaPath][i]] < field.LangType.UnionTypes[enumTypeMap[schemaPath][j]]
+				})
+			}
+		}
+	}
+
+	processedEnums, err := genGoEnumeratedTypes(ir.Enums)
```
robshakir, line 291: Here we could make this `genEnumeratedTypes`, but these internal changes seem a bit less important than the public API suggestions.

gogen/goelements.go
```diff
@@ -292,18 +291,18 @@ func (s *GoLangMapper) PackageName(*yang.Entry, genutil.CompressBehaviour, bool)
 // The skipEnumDedup argument specifies whether leaves of type enumeration that are
 // used more than once in the schema should share a common type. By default, a single
 // type for each leaf is created.
-func (s *GoLangMapper) yangTypeToGoType(args resolveTypeArgs, compressOCPaths, skipEnumDedup, shortenEnumLeafNames, useDefiningModuleForTypedefEnumNames bool, enumOrgPrefixesToTrim []string) (*MappedType, error) {
+func (s *GoLangMapper) yangTypeToGoType(args resolveTypeArgs, compressOCPaths, skipEnumDedup, shortenEnumLeafNames, useDefiningModuleForTypedefEnumNames bool, enumOrgPrefixesToTrim []string) (*ygen.MappedType, error) {
 	defVal := genutil.TypeDefaultValue(args.yangType)
 	// Handle the case of a typedef which is actually an enumeration.
-	typedefName, _, err := s.EnumeratedTypedefTypeName(args, goEnumPrefix, false, useDefiningModuleForTypedefEnumNames)
+	typedefName, _, err := s.EnumeratedTypedefTypeName(args.yangType, args.contextEntry, goEnumPrefix, false, useDefiningModuleForTypedefEnumNames)
```
robshakir, line 297: Could you explain this change? It's not immediately obvious to me what happened here -- we changed the arguments to `EnumeratedTypedefTypeName` (since I think `args.contextEntry` is a `yang.Entry` whereas `goEnumPrefix` is a string).

gogen/gogen.go
```diff
@@ -1162,46 +1163,8 @@ func (t *{{ .ParentReceiver }}) To_{{ .Name }}(i interface{}) ({{ .Name }}, erro
...
-			return b.String()
-		},
-		// stripAsteriskPrefix provides a template helper that removes an asterisk
-		// from the start of a string. It is used to remove "*" from the start of
-		// pointer types.
-		"stripAsteriskPrefix": func(s string) string { return strings.TrimPrefix(s, "*") },
-	}
 )
 
-// mustMakeTemplate generates a template.Template for a particular named source
```
robshakir, line 1199: Since I think this induces a panic, I wonder whether we actually want to create it in a shared package -- I'm OK with doing so (since it's an internal package too), but it's essentially a small saving.

robshakir (changes requested):
Thanks for this -- this looks like a big step in encapsulating the Go logic in one place, great work!

.github/workflows/go.yml
```diff
@@ -150,6 +150,11 @@ jobs:
             go test -covermode count -coverprofile profile.coverprofile -outputdir $dir $p
           done
 
+          # Overwrite results for gogen/protogen since they also cover ygen's
+          # code.
+          go test -covermode count -coverprofile profile.coverprofile -outputdir gogen -coverpkg github.com/openconfig/ygot/gogen,github.com/openconfig/ygot/ygen github.com/openconfig/ygot/gogen
+          #go test -covermode count -coverprofile profile.coverprofile -outputdir protogen -coverpkg github.com/openconfig/ygot/protogen,github.com/openconfig/ygot/ygen github.com/openconfig/ygot/protogen
```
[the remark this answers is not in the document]
wenovus: To be added by the next PR, I didn't bother deleting it.

gogen/gogen.go
wenovus: Makes sense, I moved this back into gogen.

gogen/goelements.go
wenovus: So I took away the `resolveTypeArgs` argument type and expanded it out into the two parameters at the beginning; otherwise `resolveTypeArgs` would need to be exported, and since this is the only exported function where this is needed I judged that introducing another exported type (likely residing in `genutil`) to the user is not worth the overhead of changing all the existing usages.

gogen/codegen.go
wenovus: I didn't do this because I felt there was some clarity in answer the question "which header?"

gogen/codegen.go
wenovus: Same as above, I thought specifying that this is Go generated code, while redundant, is slightly clearer to a new reader of the codebase, which I would assume to be the vast majority of the audience.

gogen/goelements.go
robshakir: Ah, I see -- I had thought that the move there was mechnical. Great, this is fine with me.

gogen/codegen.go
robshakir: No strong preference here :-)

gogen/codegen.go
wenovus: This is one place where I wasn't convinced of removing the "Go" qualifier. The reason is `ygen.GeneratorConfig` also configures some generation options such as CompressBehaviour, the enum options (e.g. `SkipEnumDeduplication`) and others like `ExcludeModules` and `GenerateJSONSchema` (maybe this one should go into GoOpts), so I think calling it `GenOptions` could be confusing since users might assume it contains all of the generation options when in reality it only contains the Go-specific ones. What do you think?

Current `CodeGenerator` definition containing all generation options.
```go
// CodeGenerator is a structure that is used to pass arguments as to                                                                                                                                                                                                                                  
// how the output Go code should be generated.                                                                                                                                                                                                                                                        
type CodeGenerator struct {                                                                                                                                                                                                                                                                           
        // Config stores the configuration parameters used for code generation.                                                                                                                                                                                                                       
        Config ygen.GeneratorConfig                                                                                                                                                                                                                                                                   
        // GoOptions stores a struct which stores Go code generation specific                                                                                                                                                                                                                         
        // options for the code generaton.                                                                                                                                                                                                                                                            
        GoOptions GoOpts                                                                                                                                                                                                                                                                              
}                                                                                                                                                                                                                                                                                                     
```

`Config` seems like a bad name. Maybe `BasicConfig` or something is better?

wenovus: Proposal to address this here: https://github.com/openconfig/ygot/pull/706 PTAL

robshakir: Thanks, that looks a good solution to the problem. I'll resolve this thread.

robshakir (approved):

PR #696, opened 2022-06-06
Split Proto generation into `protogen` package.

wenovus:
The change includes the following notable changes:
* protogen/codegen.go: Instead of using `ygen.NewYANGCodeGenerator` as the struct for all
  code generation from the single `ygen` package, `protogen.NewProtoCodeGenerator` is created for Proto
  generation. `ProtoOpts` is moved to `protogen`.
* protogen/genir_test: Since IR generation requires a suitable `LangMapper` implementation,
  the tests using `ProtoLangMapper` are moved to here.
* protogen/protogen.go, protoelements.go: These are moved to this package.
* integration_tests/generate_errs_test.go: Moved `TestGenerateErrs` out of ygen.


proto_generator/protogenerator.go
```diff
@@ -100,7 +101,7 @@ func main() {
 	}
 
 	// Perform the code generation.
-	cg := ygen.NewYANGCodeGenerator(&ygen.GeneratorConfig{
+	cg := protogen.NewProtoCodeGenerator(&ygen.GeneratorConfig{
```
robshakir, line 104: Similar comments to the Go refactor, `protogen.NewGenerator` or even `protogen.New` might be more succinct here, since I expect that there's likely not much more than the generator in the package here.

proto_generator/protogenerator.go
```diff
@@ -115,7 +116,8 @@ func main() {
 		},
 		PackageName: *packageName,
 		Caller:      *callerName,
-		ProtoOptions: ygen.ProtoOpts{
+	},
+		&protogen.ProtoOpts{
```
robshakir, line 120: `protogen.Opts`.

protogen/codegen.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/openconfig/ygot/ygen"
+)
+
+// ProtoCodeGenerator is a structure that is used to pass arguments as to
+// how the output protobuf code should be generated.
+type ProtoCodeGenerator struct {
+	// Config stores the configuration parameters used for code generation.
+	Config ygen.GeneratorConfig
+	// ProtoOptions stores a struct which contains Protobuf specific options.
+	ProtoOptions ProtoOpts
```
robshakir, line 18:
```suggestion
	Options Opts
```

protogen/codegen.go
```diff
@@ -0,0 +1,257 @@
...
+	// GoPackageBase specifies the base of the names that are used in
+	// the go_package file option for generated protobufs. Additional
+	// package identifiers are appended to the go_package - such that
+	// the format <base>/<path>/<to>/<package> is used.
+	GoPackageBase string
+}
+
+// NewProtoCodeGenerator returns a new instance of the GoCodeGenerator
+// struct to the calling function.
+func NewProtoCodeGenerator(c *ygen.GeneratorConfig, protoopts *ProtoOpts) *ProtoCodeGenerator {
```
robshakir, line 63:
```suggestion
func New(c *ygen.GeneratorConfig, protoopts *Opts) *CodeGenerator {
```

protogen/codegen.go
```diff
@@ -0,0 +1,257 @@
...
+	UsesYwrapperImport bool     // UsesYwrapperImport indicates whether the ywrapper proto package is used within the generated package.
+	UsesYextImport     bool     // UsesYextImport indicates whether the yext proto package is used within the generated package.
+}
+
+// GenerateProto3 generates Protobuf 3 code for the input set of YANG files.
+// The YANG schemas for which protobufs are to be created is supplied as the
+// yangFiles argument, with included modules being searched for in includePaths.
+// It returns a GeneratedProto3 struct containing the messages that are to be
+// output, along with any associated values (e.g., enumerations).
+func (cg *ProtoCodeGenerator) GenerateProto3(yangFiles, includePaths []string) (*GeneratedProto3, util.Errors) {
```
robshakir, line 99:
```suggestion
func (cg *ProtoCodeGenerator) Generate(yangFiles, includePaths []string) (*GeneratedProto3, util.Errors) {
```

robshakir (changes requested):
Another big step, thanks! Suggestions here are similar to the Go ones for naming - likely we should aim for some consistency between the two (I had slightly different suggestions when looking at the Go version first).

proto_generator/protogenerator.go
wenovus: Proposal to address this here: https://github.com/openconfig/ygot/pull/706 PTAL

protogen/codegen.go
wenovus: Proposal to address this here: https://github.com/openconfig/ygot/pull/706 PTAL

robshakir (approved):

PR #698, opened 2022-06-07
Use bash instead of overalls

wenovus:
Exactly the same results:
* before: https://coveralls.io/builds/49771275
* after: https://coveralls.io/builds/49782021

This is to enable more flexible coverage reporting if there are cross-package unit tests.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -131,7 +131,6 @@ jobs:
 
       - name: Install required static analysis tools
         run: |
-          go install github.com/go-playground/overalls@latest
           go install github.com/mattn/goveralls@latest
           go install honnef.co/go/tools/cmd/staticcheck@latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -144,7 +144,17 @@ jobs:
 
       - name: Run coverage
         run: |
-          overalls -project ${GITHUB_WORKSPACE} -covermode=count -ignore=".git,vendor,integration_tests,ygot/schema_tests,ygen/schema_tests,ypathgen/path_tests,demo,experimental/ygotutils,generator,ytypes/schema_tests,ypathgen/generator"
+          pkgs="$(go list ./... | grep -v 'github.com/openconfig/ygot/exampleoc' | grep -v 'github.com/openconfig/ygot/uexampleoc' | grep -v 'github.com/openconfig/ygot/proto/' | grep -v 'github.com/openconfig/ygot/demo/' | grep -v 'github.com/openconfig/ygot/integration_tests' | grep -v 'github.com/openconfig/ygot/generator' | tr '\n' ' ')"
```
robshakir, line 147: At line 134, you can remove `go install github.com/go-playground/overalls@latest` based on removing this dependency.

robshakir (approved):

PR #699, opened 2022-06-07
Fix `IdentityType` classification bug.

wenovus:
Right now all typedefs are automatically categorized as `DerivedEnumerationType`; this is wrong because identity types could also be in a typedef.

Fixed this bug and added tests.


robshakir (changes requested):
Good catch -- where was this impacting code generation? It looks like the fix didn't change the generated code at all?

wenovus:
> Good catch -- where was this impacting code generation? It looks like the fix didn't change the generated code at all?

This is a rare case that doesn't impact Go generated code but does impact proto generated code. The proto generated code is actually missing coverage. Just added coverage for it now.

(I noticed this because after https://github.com/openconfig/ygot/pull/688/files#diff-68b0138e7c95ac856621a877f13fb98d5b98811830d066e622c36fcce550b2ec `exampleoc` panicked because of the misclassification)

robshakir (approved):
Great, thanks for adding coverage. The detail of the `exampleoc` panic is useful too, since that is something that we'll check in CI too.

PR #701, opened 2022-06-07
Codify that nil map marshals to null in JSON

wenovus:
By Hyrum's Law (which is already observed for this particular case (https://github.com/openconfig/ygot/pull/685)), I think we should codify this so that this doesn't change over time.

### DIFF
--- a/ygot/render_test.go
+++ b/ygot/render_test.go
@@ -4110,6 +4110,10 @@ func TestMarshal7951(t *testing.T) {
 		},
 		want: `{"str":"test-string"}`,
 	}, {
+		desc: "nil GoStruct",
+		in:   (*renderExample)(nil),
+		want: `null`,
+	}, {
 		desc: "simple GoStruct with PreferShadowPath",
 		in: &renderExample{
 			Str: String("test-string"),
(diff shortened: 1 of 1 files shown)


robshakir:
TIL what _Hyrum's Law_ is. Thanks :-)

This is interesting, because here we are explicitly introducing something that isn't specified in the RFC as a tested behaviour. Was #685 indicating that users wanted to see `"somelist": null` in the marshalled JSON?

wenovus:
> TIL what _Hyrum's Law_ is. Thanks :-)
> 
> This is interesting, because here we are explicitly introducing something that isn't specified in the RFC as a tested behaviour. Was #685 indicating that users wanted to see `"somelist": null` in the marshalled JSON?

#685 is a case where the user differentiates between the meaning of a null vs. [] (i.e. empty) list in JSON marshalled code (not explicitly stated in the PR but Eric confirmed with me offline), and this means they depend on this behaviour.

robshakir:
_Interesting_, in general, this seems broken to depend on this - but as you say, this demonstrates that anything that is in the API will be depended upon. 

I think it would be worth clarifying that we only produce this `null` string when you specifically ask for the map to be marshalled, if you do it from the parent, then it will be excluded (as expected).

Example code:

```golang
package main

import (
        "fmt"

        oc "github.com/openconfig/ygot/exampleoc"
        "github.com/openconfig/ygot/ygot"
)

func main() {

        d1 := &oc.Device{}
        d2 := &oc.Device{}
        d2.Interface = map[string]*oc.Interface{}

        for _, o := range []interface{}{d1, d2, d1.Interface} {
                js, err := ygot.Marshal7951(d)
                if err != nil {
                        panic(err)
                }
                fmt.Println(string(js))
        }
}
```

output:

```
robjs@penguin:~/go/src/robjs/ocd$ go run main.go 
{}
{"interfaces":{"interface":[]}}
null
```

robshakir (approved):

PR #702, opened 2022-06-07
Remove expectation that nil maps and slices (YANG lists) are initialized by `BuildEmptyTree`

wenovus:
Using `pretty.Compare` has masked this issue. Since it's possible users might have already depending on the current behaviour I wanted to be conservative and just keep the current behaviour, which I think is reasonable -- given the current generated code essentially we're saying we'd initialize containers but not lists.


robshakir (approved):
Agree with your assessment that this is something that is likely expected by users already.

PR #703, opened 2022-06-08
Run go generate ./... for CI check

wenovus:
Thanks for the comment here: https://github.com/openconfig/ygot/pull/699#pullrequestreview-998955556 I realized we weren't actually checking that the `u?exampleoc` packages actually regenerates with the current CI.


robshakir (approved):
SGTM, sent you #704 to update the tests.

PR #705, opened 2022-06-09
Delete `StoreRawSchema` from `GeneratorConfig` since it's not being used

wenovus:
I don't see the value of why we need this flag: is it to save memory?
For `exampleoc` `RawJSONSchema` took 27MB, a negligible amount.


robshakir:
At this point, I think it is generally not that useful. The original reason that this was a flag is that we didn't create this output by default. (Indeed, removing the flag might cause some tests to fail that don't expect the schema to be there.)

27MB of generated code is quite a lot - it can be the difference between being able to open the code in an editor and not, but I'm OK with removing these flags. We can probably achieve the same by having the schema stored in a separate file or similar.

robshakir (approved):

wenovus:
> At this point, I think it is generally not that useful. The original reason that this was a flag is that we didn't create this output by default. (Indeed, removing the flag might cause some tests to fail that don't expect the schema to be there.)
> 
> 27MB of generated code is quite a lot - it can be the difference between being able to open the code in an editor and not, but I'm OK with removing these flags. We can probably achieve the same by having the schema stored in a separate file or similar.

Agreed that 27MB of generated code is a lot. However reading this deleted doc comment, this flag was rather supposed to be controlling whether this variable is filled. So by 27MB I actually meant the memory usage of calling `generator.go` rather than the memory of the code editor.
```
	// It is populated only if the StoreRawSchema CodeGenerator boolean is set to true.
	RawJSONSchema []byte
```

wenovus:
> > At this point, I think it is generally not that useful. The original reason that this was a flag is that we didn't create this output by default. (Indeed, removing the flag might cause some tests to fail that don't expect the schema to be there.)
> > 27MB of generated code is quite a lot - it can be the difference between being able to open the code in an editor and not, but I'm OK with removing these flags. We can probably achieve the same by having the schema stored in a separate file or similar.
> 
> Agreed that 27MB of generated code is a lot. However reading this deleted doc comment, this flag was rather supposed to be controlling whether this variable is filled. So by 27MB I actually meant the memory usage of calling `generator.go` rather than the memory of the code editor.
> 
> ```
> 	// It is populated only if the StoreRawSchema CodeGenerator boolean is set to true.
> 	RawJSONSchema []byte
> ```

Hmm I misunderstood the meaning of `RawJSONSchema`. Looking at the integration test `RawJSONSchema` is actually meant to be written out to a json file, but there is actually no existing code to do this. @robshakir is it possible we're missing this functionality that was intended to be there?

robshakir:
Interesting -- I don't believe we ever did this, I think we always compressed the schema and stored it in the generated Go. The integration tests were potentially just taking a shortcut to validating the contents are correct, but I can't see a real use case for this.

Can we change the integration test to rather uncompress the gzip and compare this to the expected schema?

wenovus:
> Interesting -- I don't believe we ever did this, I think we always compressed the schema and stored it in the generated Go. The integration tests were potentially just taking a shortcut to validating the contents are correct, but I can't see a real use case for this.
> 
> Can we change the integration test to rather uncompress the gzip and compare this to the expected schema?

Do you mean compare `gotGeneratedCode.JSONSchemaCode` to the expected schema? The problem is we actually lose the pure gzip at this point because we run the gzipped schema through a text template for code generation first. So comparing the `RawJSONSchema` is currently the only clean way to actually get the JSON schema for comparing against a golden file.

I propose to leave this be since the raw JSON schema can be used by someone for whatever processing they need during the code generation stage, so it's not completely useless, and we can use it for testing as is currently done.

robshakir:
Ah, I see -- because we don't execute the code, we don't actually get the byte array, we just have the string.

Yes, I think in this case it's reasonable to leave it, since I think having an integration test that checks that the JSON generated is correct is more valuable than the limited simplification removing the flag gets us.

PR #706, opened 2022-06-09
Reorganize Code Generation flags and Delete `ygen.GeneratorConfig`

wenovus:
* ygen now only has `IROptions`.
* `CodeGenerator` in gogen/protogen now carry IROptions and their language-specific post-IR generation options.
* Code generation specific flags that are not used during IR generation are moved to `gogen` and `protogen`, duplicating if used by both (e.g. CallerName).
* `SkipEnumDeduplication` is moved from ParseOptions to TransformationOptions.
* Renamed gogen's `NewCodeGenerator` to just `New`.


robshakir (approved):
LGTM, this looks a neat solution.

PR #707, opened 2022-06-15
Add ability split path struct by file and module

DanG100:


wenovus (approved):

generator/generator.go
```diff
@@ -491,17 +497,24 @@ func main() {
...
+		return fmt.Errorf("error while splitting path structs code into %d files: %w", pathStructsFileN, err)
+	}
+	for i, file := range files {
+		out[fmt.Sprintf(pathStructsFileFmt, i)] = file
+	}
+	if err := writeFiles(dir, out); err != nil {
+		return fmt.Errorf("Error while writing path struct files: %w", err)
+	}
+	return nil
+}
```
wenovus, line 520: Actually, should there be a test for this?

wenovus (changes requested):

generator/generator.go
DanG100: i wasn't planning on it, it doesn't seem worth it because it's not very complex and kinda a pain to test. I can add some if you want

wenovus (approved):

generator/generator.go
DanG100: thanks, tested works with sample code in this repo and ondatra

PR #712, opened 2022-07-14
Add Go generation flag `ignore_unsupported`

wenovus:
This allows generating from YANG files with unsupported statements such as `notification` without erroring out.

Testing:
* Added integration test where the expected output is the same as if the `notification` statement was never there.


ygen/codegen.go
```diff
@@ -341,15 +344,18 @@ func findMappableEntities(e *yang.Entry, dirs map[string]*yang.Entry, enums map[
...
 			}
 		case ch.IsContainer(), ch.IsList():
 			dirs[ch.Path()] = ch
 			// Recurse down the tree.
-			errs = util.AppendErrs(errs, findMappableEntities(ch, dirs, enums, excludeModules, compressPaths, modules))
+			errs = util.AppendErrs(errs, findMappableEntities(ch, dirs, enums, excludeModules, compressPaths, ignoreUnsupportedStatements, modules))
 		case ch.Kind == yang.AnyDataEntry:
 			continue
 		default:
+			if ignoreUnsupportedStatements {
```
robshakir, line 356: Can we log something here about what we're skipping? I worry that since `ignoreUnsupported` is quite generic then it becomes difficult to understand what is /not/ included in the code based on this.

ygen/codegen.go
```diff
@@ -291,7 +294,7 @@ func mappedDefinitions(yangFiles, includePaths []string, opts IROptions) (*mappe
 // mapped with path compression enabled. The set of modules that the current code generation
 // is processing is specified by the modules slice. This function returns slice of errors
 // encountered during processing.
-func findMappableEntities(e *yang.Entry, dirs map[string]*yang.Entry, enums map[string]*yang.Entry, excludeModules []string, compressPaths bool, modules []*yang.Entry) util.Errors {
+func findMappableEntities(e *yang.Entry, dirs map[string]*yang.Entry, enums map[string]*yang.Entry, excludeModules []string, compressPaths, ignoreUnsupportedStatements bool, modules []*yang.Entry) util.Errors {
```
robshakir, line 297: Can we add a unit test for `findMappableEntities` that covers this case please?

robshakir (changes requested):

ygen/codegen.go
wenovus:
```
I0725 09:56:07.350343  442601 codegen.go:358] Unsupported statement (Notification) ignored: /openconfig-simple/update
```

ygen/codegen.go
wenovus: Done.

ygen/codegen.go
robshakir: LGTM :-) thanks.

robshakir (approved):

PR #713, opened 2022-07-14
Update all go.mod dependencies

wenovus:


wenovus:
Some of the modules require newer versions of Go (e.g. 1.17). I think it's not worth upgrading unless we know we need something from the dependencies.

PR #281, opened 2019-04-07
initial commit for leaf setters

adibrastegarnia:
Initial commit for adding leaf setters

PR #717, opened 2022-07-20
Generate leaf setters

kjahed:
This PR adds the '-generate_leaf_setters' flag which enables the generation of leaf setters (SetXXX). The generated methods assign their parameter (or its address) to the corresponding fields depending on the fields' data type.


wenovus (changes requested):
What you have LGTM, could you also set this flag in a few places where `GenerateLeafGetters:     true` is set in https://github.com/openconfig/ygot/blob/master/gogen/codegen_test.go? to make sure that the generated code compiles and that we have good coverage across the different YANG types.

wenovus (approved):
LGTM, thanks! @robshakir do you want to take a quick look before we merge this? There was a similar review you did in the past if you want to reload some context: https://github.com/openconfig/ygot/pull/281

gogen/gogen.go
```diff
@@ -1644,6 +1682,11 @@ func writeGoStruct(targetStruct *ygen.ParsedDirectory, goStructElements map[stri
 			errs = append(errs, err)
 		}
 	}
+	if goOpts.GenerateLeafSetters {
```
robshakir, line 1685:
```suggestion
	
	if goOpts.GenerateLeafSetters {
```

gogen/gogen.go
```diff
@@ -1644,6 +1682,11 @@ func writeGoStruct(targetStruct *ygen.ParsedDirectory, goStructElements map[stri
...
+		if err := generateLeafSetters(&methodBuf, associatedLeafSetters); err != nil {
+			errs = append(errs, err)
+		}
+	}
```
robshakir, line 1689:
```suggestion
	}

```

gogen/gogen_test.go
```diff
@@ -1843,6 +1843,156 @@ func (t *Container) ΛEnumTypeMap() map[string][]reflect.Type { return ΛEnumTyp
...
+			methods: `
+// SetLeafStr sets the value of the leaf LeafStr in the Container
+// struct.
+func (t *Container) SetLeafStr(v string) {
+	t.LeafStr = &v
+}
+
+// SetLeafUnion sets the value of the leaf LeafUnion in the Container
+// struct.
+func (t *Container) SetLeafUnion(v Container_U1_Union) {
```
robshakir, line 1934: Not blocking in this PR, but I wonder whether in the future we should support `SetLeafUnion<Type>` which would also hide needing to create the wrapper type. This would be similar to the way that the proto library does `oneof` fields.

robshakir (approved):

robshakir:
Thanks @wenovus - this LGTM. Comments are minor here, but please accept the suggestions just for code style reasons :-)

gogen/gogen_test.go
wenovus: Nice idea, unfortunately we can't easily do this right now because Go generics doesn't support methods whose parameters are generic because it complicates the interface system: https://go.dev/doc/faq#generic_methods

PR #723, opened 2022-09-01
Add DiffOpt for shadow-paths

wenovus:


ygot/diff.go
```diff
@@ -383,6 +390,14 @@ type DiffPathOpt struct {
 	// generated structs, which can result in duplication of list key leaves in
 	// the diff output.
 	MapToSinglePath bool
+	// PreferShadowPath specifies whether the "shadow-path" struct tag
+	// annotation should be used instead of the "path" struct tag when it
+	// exists.
+	//
+	// This option is used when GoStructs are generated with the
+	// -ignore_shadow_schema_paths flag, and therefore have the
```
robshakir, line 398: Just to clarify, the `shadow-path` is only populated when we specify that shadow paths should be ignored? Is that correct?

robshakir (approved):

ygot/diff.go
wenovus: Yes, the original motivation of this flag is to ignore these paths while unmarshalling, hence the flag name. The struct tag has since become more than that: in Ondatra it is used to allow GoStructs to be multiplexed to represent both config and state values.

I'm open to changing the flag name to `-generate_shadow_schema_paths` and just documenting that by default it means shadow paths are silently ignored during unmarshalling for ygot v1

PR #726, opened 2022-09-13
Unmarshal SetRequest and Notification to a root GoStruct.

wenovus:
Possible use cases:
* Maintaining a schema-aware view of the configuration at a gNMI server.
* Simulating the effects of a SetRequest to compare against results from a server.

### DIFF
--- a/ytypes/gnmi.go
+++ b/ytypes/gnmi.go
@@ -13,6 +13,10 @@ import (
 // UnmarshalNotifications unmarshals a Notification on a root GoStruct specified by
 // "schema", and returns a reference to it.
 //
+// It does not make a copy and overwrites this value, so make a copy using
+// ygot.DeepCopy() if you wish to retain the value at schema.Root prior to
+// calling this function.
+//
 // - If preferShadowPath is specified, then the shadow path values are
 // unmarshalled instead of non-shadow path values when GoStructs are generated
 // with shadow paths.
@@ -40,7 +44,11 @@ func UnmarshalNotifications(schema *Schema, ns []*gpb.Notification, preferShadow
 }
 
 // UnmarshalSetRequest applies a SetRequest on a root GoStruct specified by
-// "schema", and returns a reference to it.
+// "schema", and returns a reference to the resulting schema.Root.
+//
+// It does not make a copy and overwrites this value, so make a copy using
+// ygot.DeepCopy() if you wish to retain the value at schema.Root prior to
+// calling this function.
 //
 // - If preferShadowPath is specified, then the shadow path values are
 // unmarshalled instead of non-shadow path values when GoStructs are generated
--- a/ytypes/gnmi.go
+++ b/ytypes/gnmi.go
@@ -1,0 +1,150 @@
+package ytypes
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/openconfig/goyang/pkg/yang"
+	"github.com/openconfig/ygot/ygot"
+
+	gpb "github.com/openconfig/gnmi/proto/gnmi"
+)
+
+// UnmarshalNotification unmarshals a Notification on a root GoStruct specified by
+// "schema", and returns a reference to it.
+//
+// If preferShadowPath is specified, then the shadow path values are
+// unmarshalled instead of non-shadow path values when GoStructs are generated
+// with shadow paths.
+func UnmarshalNotification(schema *Schema, req *gpb.Notification, preferShadowPath bool) (ygot.GoStruct, error) {
+	return UnmarshalSetRequest(schema, &gpb.SetRequest{
+		Prefix: req.Prefix,
+		Delete: req.Delete,
+		Update: req.Update,
+	}, preferShadowPath)
+}
+
+// UnmarshalSetRequest applies a SetRequest on a root GoStruct specified by
+// "schema", and returns a reference to it.
+//
+// If preferShadowPath is specified, then the shadow path values are
+// unmarshalled instead of non-shadow path values when GoStructs are generated
+// with shadow paths.
+func UnmarshalSetRequest(schema *Schema, req *gpb.SetRequest, preferShadowPath bool) (ygot.GoStruct, error) {
+	root := schema.Root
+	node, nodeName, err := getOrCreateNode(schema.RootSchema(), root, req.Prefix, preferShadowPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Process deletes, then replace, then updates.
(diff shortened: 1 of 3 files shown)


ytypes/gnmi_test.go
```diff
@@ -0,0 +1,424 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/openconfig/goyang/pkg/yang"
+	"github.com/openconfig/ygot/ygot"
+
+	gpb "github.com/openconfig/gnmi/proto/gnmi"
+)
+
+func TestUnmarshalSetRequest(t *testing.T) {
```
wenovus, line 13: Still need to add test cases on shadowPaths, and errors.

wenovus:
@robshakir or @dplore Can we add @DanG100 to ygot-committers? https://github.com/orgs/openconfig/teams/ygot-committers/members

ytypes/gnmi.go
```diff
@@ -0,0 +1,150 @@
...
+	node, ok := nodeI.(ygot.GoStruct)
+	if !ok {
+		return nil, "", fmt.Errorf("gnmit: prefix path points to a non-GoStruct, this is not allowed: %T, %v", nodeI, nodeI)
+	}
+
+	return node, reflect.TypeOf(nodeI).Elem().Name(), nil
+}
+
+// deletePaths deletes a slice of paths from the given GoStruct.
+func deletePaths(schema *yang.Entry, goStruct ygot.GoStruct, paths []*gpb.Path, preferShadowPath bool) error {
```
DanG100, line 96: nit: you could change this func and replacePaths to accept ...DelOpt you only have build slice once.

ytypes/gnmi.go
```diff
@@ -0,0 +1,150 @@
...
+// node along with its name.
+func getOrCreateNode(schema *yang.Entry, goStruct ygot.GoStruct, path *gpb.Path, preferShadowPath bool) (ygot.GoStruct, string, error) {
+	var gcopts []GetOrCreateNodeOpt
+	if preferShadowPath {
+		gcopts = append(gcopts, &PreferShadowPath{})
+	}
+	// Operate at the prefix level.
+	nodeI, _, err := GetOrCreateNode(schema, goStruct, path, gcopts...)
+	if err != nil {
+		return nil, "", fmt.Errorf("gnmit: failed to GetOrCreate the prefix node: %v", err)
```
DanG100, line 85: here and below: remove "gnmit: " from err message

DanG100 (approved):

ytypes/gnmi.go
wenovus: It's also not elegant to have to pass in both DelOpt and preferShadowPath. I think I like the current slightly better.

ytypes/gnmi.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	node, ok := nodeI.(ygot.GoStruct)
+	if !ok {
+		return nil, "", fmt.Errorf("gnmit: prefix path points to a non-GoStruct, this is not allowed: %T, %v", nodeI, nodeI)
+	}
+
+	return node, reflect.TypeOf(nodeI).Elem().Name(), nil
```
robshakir, line 114: Do you need to check that `nodeI` is a struct pointer here (`util.IsStructPtr`?) - otherwise I think this can `panic`?

ytypes/gnmi.go
```diff
@@ -0,0 +1,150 @@
...
+// with shadow paths.
+func UnmarshalNotification(schema *Schema, req *gpb.Notification, preferShadowPath bool) (ygot.GoStruct, error) {
+	return UnmarshalSetRequest(schema, &gpb.SetRequest{
+		Prefix: req.Prefix,
+		Delete: req.Delete,
+		Update: req.Update,
+	}, preferShadowPath)
+}
+
+// UnmarshalSetRequest applies a SetRequest on a root GoStruct specified by
```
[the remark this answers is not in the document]
wenovus: What do you mean by "newly created SetRequest"? This function unmarshals any SetRequest into the root GoStruct specified by ytypes.Schema.

ytypes/gnmi.go
```diff
@@ -0,0 +1,150 @@
...
+// "schema", and returns a reference to it.
+//
+// If preferShadowPath is specified, then the shadow path values are
+// unmarshalled instead of non-shadow path values when GoStructs are generated
+// with shadow paths.
+func UnmarshalSetRequest(schema *Schema, req *gpb.SetRequest, preferShadowPath bool) (ygot.GoStruct, error) {
+	root := schema.Root
```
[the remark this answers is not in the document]
wenovus: I want this to have overwrite semantics because there could be multiple runs over the input struct. Also the GoStruct could already be populated to some extent. I'll clarify this in the comments.

I was debating what kind of input to provide. This seemed like the cleanest input that's the least subject to misuse: a root GoStruct is required, and the goyang schema is always required as well. In practice users would call `oc.Schema()` which seemed convenient. They would need to copy things out but that seemed OK to me.

ytypes/gnmi.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, fmt.Errorf("SetRequest is not schema-compliant: %v", err)
+	}
+
+	return vroot, nil
+}
+
+// validatedGoStruct is an interface used for validating GoStructs.
+// This interface is implemented by all Go structs (YANG container or lists),
+// regardless of generation flag.
+type validatedGoStruct interface {
```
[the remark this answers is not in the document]
wenovus: This is also an unexported type. I created this because I don't want to import the burden of ValidatedGoStruct, which has other functions that are unnecessary.

ytypes/gnmi.go
wenovus: It's already checking `node, ok := nodeI.(ygot.GoStruct)` so I think this won't happen in practice.

ytypes/gnmi.go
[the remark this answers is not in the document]
robshakir: Sorry - I meant to correct this comment to say that it specifically updates the root in the schema. (s/SetRequest/root/g) but then got distracted by the thinking about how to do the root.

ytypes/gnmi.go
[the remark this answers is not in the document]
robshakir: Why is this a "burden" are there cases where you have types that don't implement the functions? It seems a little odd to me, because whilst this is a valid interface, it duplicates something that we do have today.

ytypes/gnmi.go
robshakir: `ygot.GoStruct` can be implemented by something that has a non-pointer receiver though right?

PR #728, opened 2022-09-16
Incorporate `IgnoreExtraFields` as a `SetNodeOpt`

wenovus:


robshakir (approved):
LGTM. It's notable that this shouldn't be the default, but I'm OK with having it as an option.

(It should generally be an error when you receive an input JSON that isn't valid from a client.)

PR #729, opened 2022-09-16
Have UnmarshalSetRequest work with IgnoreExtraFields

wenovus:
Changed API to take in UnmarshalOpt (same as ytypes.Unmarshal)


ytypes/gnmi.go
```diff
@@ -22,13 +22,13 @@ import (
...
 // after all the notifications have been unmarshalled.
-func UnmarshalNotifications(schema *Schema, ns []*gpb.Notification, preferShadowPath, skipValidation bool) error {
+func UnmarshalNotifications(schema *Schema, ns []*gpb.Notification, skipValidation bool, opts ...UnmarshalOpt) error {
 	for _, n := range ns {
 		err := UnmarshalSetRequest(schema, &gpb.SetRequest{
 			Prefix: n.Prefix,
 			Delete: n.Delete,
 			Update: n.Update,
-		}, preferShadowPath, true)
+		}, true, opts...)
```
robshakir, line 31: Why did we change to hardcoding `preferShadowPath` here? You don't really know whether the code was generated with `config` or `state` as the shadow path AIUI. It also seems like it limits general functionality to one specific use case.

wenovus: `preferShadowPath` was changed to an option input, the function signature changed to
```
(schema *Schema, ns []*gpb.Notification, skipValidation bool, opts ...UnmarshalOpt)
```
The `true` you see is skipValidation.

robshakir: Ah, got it -- can you add a comment to just say why this is (shouldn't it also be an option?). It seems accepting a `Set` should probably mean that you did validate.

wenovus: Changed skipValidation -> validate.
Added this:
// validate is not an option since UnmarshalOpt is shared with
// ytypes.Unmarshal, and currently there is no expectation to support
// validation there.

This is not a strong reason -- we could just add validation to ytypes.Unmarshal and potentially changing its input from `interface{}` to `ygot.GoStruct`, it would make the parameters more clean, I don't have a strong opinion on this, but it does seem nicer. I can add it if you also have the same leaning.

wenovus:
@robshakir see most recent review comment reply

ytypes/gnmi.go
robshakir: The alternative is just to make the comment here that explicitly states that `UnmarshalSetRequest` will not perform validation, and implementations that expect a valid configuration to be set need to perform this validation on the returned config.

That probably is needed anyway, thinking more about it, since you could unmarshal into /interfaces for example, and leafref validation would need the root.

With this in mind, my preference is just to add to the comments that this explicitly always does not validate. 

[Just to make sure I'm self-consistent, this means whatever calls `UnmarshalSetRequest` needs to call `Validate` after it unmarshals, so the `Set` still has the `Validate` coverage.]

robshakir (approved):
LGTM, with the added comment suggested.

ytypes/gnmi.go
wenovus: SG, this makes for a cleaner API. I will add the comment that both functions will NOT validate, and that it is up to the caller of these functions to do so if they wish to validate.

wenovus: To help with this I'll add a Validate() method for ytypes.Schema in the next PR.

PR #730, opened 2022-09-20
Update to latest version of gnmi dependency

wenovus:
The effect of this changes is that now ygot will marshal decimal64 as DoubleVal instead of FloatVal or DecimalVal.

I will increment the minor version and announce this as a backwards-incompatible change.

Note: The test uses `float32` but this is never generated in Go code, so we didn't lose coverage.


wenovus:
@robshakir ping on this, thanks

Issue #735, opened 2022-10-14
ygot.TogNMINotifications generated proto message has underlying memory sharing that is not documented


shichuzhu:
https://github.com/openconfig/ygot/blob/6f722a0cce2a47949294afa0c3f23b080d51e501/ygot/render.go#L329

For demo purpose

```Go
func demo(root ygot.GoStruct) {
	ns, _ := ygot.TogNMINotifications(root, 0, ygot.GNMINotificationsConfig{
		UsePathElem: true,
	})
	updates := ns[0].GetUpdate()
	fmt.Println(updates[1].GetPath().GetElem()[0].Name) // prints "interfaces"
	updates[0].GetPath().GetElem()[0].Name = "somethingElse"
	fmt.Println(updates[1].GetPath().GetElem()[0].Name) // it now prints "somethingElse" !
}
```

It looks like in the generated Notification proto messages, paths of different Updates may map to the same memory location.
Yet this 'data sharing' is not mentioned in the `GNMINotificationsConfig` doc string.

If this is intended behavior for space optimization, consider adding a comment like
```
Within the generated notifications there could be data sharing for space optimization. Make a deep copy if one plans to modify the generated message
```

@wenovus @robshakir

wenovus:
Thanks Stuart for reporting this issue. At the very least this is a documentation issue. There are multiple spots in the code that need more documenting. Thanks Stuart for the suggested comment, I've added the documentation fix at https://github.com/openconfig/ygot/pull/736.

@robshakir I think the right fix is that by default, users **don't** have to call `Clone`, and that we provide an option within `GNMINotificationsConfig` to allow this optimization to keep the behaviour. If you agree either Stuart or I can change the default behaviour and add the flag.

-------------------
If you're curious the reason is because we're shallow-copying the PathElems:
https://github.com/openconfig/ygot/blob/1d36b4ad76e5cfb41a8b67c8c37ab725575afde4/ygot/render.go#L154-L158

But there doesn't look like a way to fix this without incurring a performance penalty:

```
!?master ~/gocode/src/github.com/openconfig/lsdbparse> go test -bench=. -cpuprofile cpu.out
goos: linux
goarch: amd64
pkg: github.com/openconfig/lsdbparse
BenchmarkRenderLSP/simple_example/usePathElem=false-12             35715             32565 ns/op
BenchmarkRenderLSP/larger_example/usePathElem=false-12              6960            209299 ns/op
[...]

PR #736, opened 2022-10-15
Comments for TogNMINotifications.

wenovus:
Background at https://github.com/openconfig/ygot/issues/735


ygot/render.go
```diff
@@ -143,7 +143,10 @@ func (g *gnmiPath) isPathElemPath() bool {
 	return g.pathElemPath != nil
 }
 
-// Copy returns a copy of the current gnmiPath.
+// Copy returns a shallow copy of the current gnmiPath.
+//
+// In particular, the pre-0.4.0 gNMI path is a deep copy, but the PathElem gNMI
```
robshakir, line 148: Should we make this a deeper copy? It'll have some overhead to do it with `proto.Clone`, but I presume that the shallow copy caused an issue somewhere?

robshakir (approved):

ygot/render.go
wenovus: I'll add a flag to `GNMINotificationsConfig` for doing deep copying.

The reason `proto.Clone` is not used in relevant locations is because it would negatively impact the performance of `ToGNMINotifications`: https://github.com/openconfig/ygot/pull/609

robshakir: Makes sense. What was the trigger for this CL? If we don't care about deep-copying then no need to add a flag, but I assume someone edited something somewhere and it caused issues?

wenovus: Yes Stuart was writing something to test config/state and was converting the config -> state and that caused the issue.

Here we can either make shallow-copy the default behaviour or deep-copy the default behaviour.

wenovus: I'd prefer shallow-copy to be the default behaviour since someone might be relying on the performance characteristic. So I'll merge this and add the DeepCopy flag.

wenovus: Actually I think it's ok that we simply keep the comment. The user can just deepcopy the returned notification per the doc comment.

PR #738, opened 2022-10-17
Change demo code to use `Validate` instead of `ΛValidate`

wenovus:


robshakir (approved):
Thanks for fixing this.

Issue #734, opened 2022-10-13
DeleteNode cannot handle a deletion to a key followed by a deletion to a non-key under a list entry


wenovus:
Because it assumes the list entry is schema-valid and tries to extract the keys from the key fields instead of from the map element.

This is a problem when GoStructs are being used to mirror a server's state via a gNMI library such as ygnmi, where subscribe responses could consist solely of leaf paths and can put the deletion paths in different orders (e.g. gnmi cache).

I think the right solution is to avoid doing a key value check when one or more of the keys don't exist when traversing a list entry, and simply extracting the key from the map key when executing the delete. Further, when the last leaf from a list entry is deleted, the list entry should be deleted from the map.

Alternatively we can add an option to DeleteNode to delete the entire list entry when the deletion path points to a list key. This however makes the assumption (for the use case above) that all the other leaves will be deleted as well, so it's not as robust of a solution. So, I think at best it's a workaround.

wenovus:
Initial fix here: https://github.com/openconfig/ygot/commit/cf0bfc7828c97b83835e1c2872a782944a1103cf

not done yet, look to get some feedback before finishing.

wenovus:
Given that openconfig/gnmi's match.go will send all leaf updates in order to be explicit, as well as it being easier to modify ygot, I'll implement this idea to solve the issue.

PR #741, opened 2022-11-03
Add `ygot.GoKeyStruct` alongside `ygot.GoStruct`

wenovus:
This allows `ΛListKeyMap` to be called on the generated multi-key structs as well. The use case is for
https://github.com/openconfig/ygot/issues/734


gogen/gogen.go
```diff
@@ -167,8 +167,10 @@ func (g GoStructCodeSnippet) String() string {
 
 // goStructField contains a definition of a field within a Go struct.
 type goStructField struct {
-	Name string // Name is the field's name.
-	Type string // Type is the Go type of the field.
+	// YANGName is the YANG name of the key field.
```
robshakir, line 170: Consistently keep this comment to the right, or change the others to be the line above?

robshakir (approved):

gogen/gogen.go
wenovus: Done.

PR #742, opened 2022-11-03
`DeleteNode` Improvement

### LINKED ISSUE #734: DeleteNode cannot handle a deletion to a key followed by a deletion to a non-key under a list entry

wenovus:
Two changes:
1. Deleting a non-key field of a list element when a key field has already been deleted is valid, and does not trigger an error.
2. DeleteNode now will prune empty branches on the deletion path:
```diff
 // DeleteNode zeroes the value of the node specified by the supplied path from
 // the specified root, whose schema must also be supplied. If the node
 // specified by that path is already its zero value, or an intermediate node
-// in the path is nil (implying the node is already deleted), then the call is a no-op.
+// in the path is nil (implying the node is already deleted), then the deletion
+// operation is not executed.
+//
+// Regardless of whether the deletion operation is executed, any intermediate
+// non-leaf nodes traversed by the path that is equal to the empty struct or
+// map will be set to nil, similar to the behaviour of ygot.PruneEmptyBranches.
 func DeleteNode(schema *yang.Entry, root interface{}, path *gpb.Path, opts ...DelNodeOpt) error {
```

Fixes #734 

Consideration:
* We might want to add a flag for DeleteNode to not have behaviour (2) above, but I think it is a reasonable default behaviour especially now that we're supporting behaviour (1), where it doesn't make sense to have a completely-empty list element.


robshakir (approved):
LGTM, thanks -- this looks a good change to me for robustness' sake.

One quick question -- should we have this controlled by a flag, or do we feel that this is generally the only reasonable way to do this?

wenovus:
> LGTM, thanks -- this looks a good change to me for robustness' sake.
> 
> One quick question -- should we have this controlled by a flag, or do we feel that this is generally the only reasonable way to do this?

Now that we're supporting deleting a list element until it's empty, it doesn't make sense to me to allow the empty element to remain in the map after the deletion. Therefore I leaned towards not having a flag for it.

PR #749, opened 2022-11-23
Add support for mapping union YANG fields in protobufs.

robshakir:


protogen/protogen.go
```diff
@@ -232,8 +232,17 @@ message {{ .Name }} {
...
     {{- range $ooField := .OneOfFields }}
-    {{ $ooField.Type }} {{ $ooField.Name }} = {{ $ooField.Tag }};
+    {{ $ooField.Type }} {{ $ooField.Name }} = {{ $ooField.Tag }}
+    {{- if ne $noOptions 0 }} [
+      {{- range $i, $opt := $field.Options -}}
+        {{ $opt.Name }} = {{ $opt.Value -}}
+        {{- if ne (inc $i) $noOptions -}}, {{- end }}
+      {{- end -}}
+      ]
+      {{- end -}}
```
wenovus, line 244: nit: match this indent with the `if`

protogen/protogen_test.go
```diff
@@ -264,6 +264,65 @@ func TestGenProto3Msg(t *testing.T) {
 				}},
 			},
 		},
+	}, {
+		name: "union leaf with annotate schema paths enabled",
```
wenovus, line 268: This is similar to the test case "simple message with union leaf and leaf-list". Could these be merged into one?

wenovus (approved):
Approving to clarify the above are style comments.

protogen/protogen.go
robshakir: Done.

protogen/protogen_test.go
robshakir: It could be - but I think adding these integration tests makes sense since it means that we capture the cases where this isn't enabled and is enabled.

wenovus: I think this will always be enabled because `MappedPaths: [][]string{{"", "field-one"}},` will always be there in a real case, and this means the `Options` field will always be there too. The only difference between these two test cases is that `inMsg.Fields.MappedPaths` and `wantMsgs.Fields.Options` are both populated in this one.

robshakir: True -- I think there is a simpler case here where we're just looking at a single field, we could add them to a particular case, but this test failing says "path annotation code is broken", whereas combining the two means we don't know whether it was the generation or the annotation. I'm OK with merging it if you feel strongly, but I have a mild preference for it to separate.

wenovus: I see your argument now. Sounds good. I was looking at it from the perspective of limiting the number of test cases we have, whereas your argument is about debugging failures when they occur. So no strong opinion from me.

wenovus (approved):

PR #763, opened 2023-01-11
Address gobugs issues.

robshakir:
```
 * (M) protogen/protogen.go
  - check for correct errors being nil.
  - remove dependency on cmp in non-testing code.
```


wenovus (approved):
Where did you see these errors?

robshakir:
They were caught by internal tooling as I imported the code.

PR #769, opened 2023-02-02
Fix genenerator's flag descriptions

### LINKED ISSUE #768: Generator options and `structs_split_files_count`
👋 all,
I see that in the generator flags we mention that `structs_split_files_count` works only when `output_file` is specified.

https://github.com/openconfig/ygot/blob/faf395436e2fae19420fa3638114ddfb1909bea5/generator/generator.go#L58

But I think the description is false, because we actually need to not use `output_file` and use `output_dir` instead.

If you agree with that, I can file a PR with fixing the flag description.

n0shut:
fix #768 

Hi @wenovus, does this match what you suggested in #768

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -57,7 +57,7 @@ var (
 	ocStructsOutputFile     = flag.String("output_file", "", "The file that the generated Go code for manipulating YANG data (schema/Go structs) should be written to. Specify \"-\" for stdout.")
 	structsFileN            = flag.Int("structs_split_files_count", 0, "The number of files to split the generated schema structs into when output_dir is specified.")
 	ocPathStructsOutputFile = flag.String("path_structs_output_file", "", "The file that the generated Go code for YANG path construction (path structs) will be generated. If split_pathstructs_by_module=true, this file contains the fake root path struct. Specify \"-\" for stdout.")
-	pathStructsFileN        = flag.Int("path_structs_split_files_count", 0, "The number of files to split the generated path structs into when output_dir is specified for generating path structs")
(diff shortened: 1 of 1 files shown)


generator/generator.go
```diff
@@ -55,9 +55,9 @@ var (
...
 	ocStructsOutputFile     = flag.String("output_file", "", "The file that the generated Go code for manipulating YANG data (schema/Go structs) should be written to. Specify \"-\" for stdout.")
-	structsFileN            = flag.Int("structs_split_files_count", 0, "The number of files to split the generated schema structs into when output_file is specified.")
+	structsFileN            = flag.Int("structs_split_files_count", 0, "The number of files to split the generated schema structs into when output_dir is specified.")
 	ocPathStructsOutputFile = flag.String("path_structs_output_file", "", "The file that the generated Go code for YANG path construction (path structs) will be generated. If split_pathstructs_by_module=true, this file contains the fake root path struct. Specify \"-\" for stdout.")
-	pathStructsFileN        = flag.Int("path_structs_split_files_count", 0, "The number of files to split the generated path structs into when output_file is specified for generating path structs")
+	pathStructsFileN        = flag.Int("path_structs_split_files_count", 0, "The number of files to split the generated path structs into when output_dir is specified for generating path structs")
```
wenovus, line 60:
```suggestion
	pathStructsFileN        = flag.Int("path_structs_split_files_count", 0, "The number of files to split the generated path structs into when output_dir or split_pathstructs_by_module is specified for generating path structs")
```

wenovus (changes requested):
Yep, almost, just wanted to add one more

n0shut:
thanks @wenovus!

wenovus (approved):

PR #772, opened 2023-02-07
Add utility `flattenOCJSON` in package `gnmidiff`

wenovus:
This is useful for comparing between two SetRequests whose contents are in JSON_IETF format.

It only intends to support OpenConfig-compliant JSON because it is otherwise impossible to infer the list keys from RFC7951 JSON.

@marcushines

### DIFF
--- a/gnmidiff/json.go
+++ b/gnmidiff/json.go
@@ -1,0 +1,134 @@
+// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gnmidiff
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+
+	"github.com/openconfig/ygot/ygot"
+)
+
+// flattenOCJSON outputs all leaf path-value pairs in the root per RFC7951.
+//
+// It assumes the input JSON is with respect to OpenConfig-compliant YANG.
+// See the following for checking compliance:
+// * https://github.com/openconfig/oc-pyang
+// * https://github.com/openconfig/public/blob/master/doc/openconfig_style_guide.md
+//
+// Output path format is per gNMI's path conventions.
+//
+// e.g.
+// {
+//   "openconfig-network-instance:config": {
+//     "description": "VRF RED",
+//     "enabled": true,
+//     "enabled-address-families": [
+//       "openconfig-types:IPV4",
+//       "openconfig-types:IPV6"
+//     ],
+//     "name": "RED",
+//     "type": "openconfig-network-instance-types:L3VRF"
+//     },
+//   "openconfig-network-instance:name": "RED"
+// }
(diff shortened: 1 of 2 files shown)


gnmidiff/json.go
```diff
@@ -0,0 +1,134 @@
...
+// returns
+// {
+//   "openconfig-network-instance:config/description": "VRF RED",
+//   "openconfig-network-instance:config/enabled": true,
+//   "openconfig-network-instance:config/enabled-address-families": ["openconfig-types:IPV4", "openconfig-types:IPV6"],
+//   "openconfig-network-instance:config/name": "RED",
+//   "openconfig-network-instance:config/type": "openconfig-network-instance-types:L3VRF",
+//   "openconfig-network-instance:name": "RED",
+// }
+func flattenOCJSON(json7951 []byte) (map[string]interface{}, error) {
```
robshakir, line 58: Curious: given that you'd like `<path, value>` tuples, why not unmarshal this and then serialise it with `TogNMINotifications`? This way you don't end up with fragility around the particular YANG schema that you're dealing with AFAICS?

Sure I'm missing something, but I couldn't quite tell why this was the preferred implementation.

wenovus:
For more context see https://github.com/openconfig/ygot/pull/773 which would follow this PR.

gnmidiff/json.go
robshakir: The key extraction seems fragile -- but I agree that if you want to work with no knowledge of the schema, then there is little choice.

robshakir (approved):
This LGTM, it's relatively clean in terms of implementation, but I think generally, it's probably good to provide something more robust in the future where we have the schema accessible.

gnmidiff/json.go
wenovus: Makes sense, I'll make using schema as the default behaviour, and add a flag to invoke this implementation.

gnmidiff/json.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	return leaves, nil
+}
+
+func flattenOCJSONAux(root interface{}, path string, leaves map[string]interface{}) error {
+	switch v := root.(type) {
+	case bool, float64, string:
+		leaves[path] = root
+	case []interface{}:
+		if len(v) == 0 {
```
marcushines, line 76: why isn't this just a recursive call back to itself?

marcushines: flattenOCJSONaux(root interface[], path, leaves, alreadyList)
if alreadylist return...
then handle all teh same cases?

gnmidiff/json.go
```diff
@@ -0,0 +1,134 @@
...
+//   "openconfig-network-instance:config/type": "openconfig-network-instance-types:L3VRF",
+//   "openconfig-network-instance:name": "RED",
+// }
+func flattenOCJSON(json7951 []byte) (map[string]interface{}, error) {
+	// TODO: Add option to remove the namespace on paths of returned updates.
+	var root interface{}
+	if err := json.Unmarshal(json7951, &root); err != nil {
+		return nil, fmt.Errorf("gnmidiff: %v", err)
+	}
+	leaves := map[string]interface{}{}
```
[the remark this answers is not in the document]
wenovus: I think it's ok to keep the key as a string since this function will handle one TypedValue from a `gpb.Update` at a time, and therefore will only handle a single origin at a time.

gnmidiff/json.go
wenovus: When `len(v) == 0` or when `v[0]` is a scalar, then this is actually a leaf-list, so it is a terminal case that needs to be handled on its own.
Otherwise, it is a list, and the recursive case needs to first handle the list key extraction before calling `flattenOCJSONAux` again.

PR #773, opened 2023-02-07
SetRequestDiff (non-schema-aware)

wenovus:


wenovus:
@marcushines pinging for review

gnmidiff/gostruct_test.go
```diff
@@ -0,0 +1,154 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gnmidiff
+
+import "github.com/openconfig/ygot/ygot"
+
+// Device represents the /device YANG schema element.
```
robshakir, line 19: Given that this verbatim reproduces part of the schema, consider using `exampleoc`?

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,236 @@
...
+// MismatchedUpdate represents two different update values for the same leaf
+// node.
+type MismatchedUpdate struct {
+	// A is the update value in A.
+	A interface{}
+	// B is the update value in B.
+	B interface{}
+}
+
+// SetRequestDiff contains the intent difference between two SetRequests.
```
robshakir, line 37: Is `intent` the right word here? Is it really the difference between the result of two `SetRequest` messages?

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,236 @@
...
+// by ygot.PathToString.
+// - The value of the update fields is the JSON_IETF representation of the
+// value.
+type SetRequestDiff struct {
+	AOnlyDeletes      map[string]struct{}
+	BOnlyDeletes      map[string]struct{}
+	CommonDeletes     map[string]struct{}
+	AOnlyUpdates      map[string]interface{}
+	BOnlyUpdates      map[string]interface{}
+	CommonUpdates     map[string]interface{}
```
robshakir, line 49: Is it ever of interest to a caller whether updated values were as a result of a replace or an update operation in the `SetRequest`? I couldn't immediately think of a case where it would be, but wanted to check.

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,236 @@
...
+// If newSchema is nil, then DiffSetRequest will make the following assumption:
+// - Any JSON value in the input SetRequest MUST conform to the OpenConfig
+// YANG style guidelines. See the following for checking compliance.
+// * https://github.com/openconfig/oc-pyang
+// * https://github.com/openconfig/public/blob/master/doc/openconfig_style_guide.md
+//
+// Currently, support is only for SetRequests without any delete paths, and
+// replace and updates that don't have conflicting leaf values. If not
+// supported, then an error will be returned.
+func DiffSetRequest(a *gpb.SetRequest, b *gpb.SetRequest, newSchema func() (*ytypes.Schema, error)) (SetRequestDiff, error) {
```
robshakir, line 66: The title here says non-schema aware, but we take in the schema. Maybe I'm missing something here but can you explain the reason `newSchema` is needed here?

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,236 @@
...
+// Note: The input path must NOT end with "/".
+//
+// e.g. /a for b/c="foo" would introduce an update of /a/b/c="foo" into the intent.
+func populateUpdates(intent *setRequestIntent, path string, tv *gpb.TypedValue) error {
+	if len(path) > 0 && path[len(path)-1] == '/' {
+		return fmt.Errorf("gnmidiff: invalid input path %q, must not end with \"/\"", path)
+	}
+	var leafVal interface{}
+	isLeaf := true
+	switch tv.GetValue().(type) {
```
robshakir, line 199: You could use gnmi's `value.ToScalar` if you handle the non-scalar values first I think.

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,236 @@
...
+			for subpath, value := range updates {
+				pathToLeaf := path + subpath
+				if _, ok := intent.Updates[pathToLeaf]; ok {
+					return fmt.Errorf("gnmidiff: leaf value set twice in SetRequest: %v", pathToLeaf)
+				}
+				intent.Updates[pathToLeaf] = value
+			}
+		}
+	default:
+		return fmt.Errorf("unsupported TypedValue type %T (only string and JSON_IETF are currently supported)", tv.GetValue())
```
robshakir, line 223: bool looks supported :)

gnmidiff/setrequest.go
wenovus: Yes. Two SetRequests can look different but have the same meaning. See the test case "not same but same intent"

gnmidiff/setrequest.go
wenovus: I have `// TODO: implement newSchema handling.`
I will get this done in a later PR.

gnmidiff/setrequest.go
wenovus: `ToScalar` maps to a Go native type whereas a JSON type helps for consistency in comparison. e.g. for a JSON `5` is that a double or an integer?

ygot/pathstrings.go
```diff
@@ -234,6 +234,16 @@ func StringToStructuredPath(path string) (*gnmipb.Path, error) {
 	return gpath, nil
 }
 
+// MustStringToPath calls StringToStructuredPath and panics on error.
+// It is intended for use in tests with hard-coded strings.
+func MustStringToPath(path string) *gnmipb.Path {
```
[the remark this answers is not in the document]
wenovus: Moved to testutil.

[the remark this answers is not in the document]
wenovus: Actually, this is a problem -- it creates a circular dependency because `ygot` depends on `testutil`. The standard library uses this Must pattern, e.g. https://pkg.go.dev/regexp#MustCompile

I think this is ok since it's widely used in the standard library, what do you think?

gnmidiff/setrequest.go
wenovus: This makes sense. I can add a feature where these updates and deletes are each associated with the `Update` message in the SetRequest. If there is a need then we can add it.

ygot/pathstrings.go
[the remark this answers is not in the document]
wenovus: I've split this code to https://github.com/openconfig/ygot/pull/775 for easier review.

gnmidiff/setrequest.go
wenovus: I've split this code to https://github.com/openconfig/ygot/pull/775 for easier review.

gnmidiff/setrequest.go
wenovus: implementation at https://github.com/openconfig/ygot/pull/776

gnmidiff/setrequest.go
```diff
@@ -17,13 +17,183 @@ package gnmidiff
...
+	AOnlyDeletes      map[string]struct{}
+	BOnlyDeletes      map[string]struct{}
+	CommonDeletes     map[string]struct{}
+	AOnlyUpdates      map[string]interface{}
+	BOnlyUpdates      map[string]interface{}
+	CommonUpdates     map[string]interface{}
+	MismatchedUpdates map[string]MismatchedUpdate
+}
+
+// Format is the string format of any gNMI diff utility.
```
bormanp, line 56: "any gNMI diff utility"?  Do you really mean "any" here?

gnmidiff/setrequest.go
```diff
@@ -17,13 +17,183 @@ package gnmidiff
...
+	// ExcludeList is a list of paths that will be excluded from the output.
+	// wildcards are allowed.
+	//
+	// empty implies no paths are excluded.
+	//ExcludeList []string
+}
+
+func formatJSONValue(value interface{}) interface{} {
+	if v, ok := value.(string); ok {
+		return strconv.Quote(v)
```
bormanp, line 75: It is probably the case, but I need to ask, are we sure that the value was not already quoted?

gnmidiff/setrequest.go
```diff
@@ -17,13 +17,183 @@ package gnmidiff
...
+	diff := SetRequestIntentDiff{
+		AOnlyDeletes:      map[string]struct{}{},
+		BOnlyDeletes:      map[string]struct{}{},
+		CommonDeletes:     map[string]struct{}{},
+		AOnlyUpdates:      map[string]interface{}{},
+		BOnlyUpdates:      map[string]interface{}{},
+		CommonUpdates:     map[string]interface{}{},
+		MismatchedUpdates: map[string]MismatchedUpdate{},
+	}
+	for pathA := range intentA.Deletes {
```
bormanp, line 166: I think you can replace lines 166 through 177 with:

```
for path := range intentA.Deletes {
        if _, ok := intentB.Deletes[path]; ok {
                delete(intentA.Deletes, path)
                delete(intentB.Deletes, path)
                diff.CommonDeletes[path] = struct{}{}
        }
}
```
And then `intentA.Deletes` is your `diff.AOnlyDeletes` and same for B.

gnmidiff/setrequest.go
```diff
@@ -17,13 +17,183 @@ package gnmidiff
...
+		} else {
+			diff.AOnlyDeletes[pathA] = struct{}{}
+		}
+	}
+	for pathB := range intentB.Deletes {
+		if _, ok := intentA.Deletes[pathB]; !ok {
+			diff.BOnlyDeletes[pathB] = struct{}{}
+		}
+	}
+	for pathA, vA := range intentA.Updates {
```
bormanp, line 178: Similar to above:

```
for path, vA := range intentA.Updates {
        vB, ok := intentB.Updates[pathA]
        if !ok {
                continue
        }
        delete(intentA.Updates, path)
        delete(intentB.Updates, path)
        if vA != vB {
                diff.MismatchedUpdates[pathA] = MismatchedUpdate{A: vA, B: vB}
        } else {
                diff.CommonUpdates[pathA] = vA
        }
}
```

gnmidiff/setrequest_test.go
```diff
@@ -20,10 +20,543 @@ import (
...
+				"/interfaces/interface[name=eth2]/state/transceiver":                                     "FDM",
+			},
+			MismatchedUpdates: map[string]MismatchedUpdate{
+				"/interfaces/interface[name=eth0]/subinterfaces/subinterface[index=0]/state/logical": {
+					A: false,
+					B: true,
+				},
+			},
+		},
+	}}
```
bormanp, line 545: Did you have any "duplicate but out of order" updates where the sorted intent looks the same but the results are actually different?  I think the stuffing these into a map in order will address that but it would be nice to be sure.

gnmidiff/setrequest.go
wenovus: Yes. In that case the delete will be stored in the `delete` field of `setRequestIntent`. The deletions in the `setRequestIntent` always precede the `update` field. So it's basically reducing the initial SetRequest down to just deletes and updates.

gnmidiff/setrequest.go
wenovus: Added "in this package." The utilities I can imagine (Set vs. Set, Set vs. Get, Get vs. Get) would all use this.

gnmidiff/setrequest.go
wenovus: Since the stored JSON values are already unmarshalled into Go types, JSON strings have already been de-quoted before being passed to this function for human-readable formatting.

gnmidiff/setrequest.go
wenovus: Done.

gnmidiff/setrequest.go
wenovus: Done.

gnmidiff/setrequest_test.go
wenovus: I'm not 100% sure what you mean by "duplicate but out of order", but right now conflicting leaf updates produce an error. This limitation can be removed later. Once it's removed, the update that occurs later will replace the previous one via the map key, as you've mentioned.

bormanp: I was meaning that set request A has two updates for the same node (say node X) with the first update setting it to Y, and the second to Z.  Request B also has two updates for node X, but first setting it to Z and then setting it to Y.  Request A will result in node X being Z and request B will result in node X being Y.  So both requests have the same two update requests for X but they are in different orders in each request.  Perhaps this is not possible, I don't know.

wenovus: This is possible, and the general solution which this implementation is intended to be adaptable towards will capture this by over-writing the new value of node X with the last update seen in the SetRequest. It's just not in this MVP.

DanG100 (approved):

PR #774, opened 2023-02-08
Add flag in flattenOCJSON to remove namespace from path

wenovus:


gnmidiff/json.go
```diff
@@ -123,7 +127,12 @@ func flattenOCJSONAux(root interface{}, path string, leaves map[string]interface
 	case map[string]interface{}:
 		// This is a container or a list element.
 		for subpath, subv := range v {
-			if err := flattenOCJSONAux(subv, path+"/"+subpath, leaves); err != nil {
+			if !keepNamespace {
+				if idx := strings.IndexRune(subpath, ':'); idx != -1 {
```
robshakir, line 131: I'm not sure which is more efficient, but this seemed a bit more logical to me -- completely aesthetic though, so feel free to ignore this.

```golang
pp := strings.Split(subpath, ":")
subpath = pp[len(pp)-1]
```

gnmidiff/json.go
```diff
@@ -55,20 +56,23 @@ import (
 //   "openconfig-network-instance:config/type": "openconfig-network-instance-types:L3VRF",
 //   "openconfig-network-instance:name": "RED",
 // }
-func flattenOCJSON(json7951 []byte) (map[string]interface{}, error) {
-	// TODO: Add option to remove the namespace on paths of returned updates.
+//
+// When keepNamespace=false, then any namespace is removed from the flattened
+// *paths*, but still kept in any enum values.
```
robshakir, line 61: (nit) identity values, since enumerations don't have this namespacing.

gnmidiff/json.go
wenovus: Nice one, updated. yes I was aiming for some aesthetics here.

gnmidiff/json.go
wenovus: Done.

PR #775, opened 2023-02-08
`minimalSetRequestIntent(req *gpb.SetRequest)`

wenovus:
This was already reviewed as part of https://github.com/openconfig/ygot/pull/773 but is now split from it.

@marcushines for review

### DIFF
--- a/gnmidiff/setrequest.go
+++ b/gnmidiff/setrequest.go
@@ -1,0 +1,147 @@
+// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// gnmidiff contains gNMI utilities for diffing SetRequests and GetResponses.
+package gnmidiff
+
+import (
+	"fmt"
+
+	"github.com/derekparker/trie"
+	"github.com/openconfig/ygot/ygot"
+
+	gpb "github.com/openconfig/gnmi/proto/gnmi"
+)
+
+type setRequestIntent struct {
+	Deletes map[string]struct{}
+	Updates map[string]interface{}
+}
+
+// minimalSetRequestIntent returns a unique and minimal intent for a SetRequest.
+//
+// Currently, support is only for SetRequests without any delete paths, and
+// replace and updates that don't have conflicting leaf values. If not
+// supported, then an error will be returned.
+func minimalSetRequestIntent(req *gpb.SetRequest) (setRequestIntent, error) {
+	if req == nil {
+		req = &gpb.SetRequest{}
+	}
+	if len(req.Delete) > 0 {
+		return setRequestIntent{}, fmt.Errorf("gnmidiff: delete paths are not supported.")
+	}
+	intent := setRequestIntent{
+		Deletes: map[string]struct{}{},
+		Updates: map[string]interface{}{},
+	}
+	// NOTE: This simple trie will not work if we intend to check conflicts with wildcard deletion paths.
+	t := trie.New()
+	for _, upd := range req.Replace {
+		path, err := ygot.PathToString(upd.Path)
+		if err != nil {
+			return setRequestIntent{}, fmt.Errorf("gnmidiff: %v", err)
+		}
+		if _, ok := intent.Deletes[path]; ok {
+			return setRequestIntent{}, fmt.Errorf("gnmidiff: conflicting replaces in SetRequest: %v", path)
+		}
+		intent.Deletes[path] = struct{}{}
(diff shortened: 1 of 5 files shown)


gnmidiff/setrequest_test.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+}
+
+func TestMinimalSetRequestIntent(t *testing.T) {
+	tests := []struct {
+		desc         string
+		inSetRequest *gpb.SetRequest
+		wantIntent   setRequestIntent
+		wantErr      bool
+	}{{
```
bormanp, line 46: Can you add a test that has a set request with updates for something like:

"/interfaces/port"
"/interfaces/portname"

I realize these are no valid openconfig paths, but they test the issue where one leaf is a prefix of another.

bormanp (changes requested):

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+type setRequestIntent struct {
+	Deletes map[string]struct{}
+	Updates map[string]interface{}
+}
+
+// minimalSetRequestIntent returns a unique and minimal intent for a SetRequest.
+//
+// Currently, support is only for SetRequests without any delete paths, and
```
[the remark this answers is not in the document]
wenovus: Done. This is in consideration of the currently-identified use case since supporting these will be complicated.

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,147 @@
...
+func minimalSetRequestIntent(req *gpb.SetRequest) (setRequestIntent, error) {
+	if req == nil {
+		req = &gpb.SetRequest{}
+	}
+	if len(req.Delete) > 0 {
+		return setRequestIntent{}, fmt.Errorf("gnmidiff: delete paths are not supported.")
+	}
+	intent := setRequestIntent{
+		Deletes: map[string]struct{}{},
+		Updates: map[string]interface{}{},
```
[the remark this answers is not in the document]
wenovus: It's because ygot still supports down to Go 1.13. When we move up then we can convert.

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	// NOTE: This simple trie will not work if we intend to check conflicts with wildcard deletion paths.
+	t := trie.New()
```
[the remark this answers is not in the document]
wenovus: Thank you for bringing this up, gnmi/ctree may help us here. For now without wildcards this simple trie works, but `ctree` supports wildcards, which will be helpful.

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,147 @@
...
+		t.Add(path, nil)
+
+		if err := populateUpdates(&intent, path, upd.GetVal()); err != nil {
+			return setRequestIntent{}, err
+		}
+	}
+
+	// Do prefix match to check for conflicting replace paths.
+	for _, path := range t.Keys() {
+		if matches := t.PrefixSearch(path); len(matches) >= 2 {
```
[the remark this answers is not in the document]
wenovus: Good point! This is a bug, I'll take a look.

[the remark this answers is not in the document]
wenovus: Ok, fixed by searching for `path + "/"`

gnmidiff/setrequest_test.go
wenovus: Done.

DanG100 (approved):

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,149 @@
...
+
+	"github.com/derekparker/trie"
+	"github.com/openconfig/ygot/ygot"
+
+	gpb "github.com/openconfig/gnmi/proto/gnmi"
+)
+
+type setRequestIntent struct {
+	// Deletes are deletions to any node.
+	Deletes map[string]struct{}
```
DanG100, line 29: do these fields need to be exported?  if the struct isnt

gnmidiff/setrequest.go
```diff
@@ -0,0 +1,149 @@
...
+
+	return intent, nil
+}
+
+// populateUpdates populates all leaf updates at the given path into the intent.
+//
+// Note: The input path must NOT end with "/".
+//
+// e.g. /a for b/c="foo" would introduce an update of /a/b/c="foo" into the intent.
+func populateUpdates(intent *setRequestIntent, path string, tv *gpb.TypedValue) error {
```
DanG100, line 106: this could be a method of setRequestIntent

gnmidiff/setrequest.go
wenovus: Is there a guideline on this? I just wanted to make it capitalized in case one day we do want to export it.

gnmidiff/setrequest.go
wenovus: Good point, let me update this in a later PR. There are some conflicting changes later on in the chain.

gnmidiff/setrequest.go
DanG100: i'm not aware of any, generally i prefer not exported things until they are needed. if the plan is to export this in the future this is fine as is

bormanp (approved):

gnmidiff/setrequest.go
wenovus: I will export it in a later PR.

PR #776, opened 2023-02-09
Schema aware setrequest diff

wenovus:
I did notice this implementation is much more expensive computation-time-wise compared to the schema-non-aware version. Will take a look at it.

EDIT:
The performance is improved a lot by avoiding having the ungzip the schema, but it's still not super fast. I looked at the profile and I didn't see any big improvement opportunities (a lot of it still was ungzipping the schema, which only happens once even for a large SetRequest, and the rest of it was GC), so I'm content to wait until this gets used more widely first.


gnmidiff/setrequest.go
```diff
@@ -267,52 +271,173 @@ func minimalSetRequestIntent(req *gpb.SetRequest) (setRequestIntent, error) {
 	return intent, nil
 }
 
+// binaryBase64 takes an input byte slice and returns it as a base64
+// encoded string.
+func binaryBase64(i []byte) string {
+	var b bytes.Buffer
+	encoder := base64.NewEncoder(base64.StdEncoding, &b)
```
bormanp, line 278: How about `return base64.StdEncoding.EncodeToString(i)`?

gnmidiff/setrequest.go
```diff
@@ -267,52 +271,173 @@ func minimalSetRequestIntent(req *gpb.SetRequest) (setRequestIntent, error) {
...
-			leafVal = val
-		} else {
-			isLeaf = false
-			for subpath, value := range updates {
-				pathToLeaf := path + subpath
-				if _, ok := intent.Updates[pathToLeaf]; ok {
-					return fmt.Errorf("gnmidiff: leaf value set twice in SetRequest: %v", pathToLeaf)
+
+	// Populate updates when the schema is unknown.
+	if newSchemaFn == nil {
```
bormanp, line 326: Would it be possible to put lines 327-365 into a function (say unknownSchemaUpdate) and then here just say
```
if newSchema == nil {
    return unknownSchemaUpdate(...)
}
```

I think that will make this function. more readable.

gnmidiff/setrequest.go
```diff
@@ -267,52 +271,173 @@ func minimalSetRequestIntent(req *gpb.SetRequest) (setRequestIntent, error) {
...
+	}
+}
+
 // populateUpdates populates all leaf updates at the given path into the intent.
 //
 // Note: The input path must NOT end with "/".
 //
 // e.g. /a for b/c="foo" would introduce an update of /a/b/c="foo" into the intent.
-func populateUpdates(intent *setRequestIntent, path string, tv *gpb.TypedValue) error {
+func populateUpdates(intent *setRequestIntent, path string, tv *gpb.TypedValue, newSchemaFn func() (*ytypes.Schema, error)) error {
```
bormanp, line 320: Please add a text about what newSchemaFn is for in the above comment.

bormanp, line 320: Is there a reason you don't simply just pass in a `ytypes.Schema`?

bormanp (changes requested):

gnmidiff/setrequest.go
wenovus: Done.

gnmidiff/setrequest.go
wenovus: Done. Also added a comment block to explain how the schema-aware updates are flattened.

gnmidiff/setrequest.go
wenovus: Done.

wenovus: This is to start with a clean root object each time unmarshalling happens. Otherwise there needs to be some `reflect` code that resets the root object each time to avoid previous unmarshalling done in previous invocations of `populateUpdates` polluting the results. Now you could argue why not just unmarshal everything and then send the updates. This would be valid aside from the corner case with list keys being populated, which I described in the new comment block I added.

Added comment explaining this.

wenovus: Done. This improved performance by a lot.

DanG100 (approved):

bormanp (approved):

PR #780, opened 2023-02-14
Add Format function for `SetToNotifsDiff`

wenovus:


gnmidiff/set_to_get.go
```diff
@@ -15,18 +15,32 @@ import (
...
 }
 
+// Format outputs the SetToNotifsDiff in human-readable format.
+//
+// NOTE: Do not depend on the output of this being stable.
+func (diff SetToNotifsDiff) Format(f Format) string {
+	f.title = "SetToNotifsDiff"
+	f.aName = "want/SetRequest"
+	f.bName = "got/Notifications"
+	return SetRequestIntentDiff{
```
DanG100, line 36: i'd extract the common logic of Format into its own struct instead using the SetRequestIntentDiff struct

```
type StructuredDiff struct {} 
func (fd StructuredDiff) Format()

type SetToNotifsDiff struct { StructuredDiff } 
type SetRequestIntentDiff struct { StructuredDiff } 
```

wenovus: Made more modular, PTAL

gnmidiff/diff.go
```diff
@@ -0,0 +1,179 @@
...
+// by ygot.PathToString.
+// - The value of the update fields is the JSON_IETF representation of the
+// value. This is to facilitate comparing JSON_IETF-represented values whose
+// real data type is obscured without knowledge of the YANG schema.
+type StructuredDiff struct {
+	DeleteDiff
+	UpdateDiff
+}
+
+type DeleteDiff struct {
```
DanG100, line 46: godoc comment for this type

gnmidiff/diff.go
```diff
@@ -0,0 +1,179 @@
...
+	MissingDeletes map[string]struct{}
+	ExtraDeletes   map[string]struct{}
+	CommonDeletes  map[string]struct{}
+}
+
+type UpdateDiff struct {
```
DanG100, line 52: godoc comment for the type

DanG100 (approved):

PR #783, opened 2023-02-15
Return deprecation error message when unmarshalling using SetNode/JSON

wenovus:
Should use JSON_IETF instead.

This is backwards-compatible because it was never handling JSON correctly, this change merely gives a better error message than before.


ytypes/node.go
```diff
@@ -223,6 +223,8 @@ func retrieveNodeContainer(schema *yang.Entry, root interface{}, path *gpb.Path,
 						if err := json.Unmarshal(args.val.(*gpb.TypedValue).GetJsonIetfVal(), &val); err != nil {
 							return nil, status.Errorf(codes.Unknown, "failed to update struct field %s in %T with value %v; %v", ft.Name, root, args.val, err)
 						}
+					case args.val.(*gpb.TypedValue).GetJsonVal() != nil:
```
DanG100, line 226: does this need a test?

DanG100 (approved):

ytypes/node.go
wenovus: Added.

PR #784, opened 2023-02-15
CLI for diffing between two SetRequests

wenovus:
Dynamic loading would be required for schema-aware diff due to the need to use ygot-generated GoStructs. For CLI I think the easiest for now is to just allow schema-unaware diffing.

This diff utility ideally resides in gnmic, although for the current stage of development it is useful to have the CLI reside in ygot right now.

I didn't create a ygot command because I want to be careful about introducing a ygot command.


gnmidiff/cmd/root.go
```diff
@@ -0,0 +1,51 @@
...
+		viper.BindPFlags(cmd.Flags())
+		viper.AutomaticEnv()
+		return nil
+	}
+
+	rootCmd.AddCommand(newSetRequestDiffCmd())
+
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
DanG100, line 49: i think its cleaner if the main() calls os.Exit(), i prefer either
1. the func returns to the root command and you call cmd.Execute in the main
2. return an error here and set this exit code in the main

DanG100 (approved):

gnmidiff/cmd/root.go
wenovus: Done.

PR #785, opened 2023-02-15
Remove CI support for Go 1.13, 1.14, 1.15, 1.16, 1.17

wenovus:
The pre-1.17 versions are introducing problems for PRs such as https://github.com/openconfig/ygot/actions/runs/4207239816/jobs/7301781404 in https://github.com/openconfig/ygot/pull/784

Furthermore, I wanted to remove support for 1.17 since 1.18 is the first version of generics, and it allows us to introduce generics into ygot.

I believe this change is reasonable since 1.20 has just been released, so officially only 1.18 is no longer released, and ygot supports an extra version (1.18) as stated in the README.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: ['1.13', '1.14', '1.15', '1.16', '1.17', '1.18', '1.19', '1.20', '1.x']
+        go: ['1.18', '1.19', '1.20', '1.x']
 
     steps:
       - name: Install protobuf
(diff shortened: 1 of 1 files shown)


wenovus:
@robshakir Do you think it's time to move to generics or do you think we should hold off at version 1.17?

robshakir:
Go 1.17 was released 2021-08-16, Go 1.18 was released 2022-03-15. Given https://github.com/googleapis/google-cloud-go and similar projects then I'm OK with the approach of removing support for 1.17 officially.

What I'd ask is whether we plan to do a LSC (e.g., `s/interface{}/any/g` for example throughout the code) or whether this just opens the door for when we think about new code moving to generics?

wenovus:
> Go 1.17 was released 2021-08-16, Go 1.18 was released 2022-03-15. Given https://github.com/googleapis/google-cloud-go and similar projects then I'm OK with the approach of removing support for 1.17 officially.
> 
> What I'd ask is whether we plan to do a LSC (e.g., `s/interface{}/any/g` for example throughout the code) or whether this just opens the door for when we think about new code moving to generics?

Yes I think an LSC is good to have at some point after we move to 1.18. Although here I intend primarily for new changes to be written in generics, by allowing us to start thinking in generics rather than the old way simply because there was one more deprecated version of Go that would be nice for us to support. WDYT?

wenovus:
@robshakir Do you think this work for you? If so a reminder to also update the CI merge policy so this becomes mergeable, thanks!

robshakir:
SGTM.

robshakir (approved):

wenovus:
Thanks @robshakir!, reminder to change the `Required` CI settings for ygot (I don't have permission).

robshakir:
@wenovus I made `1.x` the only required check. You should have the ability to change these, but it seems like that would be sufficient.

robshakir:
Made you an admin @wenovus.

PR #792, opened 2023-03-09
Remove usage of the proto Syntax call.

robshakir:
Remove use of the protoreflect `Syntax` API which is being migrated away from.

### DIFF
--- a/protomap/proto.go
+++ b/protomap/proto.go
@@ -141,7 +141,7 @@ func (m unpopRange) Range(f func(protoreflect.FieldDescriptor, protoreflect.Valu
 		}
 
 		v := m.Get(fd)
-    if fd.HasPresence() {
+		if fd.HasPresence() {
 			v = protoreflect.Value{} // use invalid value to emit null
 		}
 		if !f(fd, v) {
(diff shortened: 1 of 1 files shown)


protomap/proto.go
```diff
@@ -141,9 +141,7 @@ func (m unpopRange) Range(f func(protoreflect.FieldDescriptor, protoreflect.Valu
 		}
 
 		v := m.Get(fd)
-		isProto2Scalar := fd.Syntax() == protoreflect.Proto2 && fd.Default().IsValid()
-		isSingularMessage := fd.Cardinality() != protoreflect.Repeated && fd.Kind() == protoreflect.MessageKind
-		if isProto2Scalar || isSingularMessage {
+    if fd.HasPresence() {
 			v = protoreflect.Value{} // use invalid value to emit null
```
wenovus, line 145: I wonder why the intent looks different but otherwise LGTM

wenovus (approved):

protomap/proto.go
robshakir: Oops, `gofmt` CI should have caught this. Did we remove it?

robshakir: Fixed the CI in #793.

PR #823, opened 2023-04-28
Make ordered maps a pointer in the parent GoStruct.

wenovus:
Initially I made the choice to use a non-pointer so that the user would not need to initialize this field before using it.

The advantage is all methods can hang off of the type rather than needing methods to hang off the parent type in order to nil-check and initialize.

However, this creates problems that a pointer field solves:
1. Avoid needing value-receivers in order to make the field satisfy certain interfaces (e.g. ygot.GoStruct). Mixing receiver types is confusing for the reader.
2. Helps avoid awkward reflection code that need to access the pointer receiver methods since `CanAddr()=false` for non-pointer objects: https://go.dev/play/p/5vYVVUS-GfD
3. Users of ygot are used to declaring every struct as pointers, so using pointers makes things congruent with existing generated code.

Testing: test code in TestOrderedMapFromParent is just copied from TestOrderedMap and with parent methods replacing the local methods.

### DIFF
--- a/gogen/internal/gotypes/ordered_map.go
+++ b/gogen/internal/gotypes/ordered_map.go
@@ -30,7 +30,29 @@ type RoutingPolicy_PolicyDefinition struct {
 	// value is usable. This is unlike the regular unordered lists where
 	// the various helpers need to reside on the parent struct in order to
 	// avoid a nil pointer exception on the raw map type.
-	Statement RoutingPolicy_PolicyDefinition_Statement_OrderedMap
+	Statement *RoutingPolicy_PolicyDefinition_Statement_OrderedMap
+}
+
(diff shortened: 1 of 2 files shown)


gogen/internal/gotypes/ordered_map.go
```diff
@@ -30,7 +30,29 @@ type RoutingPolicy_PolicyDefinition struct {
 	// value is usable. This is unlike the regular unordered lists where
 	// the various helpers need to reside on the parent struct in order to
 	// avoid a nil pointer exception on the raw map type.
-	Statement RoutingPolicy_PolicyDefinition_Statement_OrderedMap
+	Statement *RoutingPolicy_PolicyDefinition_Statement_OrderedMap
+}
+
+func (s *RoutingPolicy_PolicyDefinition) AppendNewStatement(key string) (*RoutingPolicy_PolicyDefinition_Statement, error) {
```
DanG100, line 36: these should have godoc comments

DanG100 (approved):

Issue #790, opened 2023-03-04
ygot does not support default integer values expressed in hexadecimal format


dbourkey:
I came across this error when running the generator (v0.26.0) against the adtran-xpon.yang model which contained leafnodes in the following format:

`
      leaf name-of-leaf {
        type uint16;
        default 0x8100;
        description
          "...";
      }
`

` $ ~/bin/generator -ignore_unsupported -path=adtran_yang -output_file=adtran_yang.go -package_name=adtran_yang adtran_yang/adtran-xpon.yang
F0304 22:56:10.747789   65264 generator.go:382] ERROR Generating GoStruct Code: default value conversion: unable to convert default value "0x8100" to uint16: strconv.ParseUint: parsing "0x8100": invalid syntax
`

May I ask, was there an intentional decision to not support hexedecimal?

Fix would be simple enough.

wenovus:
Apologies for the late reply. Thanks @jayzhudev for the fix. I think this should be working now.

PR #827, opened 2023-05-01
fix(goelements): handle Hex default int/uint values

jayzhudev:
This PR can help handle the issue described in #790 . Please let me know if there are comments or changes you'd suggest. Thanks.


jayzhudev:
Update: the CLA check is passing now.

gogen/goelements.go
```diff
@@ -642,16 +642,24 @@ func (s *GoLangMapper) yangDefaultValueToGo(value string, args resolveTypeArgs,
 		if err != nil {
 			return "", yang.Ynone, err
 		}
+
+		// In some YANG modules the default value could be written as base 16.
+		base := 10
+		if strings.HasPrefix(strings.ToLower(value), "0x") {
```
wenovus, line 648: (optional) Would be helpful if this PR also added support for octal numbers, so "0" as the leading string.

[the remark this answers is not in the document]
jayzhudev: Thanks a lot for the comment and the reference link. This is fixed in the latest commit. I checked the documentation of `strconv.ParseInt` and `strconv.ParseUint`. It turns out that the standard library already supports base inference for hexadecimal, octal, and binary integer formats. It supports sign prefix handling as well. So a simpler solution would be specifying base `0` to enable base inference.

A remaining question is that, should binary integer format trigger an error return?

Thanks!

gogen/goelements.go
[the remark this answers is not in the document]
jayzhudev: Added test cases to `TestYangDefaultValueToGo`.

gogen/goelements.go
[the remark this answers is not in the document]
wenovus: I think we can keep it simple and simply handle octal and hex as in the RFC. There are other quirks in `strconv.ParseInt` (e.g. integer literals) that we may not necessarily want to support.

gogen/goelements.go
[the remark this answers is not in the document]
wenovus: Thanks for adding the thorough tests!

gogen/goelements.go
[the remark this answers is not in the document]
jayzhudev: Thanks for the comment. Agreed, I added a couple checks to make sure that base 2 values and underscores cannot be used in default values.

gogen/goelements.go
```diff
@@ -642,16 +642,39 @@ func (s *GoLangMapper) yangDefaultValueToGo(value string, args resolveTypeArgs,
...
+			if len(value) > 1 && (value[1] == 'b' || value[1] == 'B') {
+				return "", yang.Ynone, fmt.Errorf("default value conversion: base 2 value `%s` is not allowed", value)
+			}
+		}
+
+		// Underscores are not allowed in the value, although `strconv` allows them in certain formats.
+		if strings.Contains(value, "_") {
+			return "", yang.Ynone, fmt.Errorf("default value conversion: `_` is not allowed in the value %s", value)
+		}
+
```
wenovus, line 667: Instead of checking for negative cases, it might be cleaner to check for positive cases and then use ParseInt's error handling to deal with all other unexpected cases. So I'm thinking replacing this highlighted block with

```go
base := 10
switch {
case strings.HasPrefix(value, "0"):
  base = 8
case strings.HasPrefix(value, "0x"):
  base = 16
  value = strings.TrimLeft(value, "0x")
}

Comment (jayzhudev) [reply gogen/goelements.go:667]: The RFC allows use of an optional `+/-` sign prefix for octal and hexadecimal values, so it will need a few more lines to handle all cases... In fact the previous commit needs to handle the binary negative case with a possible sign prefix too.  Personally I'd prefer adding the sign handling to the previous commit's negative case handling. But let me know your thoughts. Thanks.

Comment (wenovus) [reply gogen/goelements.go]: Thanks Jay, and sorry for the delay in review. I'm open to keeping the current PR but I'm wondering how difficult it is to check the positive cases?

```go
if strings.HasPrefix(value, "+") {
  value = value[1:]
}

base := 10
switch {
case strings.HasPrefix(value, "0"), strings.HasPrefix(value, "-0"):
  base = 8
case strings.HasPrefix(value, "0x"), ...:
  base = 16
  value = strings.TrimLeft(value, "0x")
}
```

This looks shorter than the negative case handling and is more understandable. Are there other cases which I've missed here?

jayzhudev: Thanks! This should work similarly : ) But I think it still needs a couple more lines to return an error when seeing values with underscores.

jayzhudev: The binary format `0b` needs to be checked before the switch statement too. Otherwise, the first switch case will let `0b` pass through and give it base 8.

wenovus: I see, even the current code is not handling the underscores. I think this looks fine to me, thanks!

gogen/goelements.go
```diff
@@ -644,7 +644,8 @@ func (s *GoLangMapper) yangDefaultValueToGo(value string, args resolveTypeArgs,
 		}
 
 		// Check if the value is an empty string.
-		if len(value) == 0 {
+		signStripedValue := strings.TrimRight(strings.TrimLeft(value, "-"), "+")
```
wenovus, line 647: Why is this TrimRight rather than TrimLeft?

jayzhudev: Oops, thanks for catching it! I guess my auto-completion failed me when I was typing too fast... The tests didn't catch this because of `uint` typos of the YANG types (caused by copy-pasta), and `ParseUint` returned an error anyway because of the sign prefix... I've fixed them in the latest commits.

wenovus: Thanks

wenovus (approved):

PR #828, opened 2023-05-01
perf(logging): prevent unnecessary processing for debug logging

jayzhudev:
The inline processing (while calling debug logging functions) can consume significant amount of CPU resources at scale regardless if debug logging is disabled. This commit prevents such unnecessary processing.

This commit also includes performance improvements from reducing expensive string split calls and using more performant int/uint to ascii conversions.

### DIFF
--- a/util/path.go
+++ b/util/path.go
@@ -25,15 +25,20 @@ import (
 
 // SchemaPaths returns all the paths in the path tag.
 func SchemaPaths(f reflect.StructField) ([][]string, error) {
-	var out [][]string
 	pathTag, ok := f.Tag.Lookup("path")
 	if !ok || pathTag == "" {
 		return nil, fmt.Errorf("field %s did not specify a path", f.Name)
 	}
 
+	// Early return for performance.
+	if strings.Index(pathTag, "|") == -1 {
+		return [][]string{stripModulePrefixes(strings.Split(pathTag, "/"))}, nil
+	}
+
 	ps := strings.Split(pathTag, "|")
-	for _, p := range ps {
-		out = append(out, stripModulePrefixes(strings.Split(p, "/")))
+	out := make([][]string, len(ps))
+	for i, p := range ps {
+		out[i] = stripModulePrefixes(strings.Split(p, "/"))
 	}
 	return out, nil
 }
@@ -41,15 +46,15 @@ func SchemaPaths(f reflect.StructField) ([][]string, error) {
 // ShadowSchemaPaths returns all the paths in the shadow-path tag. If the tag
 // doesn't exist, a nil slice is returned.
 func ShadowSchemaPaths(f reflect.StructField) [][]string {
-	var out [][]string
 	pathTag, ok := f.Tag.Lookup("shadow-path")
 	if !ok || pathTag == "" {
 		return nil
 	}
 
 	ps := strings.Split(pathTag, "|")
-	for _, p := range ps {
-		out = append(out, stripModulePrefixes(strings.Split(p, "/")))
+	out := make([][]string, len(ps))
+	for i, p := range ps {
+		out[i] = stripModulePrefixes(strings.Split(p, "/"))
 	}
 	return out
 }
@@ -126,11 +131,12 @@ func relativeSchemaPath(f reflect.StructField, preferShadowPath bool) ([]string,
 		return nil, fmt.Errorf("field %s did not specify a shadow-path", f.Name)
 	}
 
-	paths := strings.Split(pathTag, "|")
-	if len(paths) == 1 {
+	if strings.Index(pathTag, "|") == -1 {
 		pathTag = strings.TrimPrefix(pathTag, "/")
 		return strings.Split(pathTag, "/"), nil
 	}
+
+	paths := strings.Split(pathTag, "|")
 	for _, pv := range paths {
 		pv = strings.TrimPrefix(pv, "/")
 		pe := strings.Split(pv, "/")
@@ -314,9 +320,9 @@ func StripModulePrefixesStr(in string) string {
 // stripModulePrefixes returns "in" with each element with the format "A:B"
 // changed to "B".
 func stripModulePrefixes(in []string) []string {
-	var out []string
-	for _, v := range in {
-		out = append(out, StripModulePrefix(v))
+	out := make([]string, len(in))
+	for i, v := range in {
+		out[i] = StripModulePrefix(v)
 	}
 	return out
 }
@@ -341,15 +347,13 @@ func stripModulePrefixWithCheck(name string) (string, error) {
 // removing foo from "foo:bar". Such qualified paths are used in YANG modules
 // where remote paths are referenced.
 func StripModulePrefix(name string) string {
-	ps := strings.Split(name, ":")
-	switch len(ps) {
-	case 1:
-		return name
-	case 2:
-		return ps[1]
-	default:
-		return name
+	for i, r := range name {
+		if r == ':' && len(name) > i+1 {
+			return name[i+1:]
+		}
 	}
+
+	return name
 }
 
 // ReplacePathSuffix replaces the non-prefix part of a prefixed path name, or
--- a/util/reflect.go
+++ b/util/reflect.go
@@ -172,7 +172,9 @@ func IsStructValueWithNFields(v reflect.Value, n int) bool {
 
 // InsertIntoSlice inserts value into parent which must be a slice ptr.
 func InsertIntoSlice(parentSlice interface{}, value interface{}) error {
-	DbgPrint("InsertIntoSlice into parent type %T with value %v, type %T", parentSlice, ValueStrDebug(value), value)
+	if DebugLibraryEnabled() {
+		DbgPrint("InsertIntoSlice into parent type %T with value %v, type %T", parentSlice, ValueStrDebug(value), value)
+	}
 
 	pv := reflect.ValueOf(parentSlice)
 	t := reflect.TypeOf(parentSlice)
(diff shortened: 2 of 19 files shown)


jayzhudev:
Note: I'm working on getting my username added to our CLA group. Will keep you posted : )

jayzhudev:
Update: the CLA check is passing now.

util/path.go
```diff
@@ -341,15 +347,13 @@ func stripModulePrefixWithCheck(name string) (string, error) {
...
-	ps := strings.Split(name, ":")
-	switch len(ps) {
-	case 1:
-		return name
-	case 2:
-		return ps[1]
-	default:
-		return name
+	for i, r := range name {
+		if r == ':' && len(name) > i+1 {
```
wenovus, line 351: Why is the `len(name) > i+1` necessary? Can this term just be omitted from the if predicate?

util/reflect.go
```diff
@@ -172,7 +172,9 @@ func IsStructValueWithNFields(v reflect.Value, n int) bool {
 
 // InsertIntoSlice inserts value into parent which must be a slice ptr.
 func InsertIntoSlice(parentSlice interface{}, value interface{}) error {
-	DbgPrint("InsertIntoSlice into parent type %T with value %v, type %T", parentSlice, ValueStrDebug(value), value)
+	if DebugLibraryEnabled() {
+		DbgPrint("InsertIntoSlice into parent type %T with value %v, type %T", parentSlice, ValueStrDebug(value), value)
+	}
```
wenovus, line 177: How does this help during compilation? Doesn't this mean we have two inline another function during compiler analysis? Do you have evidence that this saves resources during compilation?

ygot/render.go
```diff
@@ -599,11 +600,67 @@ func KeyValueAsString(v interface{}) (string, error) {
 	}
 
 	switch kv.Kind() {
-	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+	case reflect.Int:
+		if val, ok := v.(int); ok {
```
wenovus, line 604: Why was this done?

util/reflect.go
jayzhudev: Good question. From what was observed in profiling, the `DbgPrint` calls were hot paths while the default value of `debugLibrary` in `debug.go` was `false`. It's particularly easy to see the high usages in the flame graph when pretty printing is called inline (and with a large config tree on a resource constrained device). It doesn't seem like the compiler will optimize away these calls when `debugLibrary` is false, possibly because it's a variable that can be changed at runtime.

util/path.go
jayzhudev: You are right, this is not needed. I've removed this in the coming commit, thanks.

ygot/render.go
jayzhudev: This is again for performance, although it resulted in slightly more complex code. After profiling, `strconv.Itoa` and `strconv.FormatInt/FormatUint` are much more efficient than `fmt.Sprintf` so it'd be nice if we could to use them whenever possible in frequently called paths such as: `node.retrieveNodeList` (loop in recursion) -> `ygot.PathKeyFromStruct` -> `keyMapAsStrings` (loop) -> `KeyValueAsString`.

util/reflect.go
jayzhudev: BTW this will *not* save resources at compile time but will noticeably improve runtime performance. I loosely remember this was the first that got patched, and there was a ~75% runtime cut (measured in our integration scenario, YANG models and config data).

wenovus: My guess is this improves performance the same way glog/klog's `.V` works. This looks fine to me.

I'm wondering if you tried adding a wrapper function that does this instead of having to call an if every time?

So something like
```go
// DbgPrint prints v if the package global variable debugLibrary is set.                                                                                                                                                                                                                              
// v has the same format as Printf. A trailing newline is added to the output.                                                                                                                                                                                                                        
func DbgPrint(v ...interface{}) {                                                                                                                                                                                                                                                                     
        if debugLibrary {                                                                                                                                                                                                                                                                             
                dbgPrint(v...)                                                                                                                                                                                                                                                                        
        }                                                                                                                                                                                                                                                                                             
}                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                      
// dbgPrint prints v if the package global variable debugLibrary is set.                                                                                                                                                                                                                              
// v has the same format as Printf. A trailing newline is added to the output.                                                                                                                                                                                                                        
func dbgPrint(v ...interface{}) {                                                                                                                                                                                                                                                                     
        out := fmt.Sprintf(v[0].(string), v[1:]...)                                                                                                                                                                                                                                                   
        if len(out) > maxCharsPerLine {                                                                                                                                                                                                                                                               
                out = out[:maxCharsPerLine]                                                                                                                                                                                                                                                           
        }                                                                                                                                                                                                                                                                                             
        fmt.Println(globalIndent + out)                                                                                                                                                                                                                                                               
}                                                                                                                                                                                                                                                                                                     
```
so it's easier for the compiler to inline the first function and then hopefully optimize the rest away.

But if you didn't try this I think we can still go with the current approach. I don't see `DbgPrint` getting heavy usage in the future as glog/klog should be the preferred logging library.

ygot/render.go
wenovus: Thanks, this looks good then. A cleaner way to write this would be to call a type switch on `v` instead of switching on each reflect.Type. This way you can avoid re-writing `return fmt.Sprintf("%d", v), nil`.

util/reflect.go
jayzhudev: The wrapper function works similarly to the existing implementation before this PR. The performance issue is caused by passing the function return as an argument directly to the debug call (maybe I shouldn't have used the `inline` word, it's different than compiler inlining, sorry for the confusion). This results in the argument function call being executed anyway regardless of the `degubLibary` value:

```go
	DbgPrint("InsertIntoMap into parent type %T with key %v(%T) value \n%s\n (%T)",
		parentMap, ValueStrDebug(key), key, pretty.Sprint(value), value)
```

Here the `pretty.Sprint` is computed anyway and is quite expensive.

In `glog` they implement logging calls on a `boolean` type alias and it's an if-statement too. It looks nicer but again if a function return is used as an argument, the `glog` implementation won't prevent unnecessary computations of such functions too, because the computations happen before the if-statement:

```go
// Infof is equivalent to the global Infof function, guarded by the value of v.
// See the documentation of V for usage.
func (v Verbose) Infof(format string, args ...any) {
	if v {
		logf(1, logsink.Info, true, noStack, format, args...)
	}
}
```

ygot/render.go
jayzhudev: > Thanks, this looks good then. A cleaner way to write this would be to call a type switch on `v` instead of switching on each reflect.Type. This way you can avoid re-writing `return fmt.Sprintf("%d", v), nil`.

This is a good catch, I've made the change in the latest commit. Thanks!

util/reflect.go
wenovus: Thanks for the insight! This looks good.

ytypes/node.go
```diff
@@ -327,13 +327,15 @@ func retrieveNodeContainer(schema *yang.Entry, root interface{}, path *gpb.Path,
...
 		return nil, status.Errorf(codes.InvalidArgument, "unkeyed list can't be traversed, type %T, path %v", root, path)
 	case len(path.GetElem()) == 0:
 		return nil, status.Errorf(codes.InvalidArgument, "path length is 0, schema %v, root %v", schema, root)
-	case !util.IsValueMap(rv):
+	}
+
+	rv := reflect.ValueOf(root)
+	if !util.IsValueMap(rv) {
 		return nil, status.Errorf(codes.InvalidArgument, "root has type %T, expect map", root)
 	}
```
wenovus, line 340: can change the last case to
```go
case !util.IsValueMap(reflect.ValueOf(root)):
```

ytypes/util_schema.go
```diff
@@ -155,10 +155,16 @@ func directDescendantSchema(f reflect.StructField) (string, error) {
...
+	}
+
 	paths := strings.Split(pathAnnotation, "|")
 
 	for _, pth := range paths {
-		if len(strings.Split(pth, "/")) == 1 {
+		if !strings.Contains(pth, "/") {
 			return pth, nil
 		}
 	}
```
wenovus, line 170: Does this actually improve performance? Split and Contains both do a full scan of the string right? I'm ok with keeping this, just it does make the code slightly harder to read.

wenovus (approved):

ygot/render.go
wenovus: Looks good, thanks!

ytypes/node.go
jayzhudev: If I understand the context correctly, `rv`is used in the code block after the `switch` and `if` statements:

```go
	switch {
	case schema.Key == "":
		return nil, status.Errorf(codes.InvalidArgument, "unkeyed list can't be traversed, type %T, path %v", root, path)
	case len(path.GetElem()) == 0:
		return nil, status.Errorf(codes.InvalidArgument, "path length is 0, schema %v, root %v", schema, root)
	}

	rv := reflect.ValueOf(root)
	if !util.IsValueMap(rv) {
		return nil, status.Errorf(codes.InvalidArgument, "root has type %T, expect map", root)
	}

	var matches []*TreeNode

	listKeyT := rv.Type().Key()
	listElemT := rv.Type().Elem()
	for _, k := range rv.MapKeys() {
		listElemV := rv.MapIndex(k)

...
```

I think the thought process when this change was made was to reuse `rv` so `reflect.ValueOf` is not called twice.

ytypes/util_schema.go
jayzhudev: Here's the `genSplit` code. It also *must* pass through the entire string as oppose to `strings.Contains` can early stop at the first occurrence:

```go
// Generic split: splits after each instance of sep,
// including sepSave bytes of sep in the subarrays.
func genSplit(s, sep string, sepSave, n int) []string {
	if n == 0 {
		return nil
	}
	if sep == "" {
		return explode(s, n)
	}
	if n < 0 {
		n = Count(s, sep) + 1
	}

	if n > len(s)+1 {
		n = len(s) + 1
	}
	a := make([]string, n)
	n--
	i := 0
	for i < n {
		m := Index(s, sep)
		if m < 0 {
			break
		}
		a[i] = s[:m+sepSave]
		s = s[m+len(sep):]
		i++
	}
	a[i] = s
	return a[:i+1]
}
```

ytypes/node.go
wenovus: I see, thanks for pointing it out.

ytypes/util_schema.go
wenovus: I was considering the case where no separators "|" or "/" were present, since that is the only case where the early return does save time. In this case I think not much heap allocations need to be made for adding new path elements, but I agree it still saves that allocation.

wenovus:
@robshakir for any comments before merge. The coverage decreased because the the debugging statements inside the if statements are not covered. Since these are debugging statements I'm ok with them being uncovered.

PR #829, opened 2023-05-01
perf(node): reduce reflect FieldByName and Field calls

jayzhudev:
The `reflect` method calls are expensive to be put inside nested for-loops. The performance can be improved if the results of these calls are cached.

Here are some rough reference numbers from e2e testing sending 2000 leaf-list updates in a SetRequest:

| Version             | Time          |
| --------------- | ---------- |
| Before #829     | ~2,300 ms     | 
| **After** #829 | **~900 ms** |


ytypes/node.go
```diff
@@ -341,6 +341,27 @@ func retrieveNodeList(schema *yang.Entry, root interface{}, path, traversedPath
 
 	listKeyT := rv.Type().Key()
 	listElemT := rv.Type().Elem()
+
+	listKeyFieldNameMap := map[int]string{}
+	schemaKeyMap := map[string]string{}
+
+	// Reduce FieldByName calls in the loop and reduce Field calls and tag lookup calls in the loop.
+	if listKeyT.Kind() == reflect.Struct {
```
wenovus, line 349: util.IsTypeStruct(listKeyT)

ytypes/node.go
```diff
@@ -341,6 +341,27 @@ func retrieveNodeList(schema *yang.Entry, root interface{}, path, traversedPath
...
+		for i := 0; i < listKeyT.NumField(); i++ {
+			fieldName := listKeyT.Field(i).Name
+			if elem, ok := listElemT.Elem().FieldByName(fieldName); ok {
+				schemaKey, err := directDescendantSchema(elem)
+				if err != nil {
+					return nil, status.Errorf(codes.Unknown, "unable to get direct descendant schema name for %v: %v", schemaKey, err)
+				}
+
+				schemaKeyMap[fieldName] = schemaKey
+			}
```
wenovus, line 359: To keep the error statement close to where it was generated, you can return an error when ok=False like the original code
```go
if !ok {
				return nil, status.Errorf(codes.NotFound, "element struct type %v does not contain key field %s", listElemT, fieldName)
			}
```

ytypes/node.go
```diff
@@ -407,22 +428,17 @@ func retrieveNodeList(schema *yang.Entry, root interface{}, path, traversedPath
...
 			fieldValue := k.Field(i)
 			if !fieldValue.IsValid() {
 				return nil, status.Errorf(codes.InvalidArgument, "invalid field %s in %T", fieldName, k)
 			}
 
-			elemFieldT, ok := listElemT.Elem().FieldByName(fieldName)
+			schemaKey, ok := schemaKeyMap[fieldName]
 			if !ok {
 				return nil, status.Errorf(codes.NotFound, "element struct type %v does not contain key field %s", listElemT, fieldName)
 			}
```
wenovus, line 440: If you place this error statement above, then this error can be changed to say codes.Internal, and that it is expected that this field should always be populated.

wenovus (commented):
A couple nits, otherwise looks good, thanks!

ytypes/node.go
jayzhudev: These are good suggestions, thanks. Changes made in the latest commit.

wenovus (approved):

PR #830, opened 2023-05-01
perf(node): caching implementation for node operations

jayzhudev:
## What
The squashed changes are ported from our local fork after making performance improvement changes for our use case.

## Why
In one of our use cases of the `ygot` library, we found that there are significant performance benefits if we add a caching layer to the `node` calls. Most of the time spent on calling these methods, especially when the config tree is very large, is to perform tree traversal and find the nodes to operate on. This involves using relatively expensive calls to `reflect` methods and is mostly repetitive computation, if the configuration tree's structure and paths are relatively static but the values may change.

## How (high-level)
The idea to implement the caching layer is to be able to use the unique node paths to locate a node immediately once its address is cached. This provides a shortcut to the tree-traversal method and brings in performance gains. In our use case, this feature addition along with other performance patches we made in other PRs brought the computation time down to <= 10ms from multiple seconds, under pressure tests.

This does come with a cost of higher complexity where the cache operations have to be managed properly. It also resulted in changes we had to make in `reflect.go` to make sure that the node's address doesn't change when its value is modified.

In addition, we hope to make sure that the `ygot` library is stateless, and there's no breaking change to the *exported* APIs after adding the caching mechanism. As a result, the applications that use the `ytypes` library maintain the cache(s) on their own, and they can optionally pass in a pointer of the `node cache` to the call option struct and get a performance boost.

### JSON Marshaling Performance
When trying to squeeze the last bit of performance out, we decided to use `go-json` at a few places instead of `encoding/json`. There was a bit of hesitation to do so, but the performance boost was significant, especially on the hot paths.

### Procedure Shortcuts and `reflect` Calls
At the very last bit, but not least, we try to optimize the use of `reflect` as much as possible, and early return from continuing unnecessary heap allocations (GC pressure at scale) and computations. Both of which could be considered refactoring with performance gains so no breaking change was introduced. Such code changes can be found, for example, in `ygot/render.go/prependmodsJSON`.

## Summary of Change
- Add caching for high performance Sets/Gets. The CPU usage and performance of GetNode/SetNode calls are significantly reduced after this change.

- Improve JSON marshaling performance using go-json. This helps squeeze out the last bit of performance particularly for embedded devices.

- Shortcut expensive heap allocations. The repetitive heap allocations may increase GC pressure and increase CPU usage when being called frequently.

**Note**: tests need to be added and completed for `caching enabled` method calls.

### DIFF
--- a/ygot/struct_validation_map.go
+++ b/ygot/struct_validation_map.go
@@ -28,6 +28,9 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"sync"
+
+	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
 
 	"github.com/openconfig/gnmi/errlist"
 	"github.com/openconfig/ygot/util"
@@ -37,16 +40,27 @@ const (
 	// indentString represents the default indentation string used for
 	// JSON. Three spaces are used based on the legacy use of EmitJSON.
 	indentString string = "   "
+
+	// pathPoolBufSize is the allocation size of the mapPathsPool, this should
+	// not be very large since the size of mapPaths is determined by the number
+	// of "|" separators.
+	pathPoolBufSize = 16
 )
 
+// mapPathsPool helps reduce heap allocations.
+var mapPathsPool = sync.Pool{
+	New: func() interface{} {
+		return make([]*gnmiPath, pathPoolBufSize)
+	},
+}
+
 // structTagToLibPaths takes an input struct field as a reflect.Type, and determines
-// the set of validation library paths that it maps to. Returns the paths as a slice of
-// empty interface slices, or an error.
+// the set of validation library paths that it maps to.
 //
-// Note: the returned paths use a shallow copy of the parentPath.
-func structTagToLibPaths(f reflect.StructField, parentPath *gnmiPath, preferShadowPath bool) ([]*gnmiPath, error) {
+// Note: the mapPaths input is modified in-place.
+func structTagToLibPaths(mapPaths []*gnmiPath, f reflect.StructField, parentPath *gnmiPath, preferShadowPath bool) (int, error) {
 	if !parentPath.isValid() {
-		return nil, fmt.Errorf("invalid path format in parentPath (%v, %v)", parentPath.stringSlicePath == nil, parentPath.pathElemPath == nil)
+		return 0, fmt.Errorf("invalid path format in parentPath (%v, %v)", parentPath.stringSlicePath == nil, parentPath.pathElemPath == nil)
 	}
 
 	var pathAnnotation string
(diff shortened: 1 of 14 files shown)


ygot/struct_validation_map.go
```diff
@@ -37,16 +40,27 @@ const (
...
 	indentString string = "   "
+
+	// pathPoolBufSize is the allocation size of the mapPathsPool, this should
+	// not be very large since the size of mapPaths is determined by the number
+	// of "|" separators.
+	pathPoolBufSize = 16
 )
 
+// mapPathsPool helps reduce heap allocations.
+var mapPathsPool = sync.Pool{
```
wenovus, line 51: Can you add more details on why this helps avoid heap allocations?

ygot/struct_validation_map.go
wenovus, line 1: Can this file and render.go be split into a separate PR?

ytypes/node_cache.go
```diff
@@ -0,0 +1,274 @@
...
+func (*NodeCacheOpt) IsGetNodeOpt() {}
+
+// IsSetNodeOpt implements the SetNodeOpt interface.
+func (*NodeCacheOpt) IsSetNodeOpt() {}
+
+// IsUnmarshalOpt marks IgnoreExtraFields as a valid UnmarshalOpt.
+func (*NodeCacheOpt) IsUnmarshalOpt() {}
+
+// IsGetOrCreateNodeOpt implements the GetOrCreateNodeOpt interface.
+func (*NodeCacheOpt) IsGetOrCreateNodeOpt() {}
```
wenovus, line 38: Is this missing IsDelNodeOpt?

ytypes/gnmi_test.go
```diff
@@ -18,6 +18,7 @@ func TestUnmarshalSetRequest(t *testing.T) {
 		inUnmarshalOpts []UnmarshalOpt
 		want            ygot.GoStruct
 		wantErr         bool
+		resetNodeCache  bool
```
wenovus, line 21: Instead of reusing existing tests that weren't designed for the cached node option, I think it might make for more effective testing if there were a dedicated nodeCache test that mixes Get/GetOrCreate/Set/Delete and to see it working.

wenovus:
By the way thanks for posting this contribution! This looks like something that can benefit others and overall this implementation looks good.

jayzhudev:
> By the way thanks for posting this contribution! This looks like something that can benefit others and overall this implementation looks good.

Thanks for supporting this change! Let's get the other 3 PRs completed and I'll use some after work time to clean up this PR and add better tests 😄

ytypes/node_cache.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+// setNodeCache uses the cached information to set the node instead of traversalling the config tree.
+// This improves runtime performance of the library.
+func (c *NodeCache) set(path *gpb.Path, val interface{}, opts ...SetNodeOpt) (setComplete bool, err error) {
+	switch val.(type) {
+	case *gpb.TypedValue:
+	default:
+		// Only TypedValue is supported by the node cache for now.
+		return
```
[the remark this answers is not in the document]
jayzhudev: Thanks for the comment. I think this type switch is used to check the type of `val` and see if it's supported by the node cache. Maybe we don't want to return an error since that will turn into a `SetNode` error and block the `Set`. 

The node cache doesn't support all types that the ordinary `SetNode` supports so the intention is to let the `SetNode` continue (which will still modify the node, without the performance boost) if the `val` is not in supported type. Hope that makes sense.

ygot/struct_validation_map.go
jayzhudev: This was an experiment in this frequently called function about how much GC load `sync.Pool` can reduce, mostly using reference to the documentation in <https://go.dev/src/sync/pool.go>. In this use case I don't think there was significant GC pressure reduced (much less than the change from switching to `go-json`). If this is overly complex we can revert the use of `sync.Pool` for now. Thanks!

ytypes/node_cache.go
jayzhudev: I see, thanks for catching this, I think the `node.DeleteNode` should be fixed too.

jayzhudev: The initial cache implementation was a singleton in the `ygot` library. When it was refactored I think some bugs were introduced too. It was used by a PoC earlier but I plan to do some major work soon to improve the tests for the node cache to help us catch such issues easier.

ytypes/gnmi_test.go
jayzhudev: Agreed, will be working on that.

ygot/struct_validation_map.go
jayzhudev: Yes, will do, that's a good idea.

ytypes/node_cache.go
```diff
@@ -0,0 +1,274 @@
...
+// an integer type, or implement encoding.TextMarshaler. The map keys are sorted
+// and used as JSON object keys by applying the following rules, subject to the
+// UTF-8 coercion described for string values above:
+//
+//   - keys of any string type are used directly
+//
+//   - encoding.TextMarshalers are marshaled
+//
+//   - integer keys are converted to strings
+func uniquePathRepresentation(path *gpb.Path) (string, error) {
```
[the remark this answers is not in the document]
jayzhudev: Might need to compare the performance of these two if that matters. Previously `path.String` (`encoding/prototext/encode.go`) was used. Its performance was more than doubled after switching to `json.Marshal` (`json.Marshal` is deterministic so it's fine to use). After switching to `go-json` its was again largely improved.

Because computing the unique path representation is one of the most expensive, if not the most expensive parts of the node cache, the performance of this function affects the performance of the node cache.

ygot/struct_validation_map.go
jayzhudev: The changes you mentioned were split into #831.

PR #831, opened 2023-05-06
perf(ygot): improve perf for `render` and `struct_validation_map`

jayzhudev:
This is a split change from the combined PR of node cache implementation (#830).


ygot/render.go
```diff
@@ -1256,8 +1331,20 @@ func structJSON(s GoStruct, parentMod string, args jsonOutputConfig) (map[string
...
+		}
+
+		return fmt.Sprintf("%d", i)
+	case reflect.Uint64:
+		if val, ok := i.(uint64); ok {
+			return strconv.FormatUint(uint64(val), 10)
+		}
+
+		return fmt.Sprintf("%d", i)
 	}
```
wenovus, line 1348: As suggested below it would be nice if the cases here can be updated and the cases below removed in favour of this to benefit all users of this function.

ygot/render.go
```diff
@@ -1203,45 +1233,90 @@ func structJSON(s GoStruct, parentMod string, args jsonOutputConfig) (map[string
...
-				parent = parent[k].(map[string]interface{})
 			}
-			k, err := p.LastStringElem()
+
+			continue
+		}
+
+		func() {
+			mapPaths := mapPathsPool.Get().([]*gnmiPath)
+			defer mapPathsPool.Put(mapPaths)
```
wenovus, line 1272: It would be nice if this can be refactored into its own function now that there are more optimizations and code added.

jayzhudev: Thanks a lot for all the comments and notes. I agree that some of them seem more like trading complexity for small gains now (many are side changes added while hunting for big performance improvements for our particular use case). The use of `sync.Pool` here isn't justified by significant GC pressure reduction as well. I'll read through all of your comments again and I think we can probably revert the use of `sync.Pool`, that way this PR can be much simplified and I think it can focus on improving `writeIETFScalarJSON` based on your comment.

ygot/render.go
```diff
@@ -391,78 +409,83 @@ func findUpdatedLeaves(leaves map[*path]interface{}, s GoStruct, parent *gnmiPat
 			}
 		}
 
-		mapPaths, err := structTagToLibPaths(ftype, parent, false)
-		if err != nil {
-			errs.Add(fmt.Errorf("%v->%s: %v", parent, ftype.Name, err))
-			continue
-		}
+		func() {
```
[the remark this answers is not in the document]
jayzhudev: I think this was for convenience of putting back to the `sync.Pool` using `defer`. I personally don't prefer this approach as well, and the use of `sync.Pool` is not giving much performance gain in this case. I'm inclined to revert the use of `sync.Pool`.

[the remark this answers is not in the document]
jayzhudev: @wenovus Sorry for the delay. June was a very busy month on multiple tasks for me. I'll get back to this soon.

jayzhudev:
Since last time we visited this PR there have been lots of changes made to the master branch. I simplified this PR as it was mentioned previously that there are micro-optimizations we may not want to include and use of `sync.Pool` we may want to remove. This PR has now become a small change that brings in some cleanups and minor improvements to the use of `reflect` in `render.go`. Thank you!

PR #840, opened 2023-05-24
Marshal ordered maps into `atomic` gNMI notifications for `TogNMINotifications`

wenovus:
This is done by putting all values underneath an ordered map into a single, ordered list of Update values, which gets put into a separate Notification for return.

### DIFF
--- a/ygot/render.go
+++ b/ygot/render.go
@@ -95,6 +95,12 @@ func (p *path) String() string {
 	return fmt.Sprintf("%v", p.p.pathElemPath)
 }
 
+// pathval combines path and the value for the relative or absolute path.
+type pathval struct {
+	path *path
+	val  interface{}
+}
+
 // gnmiPath provides a wrapper for gNMI path types, particularly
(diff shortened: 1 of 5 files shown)


ygot/render.go
```diff
@@ -415,18 +462,97 @@ func findUpdatedLeaves(leaves map[*path]interface{}, s GoStruct, parent *gnmiPat
 				errs.Add(findUpdatedLeaves(leaves, goStruct, childPath))
 			}
 		case reflect.Ptr:
-			// Determine whether this is a pointer to a struct (another YANG container), or a leaf.
-			switch fval.Elem().Kind() {
-			case reflect.Struct:
-				goStruct, ok := fval.Interface().(GoStruct)
-				if !ok {
-					errs.Add(fmt.Errorf("%v: was not a valid GoStruct", mapPaths[0]))
+			if _, ok := fval.Interface().(GoOrderedList); ok {
```
DanG100, line 465: nit: handle GoOrderedList in a separate func? it adds significant complexity to this func

wenovus: Done.

PR #845, opened 2023-05-25
Support unmarshalling JSON ordered map

wenovus:


util/reflect.go
```diff
@@ -202,27 +198,43 @@ func InsertIntoMap(parentMap interface{}, key interface{}, value interface{}) er
...
+	}
+	return method, nil
+}
+
+// InsertIntoOrderedMap inserts value with key into parent which must be a
+// GoOrderedList interface.
+//
+// If the type is a GoOrderedList, then there must not exist an existing
+// element, and the key input will be unused.
+func InsertIntoOrderedMap(orderedMap goOrderedList, value interface{}) error {
```
DanG100, line 225: nit: AppendIntoOrderedMap, to me insert would take an index

DanG100 (approved):

util/reflect.go
wenovus: Done.

PR #850, opened 2023-05-31
Support `DeleteNode` for ordered maps

wenovus:


ytypes/node_exported_test.go
```diff
@@ -818,3 +849,230 @@ func TestSetNode(t *testing.T) {
 		}
 	}
 }
+
+func TestDeleteNode(t *testing.T) {
```
DanG100, line 853: does this need to be removed from node_test.go

wenovus: Updated names.

wenovus: These are just for ordered maps which need access to `ctestschema` but would cause a circular dependency if it's in the same package.

DanG100 (approved):

PR #851, opened 2023-05-31
Support unmarshalling an atomic `Notification` for ordered lists.

wenovus:
Involves supporting `DeleteNode` at the container above the list for ordered maps.

### DIFF
--- a/util/gnmi.go
+++ b/util/gnmi.go
@@ -42,6 +42,29 @@ func PathMatchesPrefix(path *gpb.Path, prefix []string) bool {
 	return true
 }
 
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+// PathPartiallyMatchesPrefix reports whether the path partially or wholly
+// matches the prefix.
+//
+// e.g. a/b partially matches a/b/c or a/b, but doesn't match a/c.
+func PathPartiallyMatchesPrefix(path *gpb.Path, prefix []string) bool {
+	for len(prefix) != 0 && prefix[len(prefix)-1] == "" {
+		prefix = prefix[:len(prefix)-1]
+	}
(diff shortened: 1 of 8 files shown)


util/gnmi.go
```diff
@@ -42,6 +42,29 @@ func PathMatchesPrefix(path *gpb.Path, prefix []string) bool {
...
+		return a
+	}
+	return b
+}
+
+// PathPartiallyMatchesPrefix reports whether the path partially or wholly
+// matches the prefix.
+//
+// e.g. a/b partially matches a/b/c or a/b, but doesn't match a/c.
+func PathPartiallyMatchesPrefix(path *gpb.Path, prefix []string) bool {
```
DanG100, line 56: nit: somehow include that no keys are compared in the function name

util/gnmi.go
```diff
@@ -42,6 +42,29 @@ func PathMatchesPrefix(path *gpb.Path, prefix []string) bool {
...
+	for len(prefix) != 0 && prefix[len(prefix)-1] == "" {
```
DanG100, line 57: when are there empty strings in path?

DanG100 (approved):

util/gnmi.go
wenovus: This is just to guard against a trailing slash when splitting. I wouldn't have written this myself -- it is kept in uniformity with the existing `PathMatchesPrefix`

util/gnmi.go
wenovus: Done.

util/gnmi.go
wenovus: This is just to guard against a trailing slash when splitting. I wouldn't have written this myself -- it is kept in uniformity with the existing `PathMatchesPrefix`

PR #860, opened 2023-06-05
Support `ForEachDataField`, `DeepCopy`, `Merge` for ordered maps.

wenovus:
For merge if there is any overlap it simply errors out -- I don't think there is a clear way of merging them if there is a partial overlap.

For overlaps where the second ordered list is a subset of the first, then I can see it as a valid merge -- however I'm opting for simplicity for now since this seems like a more advanced use case.

### DIFF
--- a/ygot/struct_validation_map.go
+++ b/ygot/struct_validation_map.go
@@ -950,7 +950,7 @@ func copyOrderedMap(dstField reflect.Value, srcOrderedMap GoOrderedList, accessP
 		return err
 	}
 	for _, k := range keys {
-		if _, ok := srcKeys[k]; ok {
+		if _, ok := srcKeys[k.Interface()]; ok {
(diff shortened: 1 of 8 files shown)


ygot/struct_validation_map.go
```diff
@@ -906,6 +912,73 @@ func validateMap(srcField, dstField reflect.Value) (*mapType, error) {
 	return &mapType{key: st.Key(), value: st.Elem()}, nil
 }
 
+// copyOrderedMap copies srcField into dstField. Both srcField and dstField are
```
DanG100, line 915: nit: `srcField` isn't a parameter so this comment is a little hard to follow

DanG100 (approved):

ygot/struct_validation_map.go
wenovus: Done.

PR #863, opened 2023-06-07
Added UnmarshalOpt to allow a best effort unmarshal

lgomez9:
Added an option to allow a failing unmarshal will now pass - instead of failing, it will simply log the errors and continue the unmarshaling process. This is so that we can generate diffs for the entirety of the SetRequest, even if there's a compliance error.


ytypes/unmarshal.go
```diff
@@ -29,6 +29,14 @@ type UnmarshalOpt interface {
 	IsUnmarshalOpt()
 }
 
+// BestEffortUnmarshal is an unmarshal option that 'ignores' errors while unmarshalling,
```
wenovus, line 32: ignores -> accumulates?

ytypes/unmarshal.go
```diff
@@ -29,6 +29,14 @@ type UnmarshalOpt interface {
...
+// and continues the unmarshaling process. A failing unmarshal will still fail, but it will
+// fail as a series of error reports.
```
wenovus, line 34: What kind of error type will be returned?

One simple way of doing the error report is gnmi's errlist package: https://pkg.go.dev/github.com/openconfig/gnmi/errlist

However for more detailed error reporting you might consider what is being used in ygnmi:
https://github.com/openconfig/ygnmi/blob/613186339658c020a54b31ebe8e751e94e10ab43/ygnmi/unmarshal.go#L65-L74

In fact ygnmi does a very similar operation of unmarshalling Notifications to a GoStruct, and I might be able to leverage your effort here if you choose this route. I have been planning to use UnmarshalNotifications in ygnmi to do this.

ytypes/unmarshal.go
```diff
@@ -29,6 +29,14 @@ type UnmarshalOpt interface {
...
+type BestEffortUnmarshal struct{}
```
wenovus, line 35: Perhaps make this unexported for now and export it after the functionality is ready in case we want to create another release before then?

ytypes/unmarshal.go
```diff
@@ -29,6 +29,40 @@ type UnmarshalOpt interface {
 	IsUnmarshalOpt()
 }
 
+// ComplianceErrors contains the compliance errors encountered from an Unmarshal operation.
+type ComplianceErrors struct {
+	// Generic errors for now, until we make a decision on what specific types of errors should
+	// be returned
+	Errors []error
```
wenovus, line 36: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences
Comments should begin with the name of the thing being described and end in a period:

ytypes/gnmi.go
```diff
@@ -62,9 +63,10 @@ func UnmarshalNotifications(schema *Schema, ns []*gpb.Notification, opts ...Unma
 //
 // If an error occurs during unmarshalling, schema.Root may already be
 // modified. A rollback is not performed.
-func UnmarshalSetRequest(schema *Schema, req *gpb.SetRequest, opts ...UnmarshalOpt) error {
+func UnmarshalSetRequest(schema *Schema, req *gpb.SetRequest, opts ...UnmarshalOpt) (*ComplianceErrors, error) {
```
wenovus, line 66: Let's add some tests for this flag.

ytypes/gnmi.go
```diff
@@ -87,10 +89,10 @@ func UnmarshalSetRequest(schema *Schema, req *gpb.SetRequest, opts ...UnmarshalO
 	if err := deletePaths(schema.SchemaTree[nodeName], node, prefix, req.Delete, preferShadowPath); err != nil {
```
wenovus, line 89: Should this apply to deletes as well?

ytypes/gnmi.go
```diff
@@ -35,7 +36,7 @@ import (
 //
 // If an error occurs during unmarshalling, schema.Root may already be
 // modified. A rollback is not performed.
-func UnmarshalNotifications(schema *Schema, ns []*gpb.Notification, opts ...UnmarshalOpt) error {
+func UnmarshalNotifications(schema *Schema, ns []*gpb.Notification, opts ...UnmarshalOpt) (*ComplianceErrors, error) {
```
wenovus, line 39: So this would be a backwards-incompatible change. I think it's better to make ComplianceErrors satisfy the error interface and return it instead. So essentially the flag controls whether this structure is returned.

wenovus (changes requested):

ytypes/unmarshal.go
wenovus: For this we've decided to return an error type that's to-be-filled.

ytypes/unmarshal.go
wenovus: To enable early usage we've decided to keep this exported and just have the MVP functionality of accumulating errors in an error report.

ytypes/unmarshal.go
```diff
@@ -29,6 +29,49 @@ type UnmarshalOpt interface {
...
+		return &ComplianceErrors{Errors: errs}
+	}
+
+	c.Errors = append(c.Errors, errs...)
+	return c
+}
+
+// BestEffortUnmarshal is an unmarshal option that accumulates errors while unmarshalling,
+// and continues the unmarshaling process. An unmarshal will no longer fail, but
+// will return an error report.
```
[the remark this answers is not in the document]
lgomez9: Ack.

ytypes/unmarshal.go
```diff
@@ -29,6 +30,49 @@ type UnmarshalOpt interface {
...
+	}
+
+	c.Errors = append(c.Errors, errs...)
+	return c
+}
+
+// BestEffortUnmarshal is an unmarshal option that accumulates errors while unmarshalling,
+// and continues the unmarshaling process. An unmarshal now return a ComplianceErrors struct,
+// instead of a single error.
+type bestEffortUnmarshal struct{}
```
wenovus, line 71: Since we have decided on an MVP, if you intend to use this prior to adding more functionality then feel free to make this exported.

ytypes/gnmi_test.go
```diff
@@ -496,6 +496,48 @@ func TestUnmarshalSetRequest(t *testing.T) {
 			}},
 		},
 		wantErr: true,
+	}, {
+		desc: "mix of an error and a non-error update with best-effort flag",
```
wenovus, line 500: For testing I recommend using a complete set-up but testing **two** errors in each of 1) delete 2) replace 3) update and then checking that all six errors are present. This way you get coverage of all six modifications in the handling logic.

wenovus (changes requested):
PTAL at the presubmit failures.

ytypes/unmarshal.go
lgomez9: Ack.

ytypes/gnmi_test.go
lgomez9: I think I understand - I'll upload a new version in a few, let me know if I've misunderstood.

ytypes/gnmi_test.go
```diff
@@ -504,6 +641,16 @@ func TestUnmarshalSetRequest(t *testing.T) {
 			if gotErr := err != nil; gotErr != tt.wantErr {
 				t.Fatalf("got error: %v, want: %v", err, tt.wantErr)
 			}
+			if gotErr := err != nil; gotErr && hasBestEffortUnmarshal(tt.inUnmarshalOpts) {
```
wenovus, line 644: nit: can just `if err != nil && hasBestEffort`

ytypes/gnmi_test.go
```diff
@@ -496,6 +498,141 @@ func TestUnmarshalSetRequest(t *testing.T) {
 			}},
 		},
 		wantErr: true,
+	}, {
+		desc: "mix of an error and a non-error update with best-effort flag",
```
wenovus, line 502: Actually I was thinking of putting all the errors into a single SetRequest. If you just add another test case with everything in these three cases or just merge all of these into one then this LGTM.

ytypes/gnmi_test.go
lgomez9: Makes way more sense.

ytypes/gnmi_test.go
lgomez9: Moving them all into one.

wenovus (approved):
Thanks!

ytypes/unmarshal.go
```diff
@@ -29,6 +30,49 @@ type UnmarshalOpt interface {
...
+		for _, e := range c.Errors {
+			b.WriteString("\n\t")
+			b.WriteString(e.Error())
+		}
+	} else {
+		b.WriteString(" None")
+	}
+	b.WriteString("\n")
+	return b.String()
+}
```
wenovus, line 57: Can you make it so that we have coverage for this? A good way is to use this: https://pkg.go.dev/github.com/openconfig/gnmi/errdiff#Substring and just check for the presence of "Noncompliance errors" whenever `hasBestEffortUnmarshal` returns true. I think it makes sense to just add this option for one other random passing test to get coverage for the nil case. Thanks!

wenovus (changes requested):
Missed reviewing the coverage, can you add coverage for one part of the code and the rest looks good.

ytypes/unmarshal.go
lgomez9: Is it okay to have the coverage in gnmi_test.go? Currently, the BestEffortUnmarshal only has an effect when used on the UnmarshalSetRequest function. I could also add functionality for this to the unmarshalGeneric function if you think that'd be a good idea.

wenovus: It's ok that `ytypes.Unmarshal` doesn't support this option. Actually it would be good to add logic there to error out if this option is provided saying it's not supported. It's somewhat involved to get it working there since it deals with JSON and there are a lot of places where errors can happen so I'm ok with not supporting it.

As for coverage of the `String()` method I'm fine with either having direct coverage in unmarshal_test.go or via Substring.

Oh yes, since this also affects UnmarshalNotifications can you add a test case for that as well? Thanks.

lgomez9: Added!

wenovus (approved):
Thank you, LGTM

lgomez9:
Sorry, thought I had run gofmt, but forgot to.

wenovus (approved):
Thanks! merged

PR #864, opened 2023-06-07
Add `DiffWithAtomic` to support creating Atomic Notifications.

wenovus:
When comparing two GoStructs where a YANG `ordered-by user` list is involved, it is not possible using a single Notification to represent the diff such that the Notification can do the following:
1. update a gNMI cache correctly with the atomic notification.
2. unmarshal the notification to the `orig` to get back the `modified` GoStruct.

This function achieves this functionality, sharing the underlying implementation with Diff, where a disclaimer has been added to the doc comment.

Other helpers have been created as well to be shared with `TogNMINotifications`.

Other changes:
- `TogNMINotifications` has been changed to follow the same convention of putting the non-atomic Notification as the first message instead of the last. The reason is that current code MAY be using `notifs[0]` and I don't want to break that (although it is technically a bug if they're not handling the atomic notifs), and this also seems slightly better from a usability perspective if someone wants to split the non-atomic notif from the atomic notifs.


ygot/diff.go
```diff
@@ -447,17 +477,156 @@ func (*DiffPathOpt) IsDiffOpt() {}
...
+//
+// A set of options for diff's behaviour, as specified by the supplied DiffOpts
+// can be used to modify the behaviour of the Diff function per the individual
+// option's specification.
+//
+// The returned gNMI Notification cannot be put on the wire unmodified, since
+// it does not specify a timestamp - and may not contain the absolute paths
+// to the fields specified if a GoStruct that does not represent the root of
+// a YANG schema tree is not supplied as original and modified.
+func DiffWithAtomic(original, modified GoStruct, opts ...DiffOpt) ([]*gnmipb.Notification, error) {
```
DanG100, line 527: (no action required): should deprecate Diff in favor of this func?

DanG100 (approved):

PR #870, opened 2023-06-09
Initial tests for uncompressed structs with ordered maps.

wenovus:


integration_tests/schemaops/utestschema/helper.go
```diff
@@ -0,0 +1,46 @@
...
+
+package utestschema
+
+import (
+	"testing"
+
+	"github.com/openconfig/ygot/ygot"
+)
+
+func GetOrderedMap(t *testing.T) *Ctestschema_OrderedLists_OrderedList_OrderedMap {
```
DanG100, line 23: godoc for funcs in this file
nit: GetSampleOrderedMap or something to indicate that this returns an ordered with by populated values

DanG100 (approved):

integration_tests/schemaops/utestschema/helper.go
wenovus: Done. I'll also move these to an internal package in a subsequent PR.

PR #872, opened 2023-06-09
Uncompressed ordered map tests in mostly `ygot` package.

wenovus:
* Simplified `UnmarshalSetRequest` to always unmarshal w/r to the root node. This allows atomic Notifications to be deleted correctly.


integration_tests/schemaops/utestschema/helper.go
```diff
@@ -50,3 +50,52 @@ func GetDeviceWithOrderedMap(t *testing.T) *Device {
 		},
 	}
 }
+
+func GetOrderedMap2(t *testing.T) *Ctestschema_OrderedLists_OrderedList_OrderedMap {
```
DanG100, line 54: godoc  
nit: i don't like "2" suffix, it's unclear what makes this func different from other func without looking at the code

wenovus: Thanks for the feedback, I've added a godoc. I think in this case it's ok to leave it like this rather than `GetOrderedMapWeeWoo` as the readers are mostly ygot maintainers and the code is fairly easy to understand. The "2" as a indicator for a different ordered map I think is sufficient. Let me know if you have other thoughts on this.

ygot/gostruct_exported_test.go
```diff
@@ -55,6 +56,34 @@ func TestPruneConfigFalseOrderedMap(t *testing.T) {
 				return om
 			}(),
 		},
+	}, {
+		desc:     "prune through ordered map",
```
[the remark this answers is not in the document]
wenovus: Done

PR #876, opened 2023-06-10
Fix test imports: we can't have exported test importing symbols from …

wenovus:
…unexported test files

### DIFF
--- a/util/reflect_exported_test.go
+++ b/util/reflect_exported_test.go
@@ -15,16 +15,42 @@
 package util_test
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/kylelemons/godebug/pretty"
 	"github.com/openconfig/gnmi/errdiff"
 	"github.com/openconfig/goyang/pkg/yang"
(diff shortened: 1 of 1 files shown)


util/reflect_exported_test.go
```diff
@@ -27,6 +29,30 @@ import (
...
+		fn, err := util.SchemaPaths(ni.StructField)
+		if err != nil {
+			errs = append(errs, err)
+		}
+		if l := len(fn); l != 1 {
+			errs = append(errs, fmt.Errorf("invalid schema path length %d for %v", l, ni.StructField.Name))
+		}
+
+		*outs += fmt.Sprintf("%s/%s : \n%s\n, ", util.ValueStr(ni.FieldKey.Interface()), fn[0][0], util.ValueStr(s))
+		return
```
DanG100, line 52: if you never change value of the var, than make make  a normal func.

example https://go.dev/play/p/rKHcBI8Hpo4

DanG100 (approved):

util/reflect_exported_test.go
wenovus: This is just copy-paste, let me fix this.

PR #884, opened 2023-06-28
Add generation flag for ordered maps

wenovus:
Flag is on by default to send a signal to downstream users to convert. The flag is used for transitioning existing code.

Tested generator manually. presubmit also has a test for it via integration_tests/schemaops and `go generate ./...`

* Added integration test
* Fixed a few typos


DanG100 (approved):

wenovus:
Thanks for the review!

PR #886, opened 2023-07-08
Update dependencies in line with dependabot.

robshakir:
```
 * (M) go.{mod,sum}
  - Update dependencies.
```

Obsoletes #817, #858, #859, #883, #885.


wenovus (approved):

wenovus:
Looks like we'll need to move to Go 1.19 for the new glog. It would be too early if we are keeping with the 3 releases rule.

PR #891, opened 2023-07-14
Support subset merge for ordered maps

wenovus:
when merging A <- B, and B is subset of A and the ordering of elements don't contradict, allow merging to occur.


DanG100 (approved):

wenovus:
Thanks for the review!

PR #894, opened 2023-07-21
Change internal JSON Generation for OrderedMaps to be backwards-compatible

wenovus:
I did not consider JSON internal format for OrderedMaps and was generating it the same way as RFC7951.

However, Internal JSON format for YANG lists uses a JSON object/map type instead of JSON list to represent a map.

Doing this destroys the ordering and defeats the purpose of `ordered-by user` lists but it's more important to keep backwards-compatibility for this legacy JSON format.

Also refactor code.

### DIFF
--- a/ygot/render.go
+++ b/ygot/render.go
@@ -19,13 +19,13 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
-	"sort"
 	"strings"
 
 	"github.com/openconfig/gnmi/errlist"
 	"github.com/openconfig/gnmi/value"
 	"github.com/openconfig/ygot/internal/yreflect"
 	"github.com/openconfig/ygot/util"
+	"golang.org/x/exp/slices"
 	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/proto"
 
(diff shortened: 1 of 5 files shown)


wenovus:
> should this behavior be controlled by a flag?

I think this is not necessary because it preserves existing behaviour and AFAIK there is no user need for internal JSON format desiring the ordering property to be preserved. If there is a requirement then we can take a look.

DanG100 (approved):

DanG100:
> > should this behavior be controlled by a flag?
> 
> I think this is not necessary because it preserves existing behaviour and AFAIK there is no user need for internal JSON format desiring the ordering property to be preserved. If there is a requirement then we can take a look.

Should there be a TODO to remove this internal JSON? especially if it's not generated "incorrect" info

wenovus:
> > > should this behavior be controlled by a flag?
> > 
> > 
> > I think this is not necessary because it preserves existing behaviour and AFAIK there is no user need for internal JSON format desiring the ordering property to be preserved. If there is a requirement then we can take a look.
> 
> Should there be a TODO to remove this internal JSON? especially if it's not generated "incorrect" info

I think this might be reasonable. @robshakir do you think it's reasonable to mark as "deprecated" the `ygot.Internal` `JSONFormat`? The code comments mention that it is the format used by `pyangbind` so I'm not positive whether it should be marked as such, but it seems otherwise RFC7951 is becoming the default option rather than the internal format:

```diff
 const (
        // Internal is the custom JSON format that is output by the validation library, and
-       // by pyangbind. It is loosely specified - but is the default used by generator developers.
+       // by pyangbind. It is loosely specified - but is used by some generator developers.
+       //
+       // Deprecated: Prefer RFC7951 to preserve list ordering, among other
+       // advantages.
        Internal JSONFormat = iota
```

wenovus:
I'll merge this now, but depending on Rob's comments I might add a deprecation notice.

PR #895, opened 2023-07-25
Make Schema Tree public

Ankur19:
I would like to build the `SchemaTree` using yang Entries and then use the `ResolveLeafRefTarget` method to calculate the actual data type of a `leafref`.

### DIFF
--- a/ygen/codegen.go
+++ b/ygen/codegen.go
@@ -195,7 +195,7 @@ type mappedYANGDefinitions struct {
 	enumEntries map[string]*yang.Entry
 	// schematree is a copy of the YANG schema tree, containing only leaf
 	// entries, such that schema paths can be referenced.
-	schematree *schemaTree
+	schematree *SchemaTree
 	// modules is the set of parsed YANG modules that are being processed as part of the
 	// code generatio, expressed as a slice of yang.Entry pointers.
 	modules []*yang.Entry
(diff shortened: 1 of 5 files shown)


Ankur19:
@wenovus Please review.

wenovus:
@Ankur19 Does https://github.com/openconfig/goyang/blob/1d9b70a64cbe8e0ba42f95ab5c4e1ea81ac17c19/pkg/yang/entry.go#L1310 work? I'm assuming you're working with goyang structs so if the use case is not ygot-specific then goyang should be satisfying the need.

Ankur19:
> @Ankur19 Does https://github.com/openconfig/goyang/blob/1d9b70a64cbe8e0ba42f95ab5c4e1ea81ac17c19/pkg/yang/entry.go#L1310 work? I'm assuming you're working with goyang structs so if the use case is not ygot-specific then goyang should be satisfying the need.

Let me check. Thanks for the pointer.

Ankur19:
> Exporting this functionality looks fine to me since it provides handling that goyang's Entry.Find doesn't have.
> 
> Can you put it in a separate package called `yangschema`? Then we can rename the type to `yangschema.Tree`.

Done!

yangschema/yangschema.go
```diff
@@ -24,19 +24,19 @@ import (
...
 
-// buildSchemaTree maps a set of yang.Entry pointers into a ctree structure.
+// BuildSchemaTree maps a set of yang.Entry pointers into a ctree structure.
 // Only leaf or leaflist values are mapped, since these are the only entities
 // that can be referenced by XPATH expressions within a YANG schema.
 // It returns an error if there is duplication within the set of entries. The
 // paths that are used within the schema are represented as a slice of strings.
-func buildSchemaTree(entries []*yang.Entry) (*schemaTree, error) {
-	t := &schemaTree{}
+func BuildSchemaTree(entries []*yang.Entry) (*Tree, error) {
```
wenovus, line 38: nit: Change to `BuildTree` for conciseness.

wenovus (approved):

yangschema/yangschema.go
wenovus: Thanks, I'll merge this soon and cut a new release.

PR #914, opened 2023-09-22
Support nested containers in gNMI->Protobuf unmarshalling.

robshakir:
```
* (M) protomap/{proto,proto_test.go}
   - Support nested messages when unmarshalling protobufs, previously
     such messages did not have their contents mapped.
 * (M) protomap/integration_tests/integration_test.go
   - Add a testcase for gRIBI's real protobufs to ensure that
     unmarshalling is covered.
```

Currently, this leaves repeated fields as unsupported. To be added in a
subsequent CL.


protomap/proto.go
```diff
@@ -725,6 +787,17 @@ func makeWrapper(msg protoreflect.Message, fd protoreflect.FieldDescriptor, val
 	}
 }
 
+// isWrapper returns true if the field fd of the message msg is a ywrapper protobuf type.
+func isWrapper(msg protoreflect.Message, fd protoreflect.FieldDescriptor) bool {
+	newV := msg.NewField(fd)
+	switch newV.Message().Interface().(type) {
+	case *wpb.StringValue, *wpb.UintValue, *wpb.BytesValue, *wpb.BoolValue:
```
wenovus, line 794: Are Decimal64Value, IntValue missing?

Another idea to potentially make this less error-prone is
```go
newV.Message().FullName() matches "ywrapper.\w+Value"
```
But given ywrapper changes so infrequently I'm also fine with the current.

protomap/proto.go
```diff
@@ -536,6 +536,10 @@ func ProtoFromPaths(p proto.Message, vals map[*gpb.Path]interface{}, opt ...Unma
 		return fmt.Errorf("invalid protobuf message prefix supplied in options, %v", err)
 	}
 
+	return protoFromPathsInternal(p, vals, valPrefix, protoPrefix, hasIgnoreExtraPaths(opt))
+}
+
+func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix, protoPrefix *gpb.Path, ignoreExtras bool) error {
 	schemaPath := func(p *gpb.Path) *gpb.Path {
```
wenovus, line 543: Factor this out as its own helper function since it's not dependent on any function-scope variables?

protomap/proto.go
```diff
@@ -544,90 +548,148 @@ func ProtoFromPaths(p proto.Message, vals map[*gpb.Path]interface{}, opt ...Unma
...
 	}
 
-	// directCh is a map between the absolute schema path for a particular value, and
-	// the value specified.
-	directCh := map[*gpb.Path]interface{}{}
-	for p, v := range vals {
-		absPath := &gpb.Path{
-			Elem: append(append([]*gpb.PathElem{}, schemaPath(valPrefix).Elem...), p.Elem...),
-		}
+	findChildren := func(vals map[*gpb.Path]any, valPrefix *gpb.Path, protoPrefix *gpb.Path, directOnly, mustBeChildren bool) (map[*gpb.Path]any, error) {
```
wenovus, line 551: same here: suggestion to factor this out since it has a full set of parameters already specified.

protomap/proto.go
```diff
@@ -544,90 +548,148 @@ func ProtoFromPaths(p proto.Message, vals map[*gpb.Path]interface{}, opt ...Unma
...
+						directCh[pp] = v
+					}
+				}
+			case false:
 				directCh[pp] = v
 			}
 		}
+		return directCh, nil
+	}
+	// It is an error if we get called with values that aren't at least indirect children at this point.
```
wenovus, line 589: What does this comment mean?

protomap/proto.go
```diff
@@ -544,90 +548,148 @@ func ProtoFromPaths(p proto.Message, vals map[*gpb.Path]interface{}, opt ...Unma
...
+			switch {
+			case fd.IsList():
+				// TODO(robjs): Support mapping these fields -- currently we silently drop them for backwards compatibility.
+			case fd.IsMap():
+				rangeErr = fmt.Errorf("map fields are not supported in mapped protobufs at field %s", fd.FullName())
+				return false
+			default:
+				if isWrapper(m, fd) {
+					return true
+				}
```
wenovus, line 662:
```go
case isWrapper(m, fd):
   return true
default:
   ...
```

wenovus (approved):

protomap/proto.go
wenovus: I see it's there in the next PR

protomap/proto.go
wenovus: I see it's there in the next PR

protomap/proto.go
robshakir: ACK, yes, apologies. I was thinking to do the same thing once I realised it became fully parameterised.

protomap/proto.go
robshakir: They are missing -- will add them.

I thought about doing something similar, but I was wondering whether that actually is less detectable that this code needs to change if ywrapper did change in some backwards incompatible manner. I'll prefer to leave it as it as the moment (with the type additions), but we can re-examine.

protomap/proto.go
robshakir: This is just clarifying the reason for why we set the `mustBeChildren` argument to `true` here. When we are finding the children of this protobuf, we are either:

 * on the initial call -- where we expect that all paths that are supplied must be children of the root protobuf that is supplied.
 * on a recursive call to this function, at which point we have filtered the paths to find the children of the child of the root that we recursed into -- such that if we received a path that was outside of this subtree, it's an error earlier in the recursion.

I'll make the comment clearer as it doesn't quite say this :)

protomap/proto.go
robshakir: Done.

protomap/proto.go
robshakir: Done.

protomap/proto.go
robshakir: Done, thanks.

PR #915, opened 2023-09-22
Support mapping of protobuf lists.

robshakir:
This PR adds support for mapping protobuf `repeated` fields that represent YANG lists into protobuf messages. This was an unimplemented gap previously.

The large change is due to changes in the testdata protobuf that is used for mapping.

There's a little bit of a `DIFFBASE` SNAFU here, where I pushed the initial PR from a fork, and this one is in `openconfig/ygot`. The plan will be to merge #914 and then rebase this PR onto HEAD of `master`.


protomap/proto.go
```diff
@@ -652,7 +662,23 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
 		if fd.Kind() == protoreflect.MessageKind {
 			switch {
 			case fd.IsList():
-				// TODO(robjs): Support mapping these fields -- currently we silently drop them for backwards compatibility.
+				leaflist, leaflistunion, err := annotatedYANGFieldInfo(fd)
+				if err != nil {
+					rangeErr = fmt.Errorf("cannot extract field information for %s, %v", fd.FullName(), err)
+				}
+				switch {
+				case leaflist == true, leaflistunion == true:
```
wenovus, line 670: Drop the == true

protomap/proto.go
```diff
@@ -700,6 +726,137 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
...
+		// to avoid comparing the keys.
+		if !util.PathMatchesPathElemPrefix(schemaPath(absPath), schemaPath(fieldPath)) {
+			continue
+		}
+		// The key of the list is in the last element of the absolute path in the values map (the values
+		// map MUST contain data tree paths, since it is telling us list values to unmarshal).
+		k := absPath.Elem[len(fieldPath.Elem)-1]
+		// If the last element doesn't have a key, then we have not correctly found the list.
+		if k.Key == nil {
+			return protoreflect.Value{}, fmt.Errorf("invalid list schema field path does not have a list at the end %s", fieldPath)
```
wenovus, line 754: reword: invalid list data field path: does not have key values populated.

Also should the check be len(k.Key) == 0?

Another question is for unkeyed lists, since they can't be marshalled to a scalar, we can choose to not support them for now? If so, add a test on unkeyed lists that checks for an error.

protomap/proto.go
```diff
@@ -700,6 +726,137 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
...
+	case protoreflect.Uint64Kind:
+		v, err := strconv.ParseUint(val, 10, 64)
+		if err != nil {
+			return protoreflect.Value{}, fmt.Errorf("invalid uint64 value %v, err: %v", val, err)
+		}
+		return protoreflect.ValueOfUint64(v), nil
+	case protoreflect.StringKind:
+		return protoreflect.ValueOfString(val), nil
+	default:
+		return protoreflect.Value{}, fmt.Errorf("invalid kind %v", fd.Kind())
```
wenovus, line 856: Could there be other types? If so change the error message to say "unsupported"?

protomap/proto.go
```diff
@@ -700,6 +726,137 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
...
+		if k.Key == nil {
+			return protoreflect.Value{}, fmt.Errorf("invalid list schema field path does not have a list at the end %s", fieldPath)
+		}
+
+		// Find the parts of the path that are not the list -- we assume that this is 2 elements since
+		// we are in a compressed schema.
+		// TODO(robjs): Currently, this may report incorrectly in an uncompressed schema, but we don't have
+		// a signal to indicate this. One needs to be added to the generated protobufs.
+		keyPath := &gpb.Path{
+			Elem: append(protoPrefix.Elem, absPath.Elem[len(protoPrefix.Elem):len(protoPrefix.Elem)+2]...),
```
wenovus, line 762: suggestion to append from an empty slice.

I'm concerned this might not always work because if `append` here doesn't always allocate a new array then there might be conflicts between different iterations of this.

protomap/proto.go
```diff
@@ -700,6 +726,137 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
...
+	// We need to identify the keys that are within the list, as well as the data tree paths
+	// that they correspond to. We walk through the supplied values to determine which to process.
+	for p := range vals {
+		// Make the paths within the vals map absolute according to the supplied valPrefix.
+		absPath := &gpb.Path{
+			Elem: append(append([]*gpb.PathElem{}, valPrefix.Elem...), p.Elem...),
+		}
+		// Since the fieldPath is a schema path, then we need to compare just schema paths
+		// to avoid comparing the keys.
+		if !util.PathMatchesPathElemPrefix(schemaPath(absPath), schemaPath(fieldPath)) {
```
wenovus, line 746: Should this only populate fields when the ancestor keys of `fieldPath` also match with `absPath`? Is it just the last element's keys that need to be removed for this check?

protomap/proto.go
```diff
@@ -700,6 +726,137 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
...
+
+		// Find the parts of the path that are not the list -- we assume that this is 2 elements since
+		// we are in a compressed schema.
+		// TODO(robjs): Currently, this may report incorrectly in an uncompressed schema, but we don't have
+		// a signal to indicate this. One needs to be added to the generated protobufs.
+		keyPath := &gpb.Path{
+			Elem: append(protoPrefix.Elem, absPath.Elem[len(protoPrefix.Elem):len(protoPrefix.Elem)+2]...),
+		}
+		keys = append(keys, k.Key)
+		keyPaths = append(keyPaths, keyPath)
```
wenovus, line 765: I prefer to do the `seen` duplicate checking here since it would optimize on space a little better.

protomap/proto.go
```diff
@@ -700,6 +726,137 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
...
+					retErr = err
+					return false
+				}
+
+				var keyName string
+				for _, p := range paths {
+					keyName = p.Elem[len(p.Elem)-1].Name
+					break
+				}
+				pv, err := listKeyAsProtoValue(fd, key[keyName])
```
wenovus, line 824: check and test the case that the key doesn't exist in the path. Alternatively, and I think it's better, to check whether the input paths have all keys supplied in the previous key-gathering loop.

wenovus: To clarify this comment is referring to the case where `keyName` is not in `key`. This can happen if the list of keys in the path is not complete. If this is checked in the previous key-gathering loop then it would be safe to assume that `key[keyName]` always returns a valid value.

protomap/proto.go
robshakir: Done, good point :)

protomap/proto.go
robshakir: The protobuf translation itself doesn't handle unkeyed lists -- so they are never expected to show up here (it means that someone has created some proto with the annotations that they are somehow expecting to work, which seems unlikely).

Reworded the error and updated the check. I'm not sure how to add a test for the unkeyed list case, because we simply don't support them in the input proto, what does the test case that you're proposing look like? (We do not test this function with synthetic inputs because it's very difficult to create the right structures inside the protoreflect library that is a dependency).

protomap/proto.go
robshakir: Done.

protomap/proto.go
robshakir: Good point -- I had forgotten that there are cases where `append` will not allocate a new array. I've made the change.

protomap/proto.go
robshakir: In the case that we're iterating down multiple lists, we know that `vals` can only contain the subtree that we are in, at the point that we get to `a/list-one[key=a]/b/list-two[key=b]` then the only possible key of `list-one` must be `a` because otherwise we're in a different subtree -- so it seemed safe to just strip all keys here.

protomap/proto.go
robshakir: ACK, done.

protomap/proto.go
robshakir: It's not clear that we can check it beforehand, because we only know what keys there are in the list by the fields in the proto -- but the observation that we were not checking whether the value was actually specified is correct. I added a check for the case that the key is not specified in the list, and the case that we receive a key that we don't map to a field.

robshakir:
Thanks for the review, pushed an updated commit to address the comments.

protomap/proto.go
wenovus: Not handling it sounds good to me. I didn't consider that it's not supported by the input proto. I'm happy with handling that when and if this becomes a use case.

protomap/proto.go
robshakir: No problem!

robshakir:
Upstream conflicts fixed. @wenovus - PTAL when you get a chance, thanks!

protomap/proto.go
```diff
@@ -704,6 +729,153 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
...
+
+				var keyName string
+				for _, p := range paths {
+					keyName = p.Elem[len(p.Elem)-1].Name
+					break
+				}
+				if key[keyName] == "" {
+					retErr = fmt.Errorf("field %s, missing key %s, got keys: %v", fd.FullName(), keyName, key)
+					return false
+				}
```
wenovus, line 830: Should this be checking `ok` rather than "" since, perhaps, "" could be a valid key in rare situations?

wenovus (approved):
Sorry for the delay, LGTM!

Just one minor question.

protomap/proto.go
robshakir: Good catch, yes, I guess unlikely that this is the case, but possible. Fixed.

robshakir:
Thanks very much for the swift reviews Wen :-)

PR #924, opened 2023-10-25
Move to standard slices package

wenovus:
It looks like bazel is having trouble with this when I try to convert lemming to Go 1.21.


wenovus:
Abandoning since only supported on >=Go1.21

PR #926, opened 2023-11-04
Add support leaf-lists in paths->protobuf.

robshakir:
```
 * (M) protomap/integration_tests/integration_test.go
  - Add test case for a failure that was causing a panic,
    where a leaf-list of union values was being handed to
    PathsToProto. Currently not yet implemented, but test
    checks error handling works as expected.
 * (M) protomap/proto(_test)?.go
  - Add support and testing for mapping of leaf-lists of
    YANG inbuilt types to protobufs.
```

### DIFF
--- a/protomap/proto.go
+++ b/protomap/proto.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 
+	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/types/descriptorpb"
@@ -618,6 +619,11 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix, 
 			return false
 		}
 
+		leaflist, leaflistunion, err := annotatedYANGFieldInfo(fd)
+		if err != nil {
+			rangeErr = fmt.Errorf("cannot extract field information for %s, %v", fd.FullName(), err)
+		}
+
 		if len(directCh) != 0 {
 			for _, ap := range annotatedPath {
 				trimmedPrefix := schemaPath(protoPrefix)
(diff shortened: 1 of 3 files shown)


robshakir:
Sorry for the PR length -- necessary to add test coverage for all edge cases.

protomap/proto.go
```diff
@@ -920,6 +939,12 @@ func hasValuePathPrefix(opts []UnmapOpt) (*gpb.Path, error) {
 // type of payload is provided for the message. The second, boolean, return argument specifies whether
 // the message provided was a known wrapper type.
 func makeWrapper(msg protoreflect.Message, fd protoreflect.FieldDescriptor, val interface{}) (protoreflect.Message, bool, error) {
+	// If this field was a repeated then it could be a typed value -- but this is handled
+	// separately, thus we simply return false here.
+	if fd.IsList() {
+		return nil, false, nil
+	}
+
```
wenovus, line 947: Since makeWrapper is only called in once place, this means that this is triggered when,
1. the proto field is `repeated`.
2. YANG field is not a `leaf-list`.

I believe we must then be dealing with a YANG list. In this case, `makeWrapper` already handles this case, so I think we can remove this if statement.

protomap/proto.go
```diff
@@ -920,6 +939,12 @@ func hasValuePathPrefix(opts []UnmapOpt) (*gpb.Path, error) {
 // type of payload is provided for the message. The second, boolean, return argument specifies whether
 // the message provided was a known wrapper type.
 func makeWrapper(msg protoreflect.Message, fd protoreflect.FieldDescriptor, val interface{}) (protoreflect.Message, bool, error) {
```
wenovus, line 941: It looks like this function currently doesn't handle `wpb.IntValue` and `wpb.BoolValue`. Add a TODO?

wenovus (approved):
Thanks for all the test cases for the great coverage.

protomap/proto.go
robshakir: Well spotted, yes, we can -- previously, this was causing a panic because we got to the stage where we called `makeWrapper` without those checks (the leaf-list + leaf-list of unions one). I thought about removing it, but because the consequences are a panic then it seemed like it might be better to be defensive just here. WDYT?

protomap/proto.go
robshakir: Done.

protomap/proto.go
wenovus: Where was it panicking before? I'm ok either way. I tend to let the error happen in cases like these in order to test my understanding of the code, since if the code isn't behaving as I expect then it might be better than a silent error, but I've also ran into situations where that just took unnecessary debugging time later because the defensive code would've been correct.

robshakir: Maybe the most defensive thing to do would be to make it return an error now this should be handled properly. It was throwing the panic at 959 since you can't call `.Message()` on a `repeated` field.

I'll make it return an error -- which we can't easily exercise, but it will avoid the panic whilst surfacing this.

wenovus: LGTM, thanks.

PR #927, opened 2023-11-04
Add support for mapping leaf-lists of unions to protobufs.

robshakir:
```
* (M) integration_tests/integration_test.go
  - Reflect the fact that the gRIBI integration test cases is now
    implemented.
 * (M) protomap/proto.go
  - Add support for mapping both []any and gNMI TypedValue messages
    to fields within a protobuf from input gNMI paths.
 * (M) protomap/testdata/...
  - Additional fields in test protobufs.
```


protomap/proto.go
```diff
@@ -999,6 +1004,108 @@ func isWrapper(msg protoreflect.Message, fd protoreflect.FieldDescriptor) bool {
...
+				return true
+			})
+			if retErr != nil {
+				return protoreflect.ValueOf(nil), retErr
+			}
+		}
+		return newV, nil
+	}
+	var retErr error
+	for i := 0; i < inputVal.Len(); i++ {
```
wenovus, line 1064: nit: put in default case

wenovus (approved):

protomap/proto_test.go
```diff
@@ -1199,23 +1199,116 @@ func TestProtoFromPaths(t *testing.T) {
...
+		},
+		wantProto: &epb.ExampleMessage{
+			LeaflistUnion: []*epb.ExampleUnion{{
+				Str: "hello",
+			}, {
+				Str: "world",
+			}, {
+				Uint: 1,
+			}},
+		},
```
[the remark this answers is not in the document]
robshakir: Good catch -- since this actually highlighted the issue you mentioned below was a problem. Done :-)

protomap/proto.go
robshakir: Done. Added test coverage too.

PR #929, opened 2023-11-07
ygot.FormatDiff

wenovus:
Sample output:
```
deleted: /floatval
new/updated /int-val: int_val:10
```
```
no diff
```

### DIFF
--- a/ygot/diff.go
+++ b/ygot/diff.go
@@ -495,7 +495,7 @@ func Diff(original, modified GoStruct, opts ...DiffOpt) (*gnmipb.Notification, e
 // versions of the program.
 func FormatDiff(n *gnmipb.Notification) string {
 	if n == nil {
-		return "<nil> notification"
+		return "<nil> Notification"
 	}
 	var build strings.Builder
 	for _, d := range n.Delete {
--- a/ygot/diff.go
+++ b/ygot/diff.go
@@ -24,6 +24,7 @@ import (
 	"github.com/openconfig/gnmi/errlist"
 	"github.com/openconfig/ygot/internal/yreflect"
 	"github.com/openconfig/ygot/util"
+	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/proto"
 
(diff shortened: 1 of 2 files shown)


ygot/diff.go
```diff
@@ -488,6 +489,41 @@ func Diff(original, modified GoStruct, opts ...DiffOpt) (*gnmipb.Notification, e
 	}
 }
 
+// FormatDiff formats the output of ygot.Diff as a multiline string. This
+// function is only intended for human consumption and ignores errors. Do not
+// depend on the output being stable. It may change over time across different
+// versions of the program.
+func FormatDiff(n *gnmipb.Notification) string {
```
robshakir, line 496: I wonder whether this should go in `testutil` -- given that it seems like it might be a testing-only function, or do we expect production code to maybe use this for errors?

ygot/diff_test.go
```diff
@@ -1576,6 +1576,9 @@ func TestDiff(t *testing.T) {
 		t.Run(tt.desc+"Diff", func(t *testing.T) {
 			got, err := Diff(tt.inOrig, tt.inMod, tt.inOpts...)
 			testDiffSingleNotif(t, "Diff", got, err)
+			if diffout := FormatDiff(got); diffout == "" {
```
robshakir, line 1579: Shall we add one test case that shows that the string is what we expect it to be? I know there is no contract around it, but it seems like it might be a reasonable example?

robshakir (approved):

ygot/diff.go
wenovus: I think this makes it more discoverable since autocompleters might pick it up (mine does). Otherwise it would be hidden. I'm using the pattern from prototext.Format, which lives alongside prototext.Marshal/Unmarshal.

ygot/diff_test.go
wenovus: Done.

dplore:
Thanks for this Wen!

ygot/diff.go
robshakir: SGTM -- I agree with the discoverability argument.

PR #931, opened 2023-11-11
Support skipping any-xml using -ignore_unsupported.

wenovus:
Currently it fails because containers are being detected via the `default` case, so they're being classified as containers.


ejbrever (approved):
LGTM, thanks!

PR #947, opened 2024-01-10
Fix `Marshal7951` for direct union leaf calls.

wenovus:
When `jsonValue` is called using the output of `reflect.ValueOf()` instead of `reflect.StructField`, any interface type is lost through re-packing to the empty interface (any). This means the `reflect.Kind` of a union field is no longer the union type, but instead its underlying concrete type. The current code doesn't handle this case, leading to runtime errors.

This handling code is now added, with a couple more fixes related to `empty` types.

-----
Tested manually that the following ygnmi call is no longer affected by the original error:
```go
sp := gnmi.OC().NetworkInstance("DEFAULT").Protocol(oc.PolicyTypes_INSTALL_PROTOCOL_TYPE_STATIC, "STATIC")
ygnmi.Replace(context.Background(), c, sp.Static("1.1.1.1/32").SetTag().Config(), oc.NetworkInstance_Protocol_Static_SetTag_Union(oc.UnionUint32(42)))
```

### DIFF
--- a/testutil/types.go
+++ b/testutil/types.go
@@ -98,6 +98,7 @@ func (Binary) Is_UnionLeafTypeSimple()      {}
 func (UnionString) IsExampleUnion()       {}
 func (UnionFloat64) IsExampleUnion()      {}
 func (UnionInt64) IsExampleUnion()        {}
+func (UnionUint32) IsExampleUnion()       {}
 func (UnionBool) IsExampleUnion()         {}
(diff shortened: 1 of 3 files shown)


Ankur19 (approved):
Thanks for fixing this Wen!

wenovus:
> Thanks for fixing this Wen!

Thanks for the review!

PR #968, opened 2024-04-18
Remove support for go 1.18.

robshakir:
```
* (M) .github/workflows/go.yml
   - Remove testing support for go 1.18 -- these tests are failing
     based on dependencies. Our contract here is to support -3
     versions, so this is in-line with expecations of our consumers.
```

Fixes the failing CI @ HEAD.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: ['1.18', '1.19', '1.20', '1.x']
+        go: ['1.19', '1.20', '1.21', '1.22', '1.x']
 
     steps:
       - name: Install protobuf


.github/workflows/go.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: ['1.18', '1.19', '1.20', '1.x']
+        go: ['1.19', '1.20', '1.21', '1.22', '1.x']
```
wenovus, line 17: This LGTM, although optionally we can just use 'stable' and 'oldstable' which will keep in lockstep with Go releases. This is also the default in common-ci.

https://github.com/actions/setup-go?tab=readme-ov-file#using-stableoldstable-aliases

wenovus (approved):

.github/workflows/go.yml
robshakir: I looked at this, but we have this 3 version aim, so we rather need "oldoldstable" which doesn't seem to exist. :-( 

Do you think we should change our compatibility contract?

wenovus: I think it makes sense to change it to the latest 2 releases to match Go, since that is the same guarantee for security fixes: https://go.dev/doc/devel/release

This has the benefit of avoiding having to adjust the versioning time there is a security fix to be made that's not compatible with existing versions of Go, e.g. https://github.com/openconfig/ygot/pull/969

I don't want to make the assumption that ygot doesn't care that much about security, so I think for that reason it makes sense to change.

wenovus: Approving now. Will open another PR for changing to 2 releases.

PR #971, opened 2024-04-25
Add `GetOrCreate<ListName>Map` helper for lists.

wenovus:
This helps avoid nil checks when there is a need to operate on the map field directly.

This is included along the other GetOrCreate helper func.


ejbrever (approved):
Thanks!

PR #978, opened 2024-07-10
🧹: apply all @dependabot's suggested security updates.

robshakir:
```
 * (M) go.{mod,sum}
  - Update all packages that @dependabot suggests we update.
```


Ankur19 (approved):

robshakir:
Thanks for the review. :-)

PR #1012, opened 2025-02-12
Handle enums in protomap

DanG100:


robshakir (approved):

robshakir:
Thanks for the fix here, @DanG100 -- this turned up in a few places.

PR #1026, opened 2025-03-27
Add a SetWildcardKeys function to Pathtranslator to populate keys for a schema path

yunjie-lu:


robshakir (approved):
thanks for the offline discussion on this -- lgtm.

PR #1028, opened 2025-04-04
Fix a bug in protoFromPathsInternal function in proto.go causing enca…

tengyiG:
…psulation header details to be lost when mapping gNMI paths to protobuf fiedls.

Root cause: the issue stems from incorrect path construction when recursing into nested messages. Something like this will happen: "DEBUG: recursing into nested message gribi_aft.Afts.NextHop.EncapHeader.mpls, path: elem:{name:"afts"} elem:{name:"next-hops"} elem:{name:"next-hop"} elem:{name:"encap-headers"} elem:{name:"encap-header" key:{key:"index" value:"11"}} elem:{name:"afts"} elem:{name:"next-hops"} elem:{name:"next-hop"} elem:{name:"encap-headers"} elem:{name:"encap-header"} elem:{name:"mpls"}.

When findChilren uses these malformed paths to search for nested fields, it returns 0 children because no paths in the input map match this duplicated structure.

### DIFF
--- a/protomap/proto.go
+++ b/protomap/proto.go
@@ -540,7 +540,7 @@ func ProtoFromPaths(p proto.Message, vals map[*gpb.Path]interface{}, opt ...Unma
 	if err != nil {
 		return fmt.Errorf("invalid value prefix supplied, %v", err)
 	}
-	valPrefix = schemaPath(valPrefix)
+	valPrefix = util.GNMIPathToSchemaPath(valPrefix)
 
 	protoPrefix, err := hasProtoMsgPrefix(opt)
 	if err != nil {
@@ -548,15 +548,6 @@ func ProtoFromPaths(p proto.Message, vals map[*gpb.Path]interface{}, opt ...Unma
 	}
 
 	return protoFromPathsInternal(p, vals, valPrefix, protoPrefix, hasIgnoreExtraPaths(opt))
-}
-
-// schemaPath converts the path p into a schema path by removing all of the keys within the path.
-func schemaPath(p *gpb.Path) *gpb.Path {
-	np := proto.Clone(p).(*gpb.Path)
-	for _, e := range np.Elem {
-		e.Key = nil
-	}
-	return np
 }
 
 // findChildren returns the entries from the vals map that correspond to children of the specified protoPrefix path.
@@ -633,7 +624,7 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix, 
 
 		if len(directCh) != 0 {
 			for _, ap := range annotatedPath {
-				trimmedPrefix := schemaPath(protoPrefix)
+				trimmedPrefix := util.GNMIPathToSchemaPath(protoPrefix)
 				if !util.PathMatchesPathElemPrefix(ap, trimmedPrefix) {
 					rangeErr = fmt.Errorf("annotation %s does not match the supplied prefix %s", ap, protoPrefix)
 					return false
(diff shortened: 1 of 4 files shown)


protomap/proto.go
```diff
@@ -722,7 +722,15 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
 			default:
 				childMsg := m.NewField(fd).Message()
 				np := proto.Clone(valPrefix).(*gpb.Path)
-				np.Elem = append(np.Elem, util.TrimGNMIPathElemPrefix(annotatedPath[0], protoPrefix).Elem...)
+				ap := annotatedPath[0]
+				// Check if the path is already property prefixed
+				if len(protoPrefix.Elem) > 0 && len(ap.Elem) > 0 &&	proto.Equal(schemaPath(&gpb.Path{Elem: ap.Elem[:len(protoPrefix.Elem)]}), schemaPath(protoPrefix)) {
```
robshakir, line 727: I assume that the issue here is that we have a path that has keys in it, so in the previous case `util.TrimGNMIPathElemPrefix` was not trimming the prefix because of the keys being present?

protomap/proto.go
```diff
@@ -722,7 +722,15 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
...
+					childPath := &gpb.Path{Elem: ap.Elem[len(protoPrefix.Elem):]}
+					np.Elem = append(np.Elem, childPath.Elem...)
+				} else {
```
robshakir, line 730: Is this else ever actually executed? It seems like:

 * `len(protoPrefix.Elem) > 0` is always going to be `true` unless there's no prefix for the proto.
 *  `len(ap.Elem) > 0` is always going to be `true` if there's any path annotation.
 *  We know that we are in a case where the paths that we've been supplied are descendents of the `protoPrefix` (see the comment on line 606). This means that the `proto.Equal` is always going to return `true` AIUI too.

Based on this it seems like you can avoid either check and simply use some equivalent of `util.TrimGNMIPathElemPrefix` that handles the case where there are keys specified in one of the values unconditionally.

protomap/proto.go
```diff
@@ -722,7 +722,15 @@ func protoFromPathsInternal(p proto.Message, vals map[*gpb.Path]any, valPrefix,
 			default:
 				childMsg := m.NewField(fd).Message()
 				np := proto.Clone(valPrefix).(*gpb.Path)
-				np.Elem = append(np.Elem, util.TrimGNMIPathElemPrefix(annotatedPath[0], protoPrefix).Elem...)
+				ap := annotatedPath[0]
```
robshakir, line 725: Can we add a unit test that demonstrates the bug that was identified here so that we have ongoing coverage please?

robshakir (commented):
Please `gofmt` the code to make the static check pass.

protomap/proto.go
tengyiG: Yes. The problem is that when a path contains keys, util.TrimGNMIPathElemPrefix() does'nt properly trim the prefix because it compares path elements directly, including the keys.

In this case specifically, we have paths with keys that need to be processed corretly to preserve the nested structure deteails. The previous version failed tocorrectly identify that the path should be processed as a child path when keys were present. The new implementation makes sure correct path handling when keys are present.

protomap/proto.go
tengyiG: Great catch! I have added a function in util named TrimGNMIPathElemPrefixKeyAware

protomap/proto.go
tengyiG: Done. Added tests in protomap/proto_test.go and util/gnmi_test.go

robshakir (approved):
Thanks for the modifications and adding the test cases, they also serve as nice documentation. thanks!

PR #1030, opened 2025-04-15
Add fields to IR and IROptions used by settting the path origin from the YANG module name or an specified name.

keogaki:
See [ygnmi PR154](https://github.com/openconfig/ygnmi/pull/154).

### DIFF
--- a/ygen/genir.go
+++ b/ygen/genir.go
@@ -53,6 +53,13 @@ type IROptions struct {
 	// to true.
 	// NOTE: This flag will be removed by v1 release.
 	AppendEnumSuffixForSimpleUnionEnums bool
+
+	// UseModuleNameAsPathOrigin specifies whether the YANG module name is
+	// set to the origin for generated gNMI paths when producing the IR.
+	UseModuleNameAsPathOrigin bool
+
+	// PathOriginName specifies the orign name for generated gNMI paths when producing the IR.
+	PathOriginName string
 }
 
 // GenerateIR creates the ygen intermediate representation for a set of
@@ -184,5 +191,7 @@ func GenerateIR(yangFiles, includePaths []string, langMapper LangMapper, opts IR
 		opts:          opts,
 		fakeroot:      rootEntry,
 		parsedModules: mdef.modules,
+		UseModuleNameAsPathOrigin: opts.UseModuleNameAsPathOrigin,
+		PathOriginName: opts.PathOriginName,		
 	}, nil
 }
--- a/ygen/ir.go
+++ b/ygen/ir.go
@@ -312,6 +312,9 @@ type IR struct {
 	// fakeroot stores the fake root's AST node for creating a serialized
 	// version of the AST if needed.
 	fakeroot *yang.Entry
+
+	UseModuleNameAsPathOrigin bool
+	PathOriginName string
 }
 
 // OrderedDirectoryPaths returns the absolute YANG paths of all ParsedDirectory


ygen/genir.go
```diff
@@ -53,6 +53,13 @@ type IROptions struct {
 	// to true.
 	// NOTE: This flag will be removed by v1 release.
 	AppendEnumSuffixForSimpleUnionEnums bool
+
```
robshakir, line 56: Where are these fields populated? They should be populated here and then handed in to the consumers of the IR, rather than implemented here and populated from downstream consumers.

Per https://github.com/openconfig/ygot/blob/master/docs/code-generation-design.md -- the intention is that the IR provides means to not require any downstream parsing of YANG -- and it should really be considered immutable by downstream consumers.

ygen/ir.go
```diff
@@ -312,6 +312,13 @@ type IR struct {
 	// fakeroot stores the fake root's AST node for creating a serialized
 	// version of the AST if needed.
 	fakeroot *yang.Entry
+
+	// UseModuleNameAsPathOrigin specifies whether the YANG module name is
```
robshakir, line 316: This doesn't seem to make sense in the IR -- rather than storing the argument of the flag, we should rather provide information in the IR for each leaf that specifies what the origin that _should_ be used is.

ygen/ir.go
```diff
@@ -312,6 +312,13 @@ type IR struct {
...
+	// set to the origin for generated gNMI paths.
+	UseModuleNameAsPathOrigin bool
+
+	// PathOriginName specifies the orign name for generated gNMI paths.
```
robshakir, line 320:
```suggestion
	// PathOriginName specifies the origin name for generated gNMI paths.
```

robshakir (changes requested):
Thanks for the contribution -- but this doesn't look like the right design to me. We're adding things to the ygot-generated IR that are not really in-line with what the IR should be specifying -- i.e., pre-parsed details of YANG structures that code-generation can use.

ygen/genir.go
```diff
@@ -178,11 +185,13 @@ func GenerateIR(yangFiles, includePaths []string, langMapper LangMapper, opts IR
...
-		opts:          opts,
-		fakeroot:      rootEntry,
-		parsedModules: mdef.modules,
+		Directories:               dirDets,
+		Enums:                     enumDefinitionMap,
+		ModelData:                 mdef.modelData,
+		opts:                      opts,
+		fakeroot:                  rootEntry,
+		parsedModules:             mdef.modules,
+		UseModuleNameAsPathOrigin: opts.UseModuleNameAsPathOrigin,
```
robshakir, line 194: This doesn't seem the right information to be including in the IR -- you're strictly handing flags that are provided as options back to the caller, rather than doing the parsing for them.

Per the design here: https://github.com/openconfig/ygot/blob/master/docs/code-generation-design.md -- the IR is designed to abstract the details of the YANG away such that multiple downstream consumers can use the same representation.

A better design here would be to have ygen parse the information and populate a field in the IR that is specified per-node (leaf/container/list etc.) that determines what `origin` should be used for that particular field.

ygen/ir.go
keogaki: I made a mistaken Re-request review. We will fix later.

ygen/genir.go
keogaki: @robshakir 
Thanks for your prompt review.
From my understanding of [the design](https://github.com/openconfig/ygot/blob/master/docs/code-generation-design.md), it seems appropreate to add the OriginName field to NodeDetails struct defined in ygen/ir.go and put the desired origin name there depending on the modified IROptions struct value in getOrderedDirDetails() function in ygen/directory.go. 
Then, we can retrieve the origin name in pathgen/pathgen.go and remove the both added fields from IR struct.
Is this right approach?

robshakir: Yes, you can add an `Origin` field to the `YANGNodeDetails` struct in `ygen/ir.go` as one option. Alternatively, you already have `BelongingModule` there, which seems like it should include the contents that you want already?

ygen/ir.go
keogaki: As you suggested, I modified this to parse the origin name information during IR generation, remove the flag field from IR struct and move the origin name field under NodeDetails struct.

ygen/ir.go
keogaki: This comment was removed.

ygen/genir.go
dplore: @keogaki I think Rob is asking for PathOriginName to be added here: https://github.com/openconfig/ygot/pull/1030/files#diff-dc34402cead921a6f95995944775ed0225c7ce5570cadfae4fc4c461e274984eR713

And that field should be populated with `openconfig`, or optionally the user specified or module name derived origin value.

If I understand correctly, the design concept is the complete content of the IR is derived 100% from YANG.  By adding the origin name (default of openconfig or user or module derived) to the `YANGNodeDetails` we preserve this design concept.    In OC YANG modules, the  [origin extension](https://github.com/openconfig/public/blob/master/release/models/openconfig-extensions.yang#L219) with the YANGNodeDetails.origin value defines this.  But Cisco modules are missing this extension, so you are providing an option for the user to specify it or derive from the module name instead.

This is why genir.go  doesn't need  `UseModuleNameAsPathOrigin`.

keogaki: @dplore

Thanks for the comment.
OK, we will update ygen to use `YANGNodeDetails` with `Origin` field as @robshakir suggested.

>In OC YANG modules, the origin extension with the YANGNodeDetails.origin value defines this. But Cisco modules are missing this extension, so you are providing an option for the user to specify it or derive from the module name instead.
>
>This is why genir.go doesn't need UseModuleNameAsPathOrigin.

Yes. OC YANG modules with the origin extension don't need `UseModuleNameAsPathOrigin` in genir.go, but **non** OC YANG modules, i.e. Cisco modules, may not define the similar capability.
Therefore, we need `UseModuleNameAsPathOrigin` for **non** OC YANG modules.

keogaki: @robshakir @dplore 

As you two sugessted, we modified the codes as followings, could you review these?
- ygot/ygen/ir.go
Add `Origin` field to `YANGNodeDetails` struct.

- ygot/ygen/directory.go
In getOrderedDirDetails() function: 
If `PathOriginName` field of `IROptions` struct, opts, has a value, copy it to `Origin` field of `YANGNodeDetails` struct, nd.YANGDetails, used in getNodeDataMap() in ygnmi/pathgen/pathgen.go below.
Else if the value of `UseModuleNameAsPathOrigin` field of `IROptions` struct is True, copy the value of nd.YANGDetails.BelongingModule, i.e. YANG module name, to the `Origin` field.
Else copy "openconfig" to the `Origin` field as the default value.

- ygnmi/pathgen/pathgen.go
  - In getNodeDataMap() function, if `Origin` field of `YANGNodeDetails` struct, field.YANGDetails, has a value, copy it to `PathOriginName` field of `NodeData` struct, nodeData.
  - goPathStructTemplate generates PathOriginName() function returning the path origin name if the above nodeData.PathOriginName has a value except the default origin, "openconfig".

- ygnmi/pathgen/pathgen.test.go
  - In TestGeneratePathCode(), modify test cases below following the above changes:
    - simple_openconfig_test
    - simple_openconfig_test_with_preferOperationalState=false
    - simple_openconfig_test_with_excludeState=true
    - simple_openconfig_test_with_ordered_list
    - simple_openconfig_test_with_ordered_list_but_atomic_generation_turned_off
    - simple_openconfig_test_with_union_&\_typedef\_&\_identity\_&\_enum
    - simple_openconfig_test_with_union\_&\_typedef\_&\_identity\_&_enum,_with_enum_names_not_shortened
    - simple_openconfig_test_with_submodule_and_union_list_key
    - simple_openconfig_test_with_augmentations

  - In TestGetNodeDataMap(), modify the test case, "small test with PathOriginName", following the above changes.

- ygot/ygen/directory_test.go
In TestGetOrderedDirDetailsPathOrigin(), modify the test case, ""PathOriginName and UseModuleNameAsPathOrigin are not set", following the above changes.
Also, modify getPathOriginNames() support funciton following the above changes.

ygen/directory.go
```diff
@@ -276,6 +276,17 @@ func getOrderedDirDetails(langMapper LangMapper, directory map[string]*Directory
...
+			// Else if IROptions.UseModuleNameAsPathOrigin of the node is true,
+			// YANG module name is set to the Origin of the node.
+			// Else "openconfig" is set to the Origin of the node.
+			if opts.PathOriginName != "" {
+				nd.YANGDetails.Origin = opts.PathOriginName
+			} else if opts.UseModuleNameAsPathOrigin {
+				nd.YANGDetails.Origin = nd.YANGDetails.BelongingModule
+			} else {
+				nd.YANGDetails.Origin = "openconfig"
+			}
```
robshakir, line 289:
```suggestion
            switch {
            case opts.PathOriginName != "":
				nd.YANGDetails.Origin = opts.PathOriginName
			case opts.UseModuleNameAsPathOrigin:
				nd.YANGDetails.Origin = nd.YANGDetails.BelongingModule
			default:
			    // TODO: read the origin from the relevant YANG extension.
				nd.YANGDetails.Origin = ""
			}
```

Two changes:
1. Use a switch rather than if/else if/else to make this easier to extend in the future.
2. Switch the "default" origin to be "", since making it "openconfig" ties this statically to OpenConfig models, which might not be the case.
3. Add a TODO to say that we should be able to read this value from the relevant extension in the YANG model (https://github.com/openconfig/public/blob/master/release/models/openconfig-extensions.yang#L219-L228).

ygen/directory_test.go
```diff
@@ -702,3 +705,176 @@ func TestFindMapPaths(t *testing.T) {
...
+	return "", nil
+}
+
+func (*mockLangMapper) KeyLeafType(e *yang.Entry, opts IROptions) (*MappedType, error) {
+	return nil, nil
+}
+
+func (*mockLangMapper) LeafType(e *yang.Entry, opts IROptions) (*MappedType, error) { return nil, nil }
+
+func (*mockLangMapper) PackageName(e *yang.Entry, compressBehavior genutil.CompressBehaviour, nested bool) (string, error) {
```
robshakir, line 723:
```suggestion
func (*mockLangMapper) PackageName(_ *yang.Entry, _ genutil.CompressBehaviour, _ bool) (string, error) {
```

ygen/directory_test.go
```diff
@@ -702,3 +705,176 @@ func TestFindMapPaths(t *testing.T) {
...
+
+func (*mockLangMapper) DirectoryName(e *yang.Entry, complessBehavior genutil.CompressBehaviour) (string, error) {
+	return "", nil
+}
+
+func (*mockLangMapper) KeyLeafType(e *yang.Entry, opts IROptions) (*MappedType, error) {
+	return nil, nil
+}
+
+func (*mockLangMapper) LeafType(e *yang.Entry, opts IROptions) (*MappedType, error) { return nil, nil }
```
robshakir, line 721:
```suggestion
func (*mockLangMapper) LeafType(_ *yang.Entry, _ IROptions) (*MappedType, error) { return nil, nil }
```

ygen/directory_test.go
```diff
@@ -702,3 +705,176 @@ func TestFindMapPaths(t *testing.T) {
...
+
+type mockLangMapper struct{}
+
+func (*mockLangMapper) FieldName(e *yang.Entry) (string, error) { return "", nil }
+
+func (*mockLangMapper) DirectoryName(e *yang.Entry, complessBehavior genutil.CompressBehaviour) (string, error) {
+	return "", nil
+}
+
+func (*mockLangMapper) KeyLeafType(e *yang.Entry, opts IROptions) (*MappedType, error) {
```
robshakir, line 717:
```suggestion
func (*mockLangMapper) KeyLeafType(_ *yang.Entry, _ IROptions) (*MappedType, error) {
```

ygen/directory_test.go
```diff
@@ -702,3 +705,176 @@ func TestFindMapPaths(t *testing.T) {
 		})
 	}
 }
+
+type mockLangMapper struct{}
+
+func (*mockLangMapper) FieldName(e *yang.Entry) (string, error) { return "", nil }
+
+func (*mockLangMapper) DirectoryName(e *yang.Entry, complessBehavior genutil.CompressBehaviour) (string, error) {
```
robshakir, line 713:
```suggestion
func (*mockLangMapper) DirectoryName(_ *yang.Entry, _ genutil.CompressBehaviour) (string, error) {
```

ygen/directory_test.go
```diff
@@ -702,3 +705,176 @@ func TestFindMapPaths(t *testing.T) {
 		})
 	}
 }
+
+type mockLangMapper struct{}
+
+func (*mockLangMapper) FieldName(e *yang.Entry) (string, error) { return "", nil }
```
robshakir, line 711:
```suggestion
func (*mockLangMapper) FieldName(_ *yang.Entry) (string, error) { return "", nil }
```

ygen/directory_test.go
```diff
@@ -702,3 +705,176 @@ func TestFindMapPaths(t *testing.T) {
...
+
+func (*mockLangMapper) PackageName(e *yang.Entry, compressBehavior genutil.CompressBehaviour, nested bool) (string, error) {
+	return "", nil
+}
+
+func (*mockLangMapper) InjectEnumSet(entries map[string]*yang.Entry, compressPaths, noUnderscores, skipEnumDedup, shortenEnumLeafNames, useDefiningModuleForTypedefEnumNames, appendEnumSuffixForSimpleUnionEnums bool, enumOrgPrefixesToTrim []string) error {
+	return nil
+}
+
+func (*mockLangMapper) InjectSchemaTree(entries []*yang.Entry) error { return nil }
```
robshakir, line 731:
```suggestion
func (*mockLangMapper) InjectSchemaTree(_ []*yang.Entry) error { return nil }
```

robshakir (commented):
Thanks for the changes. This looks better now that it is encapsulated in the IR generation. I have one suggested logic change, and then some style changes.

ygen/directory_test.go
```diff
@@ -702,3 +705,176 @@ func TestFindMapPaths(t *testing.T) {
...
+	return nil, nil
+}
+
+func (*mockLangMapper) LeafType(e *yang.Entry, opts IROptions) (*MappedType, error) { return nil, nil }
+
+func (*mockLangMapper) PackageName(e *yang.Entry, compressBehavior genutil.CompressBehaviour, nested bool) (string, error) {
+	return "", nil
+}
+
+func (*mockLangMapper) InjectEnumSet(entries map[string]*yang.Entry, compressPaths, noUnderscores, skipEnumDedup, shortenEnumLeafNames, useDefiningModuleForTypedefEnumNames, appendEnumSuffixForSimpleUnionEnums bool, enumOrgPrefixesToTrim []string) error {
```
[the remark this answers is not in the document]
keogaki: @robshakir 
Thanks for your review.
Fixed as suggeted.

ygen/directory.go
keogaki: @robshakir 
Fixed as suggeted.

ygen/directory_test.go
keogaki: @robshakir 
Fixed as suggeted.

ygen/directory_test.go
keogaki: @robshakir 
Fixed as suggeted.

ygen/directory_test.go
keogaki: @robshakir 
Fixed as suggeted.

ygen/directory_test.go
keogaki: @robshakir 
Fixed as suggeted.

PR #1084, opened 2026-05-30
escape backslash in key values in elemToString

jmestwa-coder:
elemToString escapes = and ] in a PathElem key value but not the backslash that extractKV uses as its escape character, so the two disagree. A value like a\b serialized by PathToString re-parses to ab, and a trailing backslash lets one key absorb the next, so {k1:"a\", k2:"b"} round-trips to a single k1="a][k2=b]". Escape the backslash first so PathToString output round-trips through StringToStructuredPath.

### DIFF
--- a/ygot/pathstrings.go
+++ b/ygot/pathstrings.go
@@ -145,7 +145,8 @@ func elemToString(name string, kv map[string]string) (string, error) {
 	sort.Strings(keys)
 
 	for _, k := range keys {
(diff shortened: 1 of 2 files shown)


jmestwa-coder:
any update?

robshakir:
Thanks for this -- this looks reasonable to me, and like a bug.

robshakir (approved):

PR #1085, opened 2026-06-01
use strings.Join in PathToString to preserve key values

jmestwa-coder:
Caught this round-tripping random gNMI paths back through StringToStructuredPath. PathToString and PathToSchemaPath join elements with path.Join, which runs path.Clean: it collapses // and resolves .. inside key values, so intf[name=a/../b]/state comes out as b]/state and x//y silently becomes x/y. Join with strings.Join so the serialized path is left alone.


jmestwa-coder:
any update?

PR #1086, opened 2026-06-02
honor escapes inside keys in SplitPath

jmestwa-coder:
SplitPath treats / inside [...] as part of a key, but it ignored backslash escapes there, so an escaped ] cleared the in-key state and the next / was read as an element separator. PathToString escapes ] in key values but not /, so a key value containing ] followed by / no longer round-trips: StringToStructuredPath splits the element and errors, though extractKV parses it fine. Track escapes inside keys so the escaped ] keeps the key open, leaving the backslash for the key parser.

### DIFF
--- a/util/path.go
+++ b/util/path.go
@@ -401,6 +401,12 @@ func SplitPath(path string) []string {
 			inKey = true
 		case ch == ']' && !inEscape:
 			inKey = false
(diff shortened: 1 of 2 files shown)


jmestwa-coder:
gentle ping

PR #1092, opened 2026-07-15
gnmidiff: remove trie dependency

### LINKED ISSUE #1088: avoid dependency on trie
I notice that this module has a dependency on another package, github.com/derekparker/trie. I believe this dependency can be easily avoided, while actually making the code faster. (Google has--for reasons I cannot fathom--made me the owner of the copy of this trie package vendored into its monorepo, so I have an interest in this outcome.)

Instead of building a trie then calling the PrefixSearch method, you could instead simply build an array of elements, sort it in the usual way, then check for conflicts in this array: find the first prefix-matching element, then scan forward over all successive elements that have a prefix match. A throwaway benchmark suggests this is at least 10x faster and allocates much less.

This could be abstracted as a convenient helper utility, avoiding the dependency.

adonovan:
This change replaces the dependency on Derek Parker's trie
package with a naive (yet much faster) implementation.

Fixes #1088

### DIFF
--- a/gnmidiff/prefix.go
+++ b/gnmidiff/prefix.go
@@ -1,0 +1,48 @@
+// Copyright 2023 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


adonovan:
@robshakir

robshakir (approved):

robshakir:
Thanks for this -- apologies, I was intending to look at your other issue, but you beat me to it :-)

adonovan:
Thanks. If you care to reimport ygot into google3 at your leisure, that too would be much appreciated.

PR #1095, opened 2026-07-21
reject unterminated key predicates in extractKV

nabhan06:
extractKV doesn't check whether it is still inside a [key=value] predicate when the input runs out, so an element with a dangling predicate is silently mis-parsed rather than rejected: the element name is discarded and the trailing key value takes its place. StringToStructuredPath(`/a/b[k=c`) returns /a/c with no error, and PathToSchemaPath on a deprecated Element of `a[b=c` returns /c, so code that string-checks a path before converting it can end up resolving a different node than the one it inspected. Same story for a trailing escape character, where the backslash is just dropped.

This errors on both cases at the end of extractKV, with test cases covering the string and Element forms.

### DIFF
--- a/ygot/pathstrings.go
+++ b/ygot/pathstrings.go
@@ -294,6 +294,14 @@ func extractKV(in string) (string, map[string]string, error) {
 		inEscape = false
 	}
 
+	if inKey {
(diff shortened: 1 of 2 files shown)


robshakir (approved):
Thanks for this change! Looks a good catch.

PR #1096, opened 2026-07-22
anchor the whole YANG pattern in fixYangRegexp

nabhan06:
fixYangRegexp only wraps the converted pattern in parens when the YANG pattern does not already start with a caret, so `^ab|cd$` is passed to the regexp package unchanged and evaluated as `(^ab)|(cd$)`. Since YANG patterns are implicitly anchored to the whole string, ValidateStringRestrictions then accepts values the schema forbids, e.g. "abcdef" for a leaf restricted to `^ab|cd$`.

The same first/last character inspection also treats an escaped trailing dollar as an anchor: a pattern like `[0-9]+\$` becomes `^([0-9]+\)$`, which does not compile, so every value for that leaf fails validation with a regexp parse error.

This strips whatever anchors the pattern already carries (a trailing `$` counts as an anchor only when it is unescaped) and then always wraps the remainder in `^(...)$`. Existing expectations for `^abc` and `^abc$` change to `^(abc)$`, which is equivalent.


robshakir (approved):
Thanks -- this one is quite nuanced, but I'm a little surprised we haven't seen it before (perhaps because things are made more permissive). Thanks for the fix.

PR #1097, opened 2026-07-23
reject non-object list elements in unmarshalList

nabhan06:
guard the per-element type assertion in unmarshalList so a schema list whose JSON array holds a scalar returns an error instead of panicking, matching the non-object handling unmarshalContainer already does.


robshakir (approved):
Thanks for catching this, and adding tests.

PR #1098, opened 2026-07-29
reject non-finite decimal64 values on unmarshal

nabhan06:
sanitizeJSON and stringToKeyType parse a decimal64 leaf/key from an untrusted JSON string with strconv.ParseFloat, which happily accepts NaN, Inf and Infinity, so Unmarshal stores a non-finite float64 for a leaf whose schema can never hold one. That value then cannot be round-tripped back out since encoding/json rejects NaN and Inf, and a stored NaN quietly breaks equality and diffing. Reject a non-finite result at both parse sites so the value is turned away where it enters rather than surfacing much later.

### DIFF
--- a/ytypes/leaf.go
+++ b/ytypes/leaf.go
@@ -17,6 +17,7 @@ package ytypes
 import (
 	"encoding/base64"
 	"fmt"
+	"math"
 	"math/big"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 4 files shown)


robshakir (approved):
Thank you, merging this one too -- appreciate the catch.

PR #1099, opened 2026-08-18
reject out-of-range decimal64 precision in sanitizeGNMI

nabhan06:
sanitizeGNMI decodes a gNMI Decimal64 by raising 10 to the power of Precision, a uint32 read straight from the untrusted value with no upper bound. A YANG decimal64 has at most 18 fraction digits, so a message that sets a decimal64 leaf with a large Precision makes big.Int.Exp build a multi-gigabyte integer (Precision=20,000,000 already costs 2.3s and 227MB, and the uint32 max is far worse). Reject any Precision above 18 before the exponentiation, matching the nil check already right above it.


robshakir (approved):

robshakir:
Thanks for this change, it's a good catch.

PR #1105, opened 2026-09-18
Tolerate gzip output variation in schema tests.

robshakir:
```
 * (M) ygen/schemaparse_test.go
   - In Go 1.27, the compress/flate encoder implementation was modified,
     altering the exact byte sequence emitted by compress/gzip. Validate
     WriteGzippedByteSlice by asserting that the produced payload successfully
     decompresses to the original input, rather than comparing against a
     static byte sequence.
 * (M) gogen/codegen_test.go
   - Normalise the ySchema byte slice when comparing generated Go code to
     golden files such that variations in gzip compression output across Go
     versions do not cause test failures.
   - Explicitly verify that the extracted schema bytes decompress to the
     expected RawJSONSchema.
```
In Go 1.27, the `compress/flate` encoder was updated, altering the compressed byte sequence produced by `compress/gzip` (utilised by `ygen.WriteGzippedByteSlice`). Because our CI matrix runs both `stable` (Go 1.27) and `oldstable` (Go 1.26), asserting on exact gzipped byte slices causes `stable` to fail against golden files generated on earlier Go releases. This change decouples unit tests from underlying Go DEFLATE implementation details whilst continuing to validate that the schema decompresses correctly.
Fixes failing CI on Go `stable`.


gprateekjn (approved):
LGTM

robshakir:
> LGTM

Thanks @gprateekjn!

